PR #84, opened 2016-11-18
added exclusions list

nkartyshov:
https://github.com/AdguardTeam/AdguardDNS/issues/79


ameshkov:
Add filled in exclusions.txt file (ask @Alex-302 to help you with it).
Also add compiled filter.txt as well (just for me to understand the difference).

nkartyshov:
I added few exclusion in the file.
Added the compiled filter.txt and a temp files (diff and old file of filters)

PR #193, opened 2017-10-14
googlesyndication.com typo

alicaccs:


Alex-302:
Not typo. It is not Google domain.

alicaccs:
So you are not blocking Google AdSense. Is it intended?

Alex-302:
Look
image
We can't block it completely. A lot of sites will be damaged.

alicaccs:
Alright I get it. But nothing about the so called non-intrusive ads (in ABP language) is mentioned in your DNS product webpage. Detailed information about your criteria of what ads get blocked is welcomed.

Alex-302:
> Detailed information about your criteria of what ads get blocked is welcomed.

advertising/tracking domains only. Some domains are excluded, because we cannot afford to spoil a lot of popular sites.
Specific URLs cannot be blocked by DNS filtering, element hiding is impossible in this case.

ameshkov:
It's not about being intrusive or not, blocking some domains on the DNS-level leads to breaking quite a lot of websites'.

What for AG filters, here's the policy:
https://kb.adguard.com/en/general/adguard-filter-policy

PR #216, opened 2017-12-19
Added support parsing `##` symbols

nkartyshov:
https://github.com/AdguardTeam/AdguardDNS/issues/215


nkartyshov:
In the list of filters there were rules with ##, which should not be. And it was breaking the import of rules into iOS.

nkartyshov:
Fixed

PR #249, opened 2018-03-09
update the dnscrypt url

evilvibes:
.org to .info


evilvibes:
You should also update the url on (https://adguard.com/en/adguard-dns/overview.html#faq-2-11)

PR #254, opened 2018-03-14
clixsense.com https://github.com/AdguardTeam/AdguardFilters

Alex-302:


Filters/exclusions.txt
```diff
@@ -1,3 +1,5 @@
+! https://github.com/AdguardTeam/AdguardFilters
```
ameshkov: What exact issue?

PR #291, opened 2018-06-20
Fixing official Amazon app https://github.com/AdguardTeam/AdguardFilters/issues/18469

Alex-302:


ameshkov:
This will unblock A LOT of mobile ads.

ameshkov:
Advise user to unblock it via the

Alex-302:
But we have more than 60 exclusions(a lot of sites) with this domain in Eng filter.

Alex-302:
Ok

PR #392, opened 2018-10-18
Add Dockerfiles

ewrogers:
Added Dockerfiles for Linux (386, amd64, and arm)

Tested working on RPi3+ as well.

```
docker run -d \
    -p 53:53 \
    -p 53:53/udp \
    -p 80:3000 \
    -v adguard_data:/data/adguard \
    --restart always \
    --name adguard-home \
    ewrogers/adguard-home-arm:latest
```
Able to visit http://adguard-home in my browser and DNS queries are handled. Settings also persist across container restarts.

### DIFF
--- a/Dockerfile.arm
+++ b/Dockerfile.arm
@@ -1,0 +1,48 @@
+FROM easypi/alpine-arm:latest
+LABEL maintainer="Erik Rogers <erik.rogers@live.com>"
+
+# AdGuard version
+ARG ADGUARD_VERSION="0.9"
+ENV ADGUARD_VERSION $ADGUARD_VERSION
+
+# AdGuard architecture and package info
+ARG ADGUARD_ARCH="linux_arm"
+ENV ADGUARD_ARCH ${ADGUARD_ARCH}
+ENV ADGUARD_PACKAGE "AdGuardHome_v${ADGUARD_VERSION}_${ADGUARD_ARCH}"
+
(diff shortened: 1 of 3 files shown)


hmage:
Thank you for your contribution 🎉

HenryQW:
`-v ~/adguard/AdGuardHome.yaml:/data/adguard/AdGuardHome.yaml`

Seems this will break the container.

ewrogers:
Does it do similar if you use absolute paths instead of the `~/adguard` prefix?

I have noticed similar oddities with other containers in the past, trying to mount single files into a volume. Will take a look.

HenryQW:
No luck, my workaround:
```
docker exec AdGuardHome rm /data/adguard/AdGuardHome.yaml
docker cp ~/adguard/AdGuardHome.yaml AdGuardHome:/data/adguard/AdGuardHome.yaml
docker restart AdGuardHome
```

frenck:
So... the binary is stored in the data volume...?
How are you planning on handle upgrades?

ewrogers:
Correct, this is a problem. I only did it that way because pre-hotfix the binary wouldn't look in a data directory.

A proper solution would be to only use the volume as a data directory (non-binary). Haven't had time to fix/test that it works.

frenck:
Thx, I'll work around it for the Hass.io add-on.

PR #411, opened 2018-10-20
Formatted 'Rules count' number with thousands separator

mustakimali:
image

### DIFF
--- a/client/src/components/Filters/index.js
+++ b/client/src/components/Filters/index.js
@@ -50,6 +50,7 @@ class Filters extends Component {
         Header: 'Rules count',
         accessor: 'rulesCount',
         className: 'text-center',
+        Cell: props => props.value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")
     }, {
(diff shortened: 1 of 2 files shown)


client/src/components/Filters/index.js
```diff
@@ -50,6 +50,7 @@ class Filters extends Component {
         Header: 'Rules count',
         accessor: 'rulesCount',
         className: 'text-center',
+        Cell: props => props.value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ','),
```
ameshkov: Using a regular expression is a bit too much for that:)

Use a [culture-specific approach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) instead: `new Number(100000).toLocaleString()`

mustakimali: Thanks for the suggestion and it's great. I've pushed another commit.

ameshkov:
Thank you for the contribution!

PR #431, opened 2018-10-26
 Allow click footer link when loading

hoang-rio:


IldarKamalov:
Thank you for the contribution👍

PR #441, opened 2018-10-29
Add i18n

hoang-rio:
Add i18n and translate client dashboard to Vietnamese


ameshkov:
Uh, that's a huge PR, thank you!

Tbh, we wanted to finalize the dashboard UI first, and add translations a bit later after that. We will reuse this PR partly when it's time: https://github.com/AdguardTeam/AdGuardHome/issues/442

ameshkov:
I see you're supporting this PR, thank you for that!

In order for this PR to be merged, there's a few changes that need to be done:

1. Base language (English) strings need to be added as a separate file: `client/src/__locales/en.js`
2. Every string should have an identifier, e.g. `Blocked malware/phishing` -> `stat_blocked_malware`
3. In the JS files, we should use these identifiers instead of plain English strings

The reason for this change is translation automation via adguard.oneskyapp.com or crowdin. If we don't do it and keep using the current approach, every change to the base language will basically add a new string for translation instead of changing the existing one.

hoang-rio:
> I see you're supporting this PR, thank you for that!
> 
> In order for this PR to be merged, there's a few changes that need to be done:
> 
> 1. Base language (English) strings need to be added as a separate file: `client/src/__locales/en.js`
> 2. Every string should have an identifier, e.g. `Blocked malware/phishing` -> `stat_blocked_malware`
> 3. In the JS files, we should use these identifiers instead of plain English strings
> 
> The reason for this change is translation automation via adguard.oneskyapp.com or crowdin. If we don't do it and keep using the current approach, every change to the base language will basically add a new string for translation instead of changing the existing one.

Ok. I will update it.

client/src/__locales/en.js
```diff
@@ -0,0 +1,135 @@
...
+        example_meaning_host_block: 'AdGuard Home will now return 127.0.0.1 address for the example.org domain (but not its subdomains).',
+        example_comment: '! Here goes a comment',
+        example_comment_meaning: 'just a comment',
+        example_comment_hash: '# Also a comment',
+        all_filters_up_to_date_toast: 'All filters are already up-to-date',
+        updated_upstream_dns_toast: 'Updated the upstream DNS servers',
+        dns_test_ok_toast: 'Specified DNS servers are working correctly',
+        dns_test_not_ok_toast: 'Server "{{key}}": could not be used, please check that you\'ve written it correctly',
+        // Logs
+        ublock_btn: 'Unblock',
```
ameshkov: `ublock_btn` -> `unblock_btn`

ameshkov:
@IldarKamalov please review and also check out this task: https://github.com/AdguardTeam/AdGuardHome/issues/442#issuecomment-437709171

PR #508, opened 2018-12-31
Update README with instructions for setcap non-root use

mordyovits:
On Linux you can run it listening on port 53 without root privs.  This is the best option: clients still send on port 53 (no wonky configs) and AdGuard doesn't run as root (!).


hmage:
Thank you, much better!

PR #526, opened 2019-01-04
Added dockerfile for arm64 and using the official alpine image for th…

mquarneti:
…e arm image


ameshkov:
@mquarneti thank you for the contribution!

I've assigned this PR to the "Docker image" milestone. We're going to sort out all the docker-related issues later this month.

mquarneti:
thanks

hmage:
Hi!

We've implemented our own way on doing docker, hope you find it satisfactory -- it should be doing everything this pull request is doing. The reason why we did it that way is that we will also have automated dockerhub builds for all supported architectures as well.

If you have any more questions, or have issues, or want more functionality, feel free to open an issue or pull request based on the recent master.

PR #612, opened 2019-03-02
UI BG Language translation

mveplus:
I'm not sure this is the right way to do a pull request?
Please advise if I need to change the from.


ameshkov:
@mveplus 

Thank you! @IldarKamalov will take it from here.

Generally, the translation should've been added here: http://translate.adguard.com/collaboration/project?id=153384, and then we import it to the project in due course. Here is some more info on how AdGuard is translated: https://kb.adguard.com/en/general/adguard-translations

However, once you've already made a pull request with the translated strings, we'll upload it to oneskyapp manually: https://github.com/AdguardTeam/AdGuardHome/issues/615

mveplus:
@ameshkov Thanks, 

I found as well http://translate.adguard.com/collaboration/project?id=153384 but was after I did complete the translation. I'll review it there because it's certainly a lot easier when you see the context on the interface screenshots.
I did read through https://kb.adguard.com/en/general/adguard-translations - indeed very useful cross-language advice a while ago.

IldarKamalov:
Added in https://github.com/AdguardTeam/AdGuardHome/commit/10fcfefcfd7220d68880cd735a3844ec4e617555, thank you!

PR #671, opened 2019-03-26
Chrome nagging translation popup from portuguese

adrianrudnik:
Chrome with `de, de-de, en, en-us` language ordering keeps nagging for a translation from portuguese to german. This would disable any nagging, not sure why they do not respect the html[lang] attribute in this case even the content is clearly english.

Screenshot from 2019-03-26 12-37-20


ameshkov:
I'd prefer to have the option of using auto-translate.

We'll fix it a little bit differently, by adding a `content-language` tag.

ameshkov:
Huh, and it seems setting content-language does not fix the issue so we'll have to do what you propose :(

PR #720, opened 2019-04-17
Run as non-root user

javabean:
Containers security best practices say we should not run anything as root. In fact, this is actually enforced in some OpenStack distributions.

This patch makes AdguardHome run as `nobody`.

A drawback is that volume writing will happen as non-root (obviously), and thus file permissions will show their ugly head in the face of the user when she tries to mount a local directory in the container.  
Since the target user is advanced enough to use Docker, I think this is a reasonable tradeoff.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,14 +11,21 @@ FROM alpine:latest
 LABEL maintainer="AdGuard Team <devteam@adguard.com>"
 
 # Update CA certs
-RUN apk --no-cache --update add ca-certificates && \
-    rm -rf /var/cache/apk/* && mkdir -p /opt/adguardhome
+RUN apk --no-cache --update add ca-certificates libcap && \
+    rm -rf /var/cache/apk/* && mkdir -p /opt/adguardhome/conf /opt/adguardhome/work
 
 COPY --from=build /src/AdGuardHome/AdGuardHome /opt/adguardhome/AdGuardHome
+
+RUN chown -R nobody: /opt/adguardhome \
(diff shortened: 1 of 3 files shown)


ameshkov:
@ezbiranik plz take a look

The first run might be problematic as we made it mandatory to run the initial configuration wizard under root. To solve this, we will need to provide some default configuration `AdGuardHome.yaml` in the container.

javabean:
Beware putting a default `AdGuardHome.yml` in the container will be useless if the user mounts `/opt/adguardhome/config` (which is recommended to persist configuration).
Why do you need root on first run?

ameshkov:
> Why do you need root on first run?

We want to let the user configure the device -- for instance, set the static IP address. Also, running without root requires additional configuration (`setcap` for instance), and that's problematic for a regular user.

This all can be handled properly, but this requires additional work. I've created a feature request for that: https://github.com/AdguardTeam/AdGuardHome/issues/723

javabean:
I've cleaned up a bit the Dockerfile and removed the `nobody` user for now. Also introduced a sample docker-compose file for easy copy'n'paste.
The binary and directories are still owned by `nobody`, which is usually not a problem for the `root` user. ;-) The `setcap` ensures we can easily change the runtime user after first launch.

So now there is no functional difference compared to the currently published Docker version (100% backward compatible), but it enables more flexibility where it matters.

ameshkov:
Thank you! Summoning @ezbiranik again as I am not an expert on docker

ameshkov:
@szolin could you plz merge it properly (bitbucket first)

Issue #742, opened 2019-05-01
Implement NULL blocking in addition to NXDOMAIN


alexzeitgeist:
Filtered queries are currently answered with an empty response and status NXDOMAIN. It would be nice if there was an option to answer filtered queries with the unspecified address 0.0.0.0 instead.

For reference, see the [pi-hole documentation](https://docs.pi-hole.net/ftldns/blockingmode/).

ameshkov:
Marked as a feature request, thank you!

Please note, that it uses NXDOMAIN for adblock syntax filter lists only. If you use a hosts file, AGH will respond with the IP address specified there.

ameshkov:
@szolin considering that it's done in https://github.com/AdguardTeam/AdGuardHome/pull/743, I assigned it to v0.96.

The only thing left to be done is updating Wiki: https://github.com/AdguardTeam/AdGuardHome/wiki/Configuration#configuration-file

PR #743, opened 2019-05-04
* dnsforward, config: add unspecified IP blocking option

alexzeitgeist:
This feature adds support to answer filtered queries with the unspecified address (0.0.0.0), also referred to as [NULL blocking mode in pi-hole](https://docs.pi-hole.net/ftldns/blockingmode/). Also see [issue #742 ](https://github.com/AdguardTeam/AdGuardHome/issues/742).

At least in my non-scientific testing on an Android device, this mode works better than the default NXDOMAIN filtering, as it reduces the number of repeated dns queries.

It is enabled by adding `null_filter: true` to the dns section of AdGuardHome.yaml, for example:
```
...
dns:
  bind_host: 0.0.0.0
  port: 53
  protection_enabled: true
  filtering_enabled: true
  null_filter: true
...
```

~~Note that also host entries in filter lists with 127.0.0.1 will be answered with the unspecified address. This is intentional as 0.0.0.0 seems more compatible accross different operation systems than 127.0.0.1.~~

Lastly, an option for this feature would still have to be added to the UI frontend. I had some difficulty adding the React code to my Goland project for debugging.. :p

### DIFF
--- a/dnsforward/dnsforward.go
+++ b/dnsforward/dnsforward.go
@@ -26,6 +26,8 @@ const (
 	parentalBlockHost     = "family-block.dns.adguard.com"
 )
 
+const unspecifiedAddress = "0.0.0.0"
+
 // Server is the main way to start a DNS server.
 //
 // Example:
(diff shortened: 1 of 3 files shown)


ameshkov:
@alexzeitgeist thank you!

@szolin plz take a look

szolin:
I think it would be better to have a string-value setting `blocking_mode` instead of boolean setting `null_filter`.

for example:

`blocking_mode: nxdomain`

and

`blocking_mode: null_ip`

This allows us to extend the functionality without having configuration issues with incompatibile settings.

dnsforward/dnsforward.go
```diff
@@ -26,6 +26,8 @@ const (
 	parentalBlockHost     = "family-block.dns.adguard.com"
 )
 
+const unspecifiedAddress = "0.0.0.0"
```
szolin: Please use `net.IPv4zero` instead of this.
This will also allow us to remove `net.ParseIP()` call below and pass it directly to `genARecord()`.

alexzeitgeist: @szolin didn't know about `net.IPv4zero`. Fixed.

Also added a test for the null filter.

Not sure if it's best practice to hardcode the value like `null_ip`, or if it should be added as a constant.

PR #825, opened 2019-06-10
Update zh-cn.json

arfaWong:


ameshkov:
@arfaWong hi, and thank you!

The problem is that AdGuard Home is translated on Oneskyapp, please read the details here:
https://github.com/AdguardTeam/AdGuardHome#translate

PR #933, opened 2019-08-15
add an openrc service-script

magenbrot:
Hi!

Here's a simple openrc service-script I'm using to run AdGuardHome in a LXC container under Proxmox Virtual Environment. PVE has no support for running Docker natively, but for LXC.

regards
Oliver


ameshkov:
Sorry for the late reply, and thank you!

I guess it'd be better if we simply add this script to Wiki.

I've just created a task about that:
https://github.com/AdguardTeam/AdGuardHome/issues/976

Issue #1027, opened 2019-09-25
Content modification


ArchiveBase:
AGHome footer 'Copyright © 2019 AdGuard' > 'AdGuard' link open on same tab then prefer suggest it suitable to set for open in new tab same as Homepage Privacy policy hyperlinks.

ArchiveBase:
🤦Where's @ameshkov🤔

PR #1029, opened 2019-09-27
Footer: Open adguard.com in new tab

onedr0p:
Fixes: https://github.com/AdguardTeam/AdGuardHome/issues/1027

### DIFF
--- a/client/src/components/ui/Footer.js
+++ b/client/src/components/ui/Footer.js
@@ -32,7 +32,7 @@ class Footer extends Component {
                                 <div className="footer__column">
                                     <div className="footer__copyright">
                                         <Trans>copyright</Trans> &copy; {this.getYear()}{' '}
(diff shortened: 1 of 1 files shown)


client/src/components/ui/Footer.js
```diff
@@ -32,7 +32,7 @@ class Footer extends Component {
                                 <div className="footer__column">
                                     <div className="footer__copyright">
                                         <Trans>copyright</Trans> &copy; {this.getYear()}{' '}
-                                        <a href="https://adguard.com/">AdGuard</a>
+                                        <a target="_blank" href="https://adguard.com/">AdGuard</a>
```
IldarKamalov: According to ESlint rules we need to add `rel="noopener noreferrer"` to every link with `target="_blank"`

onedr0p: I didn't know that, let me update.

PR #1189, opened 2019-11-20
Update blocked_services.go Snapchat Rules

Cybo1927:
More Snapchat owned domains


ArchiveBase:
@Cybo1927 These domains not required. snapchat-proxy.appspot.com, snapchat-webfonts.storage.googleapis.com, snapnet-cdn.storage.googleapis.com, sc-static.net 

You can add below if necessary.
- snap.com (www.snap.com)
- sc-cdn.net (block load already signed app account contents)

Cybo1927:
Thanks, I was just going off things I've found from Snapchat. (Currently don't have a phone so couldn't check to see what they each did)

PR #1192, opened 2019-11-21
Update blocked_services.goTikTok Rules

Cybo1927:
Dunno what musemuse.cn is but it connects everytime I open TikTok on my iPad

image

Edit: Possibly related to bytecdn.cn and byteimg.com

https://who.is/whois/musemuse.cn

image

### DIFF
--- a/home/blocked_services.go
+++ b/home/blocked_services.go
@@ -114,6 +114,8 @@ var serviceRulesArray = []svc{
 		"||bytecdn.cn^",
 		"||byteimg.com^",
 		"||ixigua.com^",
+		"||musemuse.cn^",
+		"||musical.ly^",
 	}},
 }
 
--- a/home/blocked_services.go
+++ b/home/blocked_services.go
@@ -114,8 +114,8 @@ var serviceRulesArray = []svc{
 		"||bytecdn.cn^",
 		"||byteimg.com^",
 		"||ixigua.com^",
-		"||musemuse.cn^",
-		"||musical.ly^",
+		"||muscdn.com^",
+		"||bytedance.map.fastly.net^",
 	}},
 }
 


varunsridharan:
@Cybo1927 `musemuse.cn` and `musical` related domains are their CDN Domains that are used in Tiktok

ArchiveBase:
Well, currently contain 16 domains list for tiktok but none of them prevent autoplay android app videos unless block muscdn.com i guess might be different for ios.

Cybo1927:
I had AdGuard Home set to block TikTok and it didn't stop working until I added `musical.ly`. But, I believe `musemuse.cn` might be a tracker for it.

Tested on iPad Mini 2 on iOS 12.4.2

Note: Looking into blocking only nessicary domains

Cybo1927:
`muscdn.com` seems to block autoplay on the homepage but not the content in discover. Blocking `bytedance.map.fastly.net` seems to work for the content in discover. Updating accordingly.

ArchiveBase:
@Cybo1927 Sorry i don't own apple device then cannot help for verify that easily but blocking muscdn.com prevent load all videos not only homepage and 'discover' window contents in android app but 1 or 2 random image may appear just i dont take it serious. additionally i had to block dns.google.com, dns.google prior test above each domains separately

Cybo1927:
@ArchiveBase I get the problem with random images loading too, I can't figure out how to fix that. I blocked all the domains related to TikTok I could find and it happens every now and again.

What do you mean by
> additionally i had to block dns.google.com, dns.google prior test above each domains separately

ArchiveBase:
@Cybo1927
> I get the problem with random images loading too, I can't figure out how to fix that.

Ignore it because no one prefer to use app like that. Currently we can't block IPs.

> What do you mean by

Because it makes useless about block muscdn.com on tested android version and you said 'but not the content in discover. i thought that caused for it.

Cybo1927:
@ArchiveBase Kinda taking a guess here... you're saying `muscdn.com` isn't worth blocking?

Note: If you don't speak good English when talking to me you can use your primary language and I'll figure it out.

ArchiveBase:
@Cybo1927 Sorry for that :)

> you're saying muscdn.com isn't worth blocking?

No.  this's a separate issue then i will test it further in my free time or adguard may reproduce it

PR #1356, opened 2020-01-17
fix(Menu): Close menu on navigation

longebane:
The side menu remains open after selecting a link. This is especially problematic on mobile because you may accidentally select a toggle after navigating to another page and the menu is still open.


ArtemBaskal:
it's already done in 2bc2b342ebda896d7514acfd3c453494a1f93eba.

PR #1375, opened 2020-01-27
README -- plural word instead of singular in requirements

hmage:
More than one requirement, so instead of `it` use plural `them`.


szolin:
56b0dba1482165e1db67855921ddc2e09f5afc38

Issue #1517, opened 2020-03-26
Adguard "Non-filtering" DNS as Default


ZExplore:
> @ameshkov Well, we can consider using AdGuard DNS by default in one ane only case -- if we add an endpoint that does no ad & tracking filtering. But it kinda defeats the purpose, the idea of AdGuard DNS is to block this stuff:)

ameshkov:
Well, yeah, we can consider it now indeed :)

ZExplore:
v0.102?

ArchiveBase:
> v0.102?

Done.

ameshkov:
Nah, v1.0. Hopefully, by that time AGH will cover all necessary locations, currently, there are some blank spots.

PR #1524, opened 2020-03-28
Adguard DNS (non-filtering) as default

ArchiveBase:
https://github.com/AdguardTeam/AdGuardHome/issues/1517


ameshkov:
Let's wait a little bit, we're still not entirely happy with AG DNS coverage, we need more locations

Archive5:
That bad bot has more permissions to close PR.

> Let's wait a little bit, we're still not entirely happy with AG DNS coverage, we need more locations

I dont think that's a serious problem because user can try different providers in AGH otherwise that may take century

KizunaH:
Ping

KizunaH:
> > Let's wait a little bit, we're still not entirely happy with AG DNS coverage, we need more locations
> 
> I dont think that's a serious problem because user can try different providers in AGH otherwise that may take century

@ameshkov this time unable to remove stale bot "wontfix" label by any comment unless you manage it and the source feature request #1517 that had assigned to v1.0 already disappeared from repository.
I'm not prefer third-party services remain on adguard home or get advertised by this product while adguard has a same facility.

KizunaH:
Ah! stale bot removed the "wontfix" label. ;)

ameshkov:
Wait a bit, we're going to migrate AG DNS to new IP addresses, after that we may consider it again.

ainar-g:
The author has disappeared, and there seems to not have been any activity for almost a year.  Unless @ameshkov decides that we are ready to do something like this, I think it's better to close this.

PR #1584, opened 2020-04-14
add IoT device type

TXC:


ameshkov:
Hm, I am not sure about this one, we have `device_other` for unspecified decides already, and this looks unspecified.

TXC:
@ameshkov This tag would actually take `device_camera`and `device_alarm` under its name. But those can be of "sub tags". `device_other` can be a very large "category" when you have lots of different devices that call home.

I've lots of devices that can be divided into different tags:
* Cameras (both security and baby monitor) `device_camera`
* Temperature monitors
* Lights (Hue and Trådfri)
* Switches (Z-Wave and Hue/Trådfri)
* Alarm `device_alarm`
* Cars
* Voice Assistants (Google Home / Alexa / Apple Homepod)

These are alot of IoT devices, and i've over 80 IoT devices.

ameshkov:
> Voice Assistants (Google Home / Alexa / Apple Homepod)

Btw, these are rather widespread devices. Let's add `device_voice_assistant` as a separate tag

PR #1637, opened 2020-04-30
Fix #771, filter skips comment lines starting with '#'.

### LINKED ISSUE #771: Comments are counted as rules
### Steps to reproduce

1. Add `https://bestpika.github.io/abp/hosts.txt`

### Expected behavior

Rules count is zero.

### Actual behavior

Rules count show 1.

<details><summary>Screenshot</summary>

image
image

</details>

### Your environment

| Description | Value |
| - | - |
| Version of AdGuard Home server: | v0.95-hotfix
| How did you setup DNS configuration: | System
| If it's a router or IoT, please write device model: | PC
| Operating system and version: | Windows 10 x64/Windows 8.1 x64

DannyHinshaw:


home/filter.go
```diff
@@ -468,7 +468,8 @@ func (f *Filtering) parseFilterContents(file io.Reader) (int, uint32, string) {
 		checksum = crc32.Update(checksum, crc32.IEEETable, []byte(line))
 
 		line = strings.TrimSpace(line)
-		if len(line) == 0 {
```
ameshkov, line 471: Makes sense, but you should also ignore lines starting with `!`:
https://github.com/AdguardTeam/AdGuardHome/wiki/Hosts-Blocklists

DannyHinshaw: Sure thing, just didn't see that referenced on the og issue. Not a problem.

szolin: This is incorrect, because there's `if line[0] == '!'` check below.
And now we'll enter `if len(line) == 0 || isComment` if-statement.

What we should do is just add another check `if line[0] == '#' {` after `if line[0] == '!' {`

DannyHinshaw:
@szolin can't believe I missed that, thanks! Better now?

home/filter.go
```diff
@@ -472,7 +472,7 @@ func (f *Filtering) parseFilterContents(file io.Reader) (int, uint32, string) {
 			continue
 		}
 
-		if line[0] == '!' {
+		if line[0] == '!' || line[0] == '#' {
```
szolin, line 475: Unfortunately it won't work, because of `f.filterTitleRegexp = regexp.MustCompile(`^! Title: +(.*)$`)`
This regex expects only `!` at the beginning.

OK, so let's just make it simple, shall we?  Just add your check below like this:

```
if line[0] == '!' {
...
else if line[0] == '#'  {
// skip comment
} else if line[0] == '#'  {
			rulesCount++
		}
```

DannyHinshaw: My apologies, I do hope my latest change is what you're looking for.

DannyHinshaw:
How did that fail the Travis CI?

home/filter.go
```diff
@@ -478,6 +478,8 @@ func (f *Filtering) parseFilterContents(file io.Reader) (int, uint32, string) {
 				name = m[0][1]
 				seenTitle = true
 			}
+		} else if line[0] == '#' {
+			continue
```
szolin, line 482: I think the Travis build is failing because you've left the trailing white space after `continue`.  Can you remove it?

DannyHinshaw: @szolin it looks like that did it, thank you.

szolin:
Merged, thank you!

PR #1640, opened 2020-05-01
Correct architecture builds

popey:
The current syntax means it will build on one of the architectures to run on all. That would be useful if it was an arch-independent snap, such as a shell script. But this contains arch-specific binaries. Using the syntax here, will get four separate builds (one per arch) which is the desired outcome. More information in the docs at https://snapcraft.io/docs/architectures


ameshkov:
Thank you!

Issue #573, opened 2019-02-03
Run 2 Instances Of AdGuard Home For Redundancy


rashidjehangir:
Hi Guys, is it possible to run 2 instances of AdGuard Home on different PC's on the network for redundancy? I Can use the 2 DNS addresses in the router.

ameshkov:
Well, usually it is possible to configure multiple DNS servers in the router DHCP settings. However, one of them will be the primary server, and it will be used most of the time anyway.

iSmigit:
Think this should be reopened and made a feature request. Routers don't strictly have a primary and secondary DNS in the sense that ones a backup only server, it's just poor terminology that gets used around the place and whether a router favours one DNS server over another is up to implementation and vendor. Generally if multiple DNS servers are configured, they will both be used. Even if one was used more frequently than another, that leaves a percentage of requests going to a different DNS server and if thats not an Adguard server the security measures provided by this product are null and void for those requests.

There's various reasons to want to want to run multiple instances, with the main one being people simply may need to reboot a system from time to time. Either the router will have one DNS server configured which is expected to go down occasionally, or theres multiple and traffic 24/7 will be going to both. Ideally rather than the other DNS server being a public one, it could be another Adguard instances so irrespective of what server is used, the traffic is protected.

Right now thats possible no doubt, but any whitelist, blacklist etc etc will need to be manually applied to two servers. It'd be a great advantage if the web UIs allowed for synchronisation of config which would make managing multiple instances substantially easier and encourage more secure measures. FWIW this is the most upvoted, not implemented, feature over at Pi-Hole so there is demand and I suspect implementing it would win many a customer over.

ameshkov:
@iSmigit well, there's nothing that prevents running multiple instances of AGH as long as they have different configurations / listen ports.

iSmigit:
Thanks for the response. The subject title and my assumption is that the OP and myself would both like to see identically configures instances running on two systems. If one instance was to go down the endpoint or router could direct traffic to the other DNS server, which are both identically configured.

As you note this is already doable, but if you have config that’s not out of the box then you need to manually apply it to both instances. In particular adding filters means managing multiple instances where I think it would be awesome if the application itself could accept the host name of a second instance of the application and have them replicate the system config (filter rules) whenever a change is made to one instance.

Unsure if it’s work well if the DHCP service was enabled, but for pure DNS filtering I think being able to mirror config automatically would be awesome.

ameshkov:
[...]

PR #1649, opened 2020-05-05
Webhooks for syncing and automation

subdavis:
This is my attempt at a preliminary solution to #573.  That's a pretty popular issue, and I think it deserves to be addressed.

See my comment on the issue for more details: https://github.com/AdguardTeam/AdGuardHome/issues/573#issuecomment-623792666

Adds new config, example:

```yaml
webhooks:
- url: http://target.url/endpoint
  authorization: "Token iuaorfjaljinfilauhiolkjsnaliufvwqier"
  categories:
  - dns_config
  - dns_rewrite
  - filter
```

I've opened this issue to start a discussion with the maintainers, but if it makes sense to do something like this, we still need

- [ ] tests 
- [ ] documentation
- [ ] UI support (maybe we can push this off for now?) 

I've also created https://github.com/subdavis/adguardhome-sync as a proof-of-concept microservice to subscribe to these webhooks and propagate updates.  I think the cleverness of that project comes from using TypeScript together with [openapi-generator](https://github.com/OpenAPITools/openapi-generator) so that future API changes will be detectable at build-time, making it easy to maintain.

EDIT: removed DRAFT status to avoid confusion.  Even though this is a work in progress, I'm asking for feedback now.

### DIFF
--- a/dhcpd/dhcp_http.go
+++ b/dhcpd/dhcp_http.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/AdguardTeam/AdGuardHome/event"
 	"github.com/AdguardTeam/AdGuardHome/util"
 
 	"github.com/AdguardTeam/golibs/log"
@@ -92,7 +93,7 @@ func (s *Server) handleDHCPSetConfig(w http.ResponseWriter, r *http.Request) {
 		httpError(r, w, http.StatusBadRequest, "Invalid DHCP configuration: %s", err)
 		return
 	}
-	s.conf.ConfigModified()
+	s.conf.ConfigModified(event.DHCP)
 
 	if newconfig.Enabled {
 		staticIP, err := HasStaticIP(newconfig.InterfaceName)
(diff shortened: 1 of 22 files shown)


ameshkov:
Hi @subdavis, thank you for the PR, and sorry for the late reply and for messing it up by my refactoring:(

Could you please explain why do you need webhooks and how you're going to use it?

ameshkov:
Ah, my bad, I see why. Let me please think about it a little bit.

subdavis:
I lost interest, sorry.

FWIW, I think this was the correct approach. It doesn't make sense to have users pushing and pulling entire configuration files when all they wanted to do was make 2 instances enable and disable in unison or synchronize block lists.  It also doesn't really make sense to me to discard information about the event source: in all observer/observable patterns, event source is the most important piece of data, and removing it would really just result in users relying on other leaky abstractions to make a best guess at what the event source was.

Maybe PiHole, AdGuard Home, Blocky, or something else will eventually support this.

PR #1651, opened 2020-05-05
Update discord block for new domain

ZusorCode:
Discord now also uses discord.com as domain, this adds it to the blocked services list


mHatsune:
#1863

Issue #1203, opened 2019-11-24
Timed blocklists and whitelists


ArchiveBase:
#### Problem Description

Pi-hole has a option for disable filters for certain time (10 sec, 30 sec, 5 min, custom time). I know someone will request for copy paste the feature.

Regardless it, this feature useful for activate then leave, when we need to unblock specific domain list for perform client devices software updates within 10 min or allow a temp service ex: skype calls for 1hr (also possible to fulfill through https://github.com/AdguardTeam/AdGuardHome/issues/951 but not exactly)
_(Endpoint management software come with profile feature for fulfill this requirement)_

People also disable whole protection (all filters) when they often or randomly visit/use interested service but adguardhome bother for it (just annoying task add/remove custom filtering rules or maintain remote list) and forget to re-enable for long time then all devices expose to tracking, ads etc.

#### Proposed Solution

I think it's convenient to add separate column in filters page with special settings for manage each timed rules list.
https://github.com/AdguardTeam/AdGuardHome/issues/822 (whitelists feature)

If implemented, it can be use in two ways
- assign using entries https://github.com/AdguardTeam/AdGuardHome/issues/1081
- assign through UI https://github.com/AdguardTeam/AdGuardHome/issues/817#issuecomment-534275463 (tags) https://github.com/AdguardTeam/AdGuardHome/issues/435

ArchiveBase:
![Screenshot_22](https://user-images.githubusercontent.com/46082456/69730490-c058df00-115a-11ea-9a76-8022f3b53ec3.png)

ArchiveBase:
Add two type of configuration methods to 'Configure Time' field.
- Default (specific duration for activate manually)
- Recurring (scheduler) then it turn on/off automatically.
_you can check screenshot on https://github.com/AdguardTeam/AdGuardHome/issues/951#issuecomment-523906910_

Now AGH has blocklists/allowlists options and now users able to create local lists #817 as well. it'll be more easier to maintain when #1450 available. So we can stop adding new services to 'blocked services' component. We are really advertising those companies through AGH ;)

ExtraLog:
I voted this feature months ago. It seems take time for adguard's consideration so today decided to add comment to explain why we want this!. I use software firewalls for my home computers and it's headache to enable/disable rules for non frequently use tools and updates so on....
You guys know most popular managed security softwares are not available for free. i guess many people interest for scheduled or manual internet accesses and this feature has explained everything so we wait to use it.

[...]

PR #1652, opened 2020-05-06
add Disney Plus streaming

andre68723:
missing: Icon

### DIFF
--- a/client/src/helpers/constants.js
+++ b/client/src/helpers/constants.js
@@ -189,6 +189,10 @@ export const SERVICES = [
         name: 'Netflix',
     },
     {
+        id: 'disneyplus',
+        name: 'Disney Plus',
+    },
+    {
         id: 'snapchat',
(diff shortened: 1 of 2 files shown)


Archive5:
@ameshkov I dont like such services add to blocked services and not interest to see about copying and doing the same poor thing that nextdns does. It's good to have limitation eg: only add world most popular services that social friendly, privacy hostiles, adult friendly(that still unable to consider for parental service) etc.

Several days ago added a comment related that https://github.com/AdguardTeam/AdGuardHome/issues/1203#issuecomment-615553752

Archive5:
@andre68723 see https://github.com/AdguardTeam/AdGuardHome/issues/1692

szolin:
Is there any icon that we can add here?

uNitrous:
@subdavis @me two votes to stop add

andre68723:
1692 is even better from parental perspective

PR #1671, opened 2020-05-14
Add IntelliJ Project files (*.iml) to .gitignore

glocknerc:
Rename "Portuguese" to "Português" so it's localized like all other languages


ameshkov:
Already done

Issue #1692, opened 2020-05-17
Add possibility to define custom blocked services


alexruf:
The feature just quickly block & unblock popular sites and services in the general tab with a simple click is awesome.
As a father of children who sometimes overdo it with online gaming, it would be great if you could define your own custom services there to extend and customize the list of services.
I think of it like this:
You simply enter the name of your service and it creates a small separate block list associated with this service name. With a toggle button somewhere in the UI you now can simply enable enable and disable the blocking of these DNS entries.

In theory you could go even further and also add the ability to schedule times when those services are automatically blocked.
I think this could be a really useful feature.

Archive5:
>In theory you could go even further and also add the ability to schedule times when those services are automatically blocked.

#1203 #951

@ameshkov this feature had requested by someone past and closed due certain reason. Opened issue exist for #1605 too as i remember.
I forgot to say about this problem past when was my old account blocked. I had to beg them week for unblock temporally  (usually small reason against them or big brother enough rather than unacceptable or illegal things) 

Below two suggestions to use for adguard's closed source applications atleast :)

- https://getfider.com/ open source 
interface: https://feedback.fider.io
(will available better themes, may be add ways to sync issues, comments between github etc.)
- https://www.zoho.com/desk/ paid
interface: https://desk.zoho.com/portal/zeroalt/community
https://help.zoho.com/portal/en/community
(it's perfect but hard to hope such issue sync feature as above neither fider implement that easily)

szolin:
Is it really necessary to have a custom list of blocked services?

1. We have a compiled static list for the most popular services - it's much better to add more services there, because it will automatically help other people.  Plus our community will keep this list up to date.
2. If there are several domains you want to block for specific clients - you can use `$ctag` filters for that.  I agree it's not *that* simple as click on enable/disable button, but it will do exactly the same.

Archive5:
Well, +1 vote for this feature and due #1721 @eoopx

It's good to add explanation for 'blocked services' creation wizard. below some basic.

- necessary to explore all domains belongs to particular service and add otherwise it may not get blocked properly
- necessary to find all domains related desktop (windows, linux etc), mobile (android, iOS etc), consoles (xbox, nvidea), equipment (switches, smart devices) and hardware (alexa, google home) products otherwise some are may remain accessible
- necessary to check them regularly and update (eg, per month) to prevent occur unwanted accesses.

ameshkov:
[...]

PR #1775, opened 2020-06-07
Add qq to blocked services

ifurther:
Add the qq to blocked services

### DIFF
--- a/dnsfilter/blocked_services.go
+++ b/dnsfilter/blocked_services.go
@@ -161,6 +161,7 @@ var serviceRulesArray = []svc{
 		"||douyin.com^",
 		"||tiktokv.com^",
 	}},
+	{"qq", []string{"||qq.com^", "||qqzaixian.com^"}},
 }
 
 // convert array to map


Archive5:
https://github.com/AdguardTeam/AdGuardHome/issues/1692

ameshkov:
Bad bot. @ArtemBaskal please take care

ArtemBaskal:
Merged https://github.com/AdguardTeam/AdGuardHome/commit/32b24ce0930cbdb9b759f77f2c0773f3e61cb7b3

gtaphash:
@ameshkov  when block `qq.com` domain (qq app, mailbox, office suites) then it brake access to `wechat.com`, localized web browser/ updates `browser.qq.com`, desktop/security solutions updates `guanjia.qq.com` and other third-party sites depend on qq.
This blocked service cannot take control over all products or services compared to existing services also leave some applications vulnerable 🤔

ameshkov:
@gtaphash what do you suggest?

gtaphash:
> @gtaphash what do you suggest?

This may need to limit for `im.qq.com` (IM app) and `mail.qq.com` (mailbox) which means by qq but `qq.com` has local news site. otherwise add ⓘ button to explain about possible issues when block it in a network. Generally IM app site, root domain etc available in local language and said browser, desktop/security softwares so far use or purchase by local people although might has attraction for qq app or this service can remove according to one of my opinion and let people to add these type of services via custom blocked services https://github.com/AdguardTeam/AdGuardHome/issues/1692

ameshkov:
Well, we could do that I guess, but I think that would be kinda misleading. Offering to block QQ, but not even blocking its homepage is weird.

Let's maybe consider unblocking specifically the domain that breaks `wechat.com`? 

> and other third-party sites depend on qq.

Depend in which way? Any examples?

gtaphash:
>Let's maybe consider unblocking specifically the domain that breaks wechat.com?
>Depend in which way? Any examples?

Depend on `qq.com` CNAME and that looks like a kill-switch (PR seems come from taiwan, i don't know) 😄

>Let's maybe consider unblocking specifically the domain that breaks wechat.com?

Site: `weixin.qq.com` `wx.qq.com`(static) `dldir1.qq.com`(downloads) `akm.qq.com`(downloads)
other usage: `weixin.qq.com` `imtt.qq.com` maybe has other. sign-up stuck from my country (not ISP issue)

Google Play
QQ - 10 million installs (since 2011 sep release)
QQ mail - 1 million installs (since 2013 jan release)
Wechat - 100 million installs (since 2011 jan release)

It's much lower than telegram - 500 million installs

Or isn't it good to add 'WeChat' instead? 
Site available in english by default and ago USA wanted to block it

ameshkov:
> Depend on qq.com CNAME and that looks like a kill-switch (PR seems come from taiwan, i don't know) 😄

Oh, this makes sense.

> Or isn't it good to add 'WeChat' instead?

I guess we should add an option for WeChat as well.
Maybe Weibo as well? Baidu?

ameshkov:
> Google Play

Well, since Google Play does not work on the mainland, I don't think the numbers from there are reasonable.

Also, the point of "Blocked services" is to allow users quickly disabling this or that service on the per-client basis.
For instance, if you'd like to prevent kids from chatting/playing, you could do that for their devices.

gtaphash:
>Well, since Google Play does not work on the mainland, I don't think the numbers from there are reasonable.

It's about how much particular app popular worldwide and there're so many chinese people use skype(also companies), gmail(even it blocked), hotmail and i clearly know that there also 2-3k members contain chinese groups in telegram (mostly those are not counted). eg: baidu seems just a search engine in chinese language https://www.baidu.com/

>Or isn't it good to add 'WeChat' instead?
>>I guess we should add an option for WeChat as well.

so more, kik, imo, hike/jio/indian messenger, viber, icq and others
(actually need to have standard for blocked services part) eg: add only dominated services or lack privacy friendly services or add option to select services under particular categories (regardless popular or not)

there're world dominated services out there. people interest to use them, usually tend to despite or lack interest for own country things so i'm really not interested ***local friendly(qq, yandex messenger)** or ***privacy friendly(wire,telegram, signal)** services get added. 

As i said above, this feature extension #1692 may help for custom services and it's upto you to decide what to do and treat the product. ago i saw qq domain brake wechat and other sites, that was another reason for add a comment

KizunaH:
>(actually need to have standard for blocked services).... or select services under particular categories
Screenshot from 2020-10-24 20-40-53

KizunaH:
Above argument reasonable; regional focused apps often has low installation rate even they're accessible around the world.

Screenshot from 2020-10-25 19-5


Past i addressed an issue about add microsoft, amazon, reddit, google, cloudflare, yandex, origin and ignored 1* 4* 6* companies because of their products, adguardhome especially designed for home users (not for enterprises that aware about IT). Amazon partially blocked due their hosting platform  and yandex ignored due their web browser (also same for QQ browser and their offline usable tools)

Current blocked services page has 'block all' and 'unblock all' buttons on the top. It means for home users, Oh! let's block these hipop services with just one click

>Also, the point of "Blocked services" is to allow users quickly disabling this or that service on the per-client basis.
For instance, if you'd like to prevent kids from chatting/playing, you could do that for their devices.

Then need to implement feature similar below for advanced blocking(for only effect to each category services and cover worldwide solutions) while let to remaining existing blocked services component with custom option #1692, current behavior doesn't make sense. just kids find alternative in next minute

>(actually need to have standard for blocked services).... or select services under particular categories

Screenshot from 2020-10-24 20-40-53

ameshkov:
> (actually need to have standard for blocked services) eg: add only dominated services or lack privacy friendly services or add option to select services under particular categories (regardless popular or not)

The reason for blocking a service may not just be whether it's privacy-friendly or not.

Anyways, I filed a new issue so that we don't forget about that: https://github.com/AdguardTeam/AdGuardHome/issues/2224

@KizunaH
> Then need to implement feature similar below for advanced blocking

I kinda agree and disagree at the same time:)

IMO, it's hard for people to realize what actual services are behind these categories. It's just easier to understand what exactly is being blocked when you see the names and the logos.

KizunaH:
>Also, the point of "Blocked services" is to allow users quickly disabling this or that service on the per-client basis.
>For instance, if you'd like to prevent kids from chatting/playing, you could do that for their devices.

>>I kinda agree and disagree at the same time:)
>>
>>IMO, it's hard for people to realize what actual services are behind these categories. It's just easier to understand what exactly is being blocked when you see the names and the logos.

Just a comment. let me know if it good to implement and need to open feature request.

I can realize the point of view but this isn't a perfect solution for parental control or access prevention to certain category services at all (belongs to individual/kids/co-workers) unless let users to disable access for dominant internet companies.
The propose new solution will address that properly and adguardhome has lot of users already so they'll definitely help if missing something.
- block youtube but there're dtube, vimeo, rutube...
- block tiktok but there're chingari, dubsmash, triller...

The benefit: Each category will provide most effective coverage. eg: block IM apps then DNS server owner can unblock only the required app for server users to communicate (eg: skype) 

About social media:
May be okey to add weibo since far known as chinese social media but i don't recommend to add other regional services. Already has 10M+ adguardhome docker downloads. i dont want see this become one of another source that bring foreigners to regional sites and expand various disappointing troubles than today.
(Since this 'blocked services' component come from nextdns and only has facebook in it; later they started to offer special toggle option called 'social media' to cover rest of all services). it's similar above proposed solution

ameshkov:
What I mean is that ideally, we should allow both: blocking by categories AND/OR blocking a single popular service.

I haven't given this enough thought though.

PR #1803, opened 2020-06-14
Update README.md

gumadozucia:
Per client configuration was added in Pi-hole 5.0. 

I understand the reasoning behind not explaining that blocking of malware domains is possible in Pi-hole. However, people online, especially on reddit, are quick to call this comparison dishonest, due to the fact that blocklists added by the user make it possible.


ameshkov:
Makes perfect sense, thank you!

PR #1836, opened 2020-06-26
Add web launcher for desktop users

popey:
One of the common review complaints about the snap when installed locally on a desktop or laptop, is that once installed, there's no desktop icon in the menu/launcher to get to the application. Many novice users expect this, even for web applications. We added this type of launcher for other snaps and they're well received.

So this PR adds a simple shell script which gets the current bind_port (if set) and launches a browser via the simple `xdg-open`, and if the config doesn't exist yet, it'll revert to the default port 3000. I modified both sets of snapcraft.yaml, the one for edge in root and the one under the packaging folder. Hopefully that's all okay.  Tested here, and I get a nice Adguard Home icon in my launcher which just opens a browser window in my default desktop browser, as expected.


ameshkov:
Hmm, so people try to install it on their desktop machines?

In this case, there's another issue. Ubuntu's `systemd-resolved` by default is enabled and uses port 53. I reckon this is quite a problem for anyone who's trying to set up AGH on their desktop/laptop

ameshkov:
@popey I'll take some time to test this change on a laptop with Ubuntu, thank you!

I'd appreciate if you could answer [this question](https://github.com/AdguardTeam/AdGuardHome/pull/1836#issuecomment-650130632) :)

popey:
Sorry, I missed the question. I didn't have a problem with that, during the adguard setup page I just told it to bind to my external IP (the wifi IP) and it works fine.

Issue #1733, opened 2020-05-25
Extend rules support for TLD and SLD


Archive5:

ameshkov:
What's SLD?

Archive5:
Second-level domain (SLD or 2LD)

ameshkov:
Hm, I can't see where one would block `example.SLD` and not `example.TLD`

Archive5:
> Hm, I can't see where one would block `example.SLD` and not `example.TLD`

Can't understand.

`bigcompany.com`, `bigcompany.org`, `bigcompany.com.br`,` bigcompany.com.mx`

Archive5:
I only mentioned about it to avoid forget but you can implement it in your own way 😊

ameshkov:
I am inclined to implement just the TLD and do it in the following way:

If a domain name ends with `.TLD` (in the upper case) then it means that all registered domains matching the rule must be blocked

Example: `||example.TLD^` will block `example.com`, `example.com.br`, etc

Archive5:
It's good and only update wiki about this new rule because only suitable to silence large enterprises

DandelionSprout:
Blocking entire TLDs can already be done very easily with `||tk^` and `||co.uk^`

I am however not aware of a way to block second-level domains in any adblocker tool.

Archive5:
> Blocking entire TLDs can already be done very easily with `||tk^` and `||co.uk^`
> I am however not aware of a currently existing way to block second-level domains in any adblocker tool.

No. we mean about something.TLD then it applies all extensions

ameshkov:
> Blocking entire TLDs can already be done very easily with ||tk^ and ||co.uk^

I guess we're talking about rules like this:
`||example.TLD`

mHatsune:
Merge https://github.com/AdguardTeam/AdGuardHome/pull/1863 once completed

ameshkov:
I've tried quickly implementing it, but it appeared to be not as easy as I thought at first.

Implementing it properly so that it didn't hurt performance would take more time.

PR #1863, opened 2020-07-06
Update blocked services

KizunaH:
This covers user opened PR for blocked services recently


DandelionSprout:
Personally I'm not sure if applying `.TLD` to `||vk.`, `||6pm.` or `||novi.` is a good idea, due to how short they are and the possibility they're used for non-corp purposes in smaller countries.

I also think we need an expert on `Cloudflare` and Amazon` to look through this PR.

Other than that it looks pretty good to me.

szolin:
Looks like a pretty big change that can potentially break something.
Maybe it's better to apply just 1 smaller change at a time during several next releases? @ameshkov

ameshkov:
@szolin well, that's why we have "edge" update channel :)

KizunaH:
@DandelionSprout Sorry, i didnt noticed mail notifications.

>Personally I'm not sure if applying .TLD to ||vk., ||6pm. or ||novi. is a good idea, due to how short they are and the possibility they're used for non-corp purposes in smaller countries.

Updated. then let's wait for other hand (till someone say about certain ccTLD or gTLD not blocked)

>I also think we need an expert on Cloudflare and Amazon to look through this PR.

I checked properly then no problem with it except you want to add more domains because this list doesn't contain all amazon domains for example.

>@ameshkov @szolin well, that's why we have "edge" update channel :)

Does that already works without #1733 ? ;)

ameshkov:
@KizunaH 

> Does that already works without #1733 ? ;)

Nope, not implemented yet

szolin:
9b9902f0048e1bca5d0c55ff715fb226f6fbc6ac
@KizunaH Please note that we reverted the patch as it depends on #1733

PR #1968, opened 2020-07-29
TranslationError

wysnxzm:


KizunaH:
@wysnxzm You need to update through [crowdin](https://crowdin.com/project/adguard-applications/en#/adguard-home) otherwise that will get replaced once again with incorrect translation.

ameshkov:
Yeah, all translation changes must go through Crowdin

Issue #1985, opened 2020-08-02
Unable to build on Arch Linux


graysky2:
### Issue Details

* **Version of AdGuard Home server:**
  * 0.103.3
* **Operating system and version:**
  * Arch Linux

### Expected Behavior
Running `make` should successful build AdGuardHome

### Actual Behavior
Make depends installed on Arch beyond standard toolchain are: `go nodejs npm git golangci-lint`

Downloaded and extracted the sourcetarball: https://github.com/AdguardTeam/AdGuardHome/archive/v0.103.3.tar.gz
```
% cd AdGuardHome-0.103.3
% make init
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
git config core.hooksPath .githooks
fatal: not in a git directory
make: *** [Makefile:99: init] Error 128
```

If I run `git init` in the untarred source first, it gets a bit further but errors out:
```
% cd AdGuardHome-0.103.3
% git init
Initialized empty Git repository in /build/adguardhome/src/AdGuardHome-0.103.3/.git/

% make init
fatal: No names found, cannot describe anything.
fatal: No names found, cannot describe anything.
git config core.hooksPath .githooks

% make
atal: No names found, cannot describe anything.
fatal: No names found, cannot describe anything.
npm --prefix client ci
npm --prefix client run build-prod

> dashboard@0.1.0 build-prod /build/adguardhome/src/AdGuardHome-0.103.3/client
> cross-env BUILD_ENV=prod webpack --config webpack.prod.js

sh: cross-env: command not found
npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! dashboard@0.1.0 build-prod: `cross-env BUILD_ENV=prod webpack --config webpack.prod.js`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the dashboard@0.1.0 build-prod script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
npm ERR!     /build/.npm/_logs/2020-08-02T11_26_32_313Z-debug.log
make: *** [Makefile:107: client] Error 1
make: *** Waiting for unfinished jobs....

[...]

PR #1989, opened 2020-08-03
update Readme based on learning from #1985

graysky2:
Based on the time we spent trouble shooting in #1985, this can help others.  Also note that I removed the requirement for golangci-lint since the make process works just fine without it.  Perhaps it is needed for something else?

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,10 @@ Run `make init` to prepare the development environment.
 You will need this to build AdGuard Home:
 
  * [go](https://golang.org/dl/) v1.14 or later.
- * [node.js](https://nodejs.org/en/download/) v10 or later.
+ * [node.js](https://nodejs.org/en/download/) v10.16.2 or later.
+ * [npm](https://www.npmjs.com/) v6.14 or later.
+
+Optionally, for Go devs:
  * [golangci-lint](https://github.com/golangci/golangci-lint)
  
 ### Building


README.md
```diff
@@ -155,8 +155,8 @@ Run `make init` to prepare the development environment.
 You will need this to build AdGuard Home:
 
  * [go](https://golang.org/dl/) v1.14 or later.
- * [node.js](https://nodejs.org/en/download/) v10 or later.
- * [golangci-lint](https://github.com/golangci/golangci-lint)
```
szolin, line 159: This is needed for Go devs (a code linting tool)

README.md
```diff
@@ -155,8 +155,8 @@ Run `make init` to prepare the development environment.
 You will need this to build AdGuard Home:
 
  * [go](https://golang.org/dl/) v1.14 or later.
- * [node.js](https://nodejs.org/en/download/) v10 or later.
```
szolin, line 158: Well, I have node v10.16.2 and it works fine.

README.md
```diff
@@ -165,7 +165,7 @@ Open Terminal and execute these commands:
 ```bash
 git clone https://github.com/AdguardTeam/AdGuardHome
 cd AdGuardHome
-make
+make -j1
```
szolin, line 168: We should try to fix the Makefile first.  I'll post an update here on this.

graysky2:
I adjusted and rebased the PR based on your feedback above.  We can still wait on the `make -j1` part as you see fit.

README.md
szolin: Resolved in 5188da60cfea2b1018e73dbf4d4d1b1342906352

graysky2:
I verified that ( 5188da6) works on Arch's build system.  Rebased.

PR #2035, opened 2020-08-22
Update vetted blocklists

KizunaH:
- Removed duplicated adaway (adaway and adaway default blocklist) and applied same domain instead github (adaway.org domain has in `home_test.go`)
- Updated barblock domain (check repository readme)


client/src/helpers/filters/filters.json
```diff
@@ -24,11 +24,11 @@
             "homepage": "https://github.com/AdguardTeam/AdGuardSDNSFilter",
             "source": "https://adguardteam.github.io/AdGuardSDNSFilter/Filters/filter.txt"
         },
-        "adaway-default-blocklist": {
-            "name": "AdAway default blocklist",
+        "adaway": {
+            "name": "AdAway",
```
ameshkov, line 28: I think the name was okay, AdAway is the name of the product, not blocklist

ameshkov (changes requested):

client/src/helpers/filters/filters.json
KizunaH: Name updated

ameshkov:
@KizunaH if you don't mind, could you please recreate this PR with one commit instead of 10? I'd like to merge it on our side and keep you as the author, but with 10 commits we'd have to do a squash commit that will override authorship

KizunaH:
> @KizunaH if you don't mind, could you please recreate this PR with one commit instead of 10? I'd like to merge it on our side and keep you as the author, but with 10 commits we'd have to do a squash commit that will override authorship

That happen when update repo through web interface even created separate branch. i couldn't find anyway to avoid it then check new PR

PR #2047, opened 2020-08-25
Update vetted blocklists

KizunaH:


ameshkov:
Hm, it seems this is obsolete, everything is fixed

PR #2081, opened 2020-09-07
Support  system service options

dsheets:
I found that when running AGH with `-s run`, no other command line options were respected. This behavior made it difficult to specify custom working directory, configuration file, pid file, update check disablement, or GL-Inet emulation mode (whatever that is) when running AGH as a service.

This PR is organized into 3 commits:
1. Factor the CLI argument parsing out of `home/home.go` into `home/options.go` and make it generally side-effect free; add some tests
2. Provide a pseudo-inverse serialization operation from an options struct back to CLI arguments (roundtrip is idempotent and canonicalizing)
3. Integrate serialization into the service system so that `-s install` will create a service using whatever other CLI arguments are passed to it and `-s run` will actually use other passed CLI args rather than silently ignoring them.

I'm happy to change names or reorganize tests or whatever you would like to have happen before this change is acceptable to merge. I don't have a great idea to easily test the service integration but it is relatively straightforward to inspect manually for correctness. I am running this patchset now.

### DIFF
--- a/home/options.go
+++ b/home/options.go
@@ -40,6 +40,33 @@ type arg struct {
 	updateWithValue func(o options, v string) (options, error)         // the mutator for arguments with parameters
 	updateNoValue   func(o options) (options, error)                   // the mutator for arguments without parameters
 	effect          func(o options, exec string) (f effect, err error) // the side-effect closure generator
+
+	serialize func(o options) []string // the re-serialization function back to arguments (return nil for omit)
+}
+
+// {type}SliceOrNil functions check their parameter of type {type}
+// against its zero value and return nil if the parameter value is
+// zero otherwise they return a string slice of the parameter
+
+func stringSliceOrNil(s string) []string {
+	if s == "" {
+		return nil
+	}
+	return []string{s}
+}
+
+func intSliceOrNil(i int) []string {
+	if i == 0 {
+		return nil
+	}
+	return []string{strconv.Itoa(i)}
+}
+
+func boolSliceOrNil(b bool) []string {
+	if b {
+		return []string{}
+	}
+	return nil
 }
 
 var args []arg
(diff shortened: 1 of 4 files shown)


home/options.go
```diff
@@ -0,0 +1,255 @@
...
+	"Port to serve HTTP pages on",
+	"port", "p",
+	func(o options, v string) (options, error) {
+		var err error
+		var p int
+		minPort, maxPort := 0, 1<<16-1
+		if p, err = strconv.Atoi(v); err != nil {
+			err = fmt.Errorf("port '%s' is not a number", v)
+		} else if p < minPort || p > maxPort {
+			err = fmt.Errorf("port %d not in range %d - %d", p, minPort, maxPort)
```
dsheets, line 107: This could probably be done with `ParseInt` in a better/shorter/neater way. Let me know if you'd like that.

ameshkov:
So the idea is to let users specify command-line arguments when calling `-s install`, and these arguments will be then used by the service?

ameshkov:
Generally, the PR looks very good, we'll need some time to test it though

dsheets:
Yes, the idea is that `-s install` carries the additional arguments with it into the service description and `-s run` carries whatever arguments occur next to it into the async service invocation. This functionality may want some documentation but the docs don't currently specify that this _doesn't_ work (and I assumed it did) so it may be fine.

Of course, I understand your desire to test thoroughly -- I didn't have any great ideas on how to do that or the familiarity with all of the options and platforms necessary to test live deployments. I hope integration doesn't take up too much of your time -- I'd really like you to keep working on the stuff you have planned :-).

Thanks for running a good project!

ameshkov:
Well, it looks definitely better than what we had before. Tomorrow, it'll be available in the "edge" channel build.

PR #2100, opened 2020-09-13
fix typo th.json

sopanatx:


ameshkov:
Thank you! But please do this via Crowdin: https://github.com/AdguardTeam/AdGuardHome#translate

Translations are synced from there.

PR #2156, opened 2020-10-06
Update DNS examples

KizunaH:
Update rest of examples to AG DNS not only quic protocol address

### DIFF
--- a/client/src/components/Settings/Dns/Upstream/Examples.js
+++ b/client/src/components/Settings/Dns/Upstream/Examples.js
@@ -8,10 +8,10 @@ const Examples = (props) => (
         <Trans>examples_title</Trans>:
         <ol className="leading-loose">
             <li>
-                <code>9.9.9.9</code> - {props.t('example_upstream_regular')}
+                <code>94.140.14.14</code> - {props.t('example_upstream_regular')}
             </li>
             <li>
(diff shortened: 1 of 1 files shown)


KizunaH:
@ArtemBaskal why that commit seems deleted from master?

ameshkov:
@KizunaH our bad, we should've merged it to internal repo first, otherwise it gets overwritten.

Could you please create a new one. Just one note - use `dns-unfiltered.adguard.com` in examples.

KizunaH:
> Could you please create a new one. Just one note - use `dns-unfiltered.adguard.com` in examples.

@ameshkov it already added to #2157 and this's about examples. it can be adguard or other provider according to user's choice. Everyone use `dns.provider.com` style subdomain then `dns-unfiltered` name may feel complicated to non-technical user

Should i use `dns-unfiltered.adguard.com` for examples?

ameshkov:
Tbh, I think you should. People will still copy/paste it from the example, and they'd better copy/paste the unfiltered version

KizunaH:
> @KizunaH our bad, we should've merged it to internal repo first, otherwise it gets overwritten.

See https://github.com/AdguardTeam/AdGuardHome/pull/2188 Can't PR again 🤗

PR #2157, opened 2020-10-06
Update DNS servers

Aikatsui:
Update to adguard non-filtering DNS (close #1524)

### DIFF
--- a/dnsforward/dnsforward.go
+++ b/dnsforward/dnsforward.go
@@ -27,9 +27,9 @@ const (
 )
 
 var defaultDNS = []string{
-	"https://dns10.quad9.net/dns-query",
+	"https://dns-unfiltered.adguard.com/dns-query",
 }
-var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "2620:fe::10", "2620:fe::fe:10"}
+var defaultBootstrap = []string{"94.140.14.140", "94.140.14.141", "2a10:50c0::1:ff", "2a10:50c0::2:ff"}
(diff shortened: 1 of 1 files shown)


ameshkov:
We'll need to change much more than that in order to change this - check all the places where Quad servers are used in the code. Also, we'll need to change strings and translations.

KizunaH:
> We'll need to change much more than that in order to change this - check all the places where Quad servers are used in the code. 

It seems no other places except https://github.com/AdguardTeam/AdGuardHome/blob/master/scripts/querylog/test/querylog.json

>Also, we'll need to change strings and translations.

Since recent versions, UI doesn't contain any details about default servers.

ameshkov:
> Since recent versions, UI doesn't contain any details about default servers.

Quad9 is used in examples, I guess it makes sense to replace it there as well.

KizunaH:
> Quad9 is used in examples, I guess it makes sense to replace it there as well.

I updated it seperately #2156

KizunaH:
@ameshkov it take time to review commits history. maybe @szolin or @ArtemBaskal has removed it but why removed "If you keep this field empty. AdGuardHome will use [provider] as upstream" line.
I see it's possible to leave bootstrap field empty and use DOH address in upstream or upstream field empty and use boostrap address so i guess it's necessary to explain what happen when user keep particular field empty

"If you keep upstream or bootstrap field empty, AdGuardHome will use [provider] if necessary" you can suggest a beautiful sentence

PR #2165, opened 2020-10-07
ipset wildcard match support (dnsmasq parity)

dsheets:
I started using the ipset functionality to grant restricted internet access to a smart projector and found that it did not behave as dnsmasq's ipset functionality. Specifically, `man dnsmasq` for `--ipset` states:

>  Domains and subdomains are matched in the same way as --address.

`man dnsmasq` for `--address` states:

> so --address=/example.com/ is equivalent to --server=/example.com/ and returns NXDOMAIN for example.com and all its subdomains.

and indeed we find at <http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob;f=src/forward.c;hb=f60fea1fb0a288011f57a25dfb653b8f6f8b46b9#l588>:

```c
 591       /* Similar algorithm to search_servers. */
 592       struct ipsets *ipset_pos;
 593       unsigned int namelen = strlen(daemon->namebuff);
 594       unsigned int matchlen = 0;
 595       for (ipset_pos = daemon->ipsets; ipset_pos; ipset_pos = ipset_pos->next) 
 596         {
 597           unsigned int domainlen = strlen(ipset_pos->domain);
 598           char *matchstart = daemon->namebuff + namelen - domainlen;
 599           if (namelen >= domainlen && hostname_isequal(matchstart, ipset_pos->domain) &&
 600               (domainlen == 0 || namelen == domainlen || *(matchstart - 1) == '.' ) &&
 601               domainlen >= matchlen) 
 602             {
 603               matchlen = domainlen;
 604               sets = ipset_pos->sets;
 605             }
 606         }
```

Prior to this patchset, AGH was matching exactly and only the domain name passed in the ipset configuration setting and provided no way to specify the domain as a wildcard domain. One could argue that all topological operations should be available for this functionality (i.e. 'boundary domain', 'open set below domain', and 'closed set below domain') but dnsmasq does not offer that functionality and I do not require it for my deployment so it's out of scope.

In updating this behavior of `ipset`, I've refactored `dnsforward/ipset.go` slightly to make it easier to test by moving the side effects into a parameter function. I then expanded the tests and caught a segfault on some edge cases for DNS responses (missing question and/or answer sections... I'm not sure these arise in practice).

I have found some further issues that I have not included in this PR as I think they require a fundamental rethink of the way AGH interacts with ipsets. Specifically, invoking the `ipset` command is slow, the stderr is thrown away when an error occurs (and stdout which is logged is empty), there is no easy way to distinguish between IPv4 and IPv6 ipsets except for parsing `ipset list` (and this isn't done currently), so if you have both IPv4 and IPv6 ipsets, the logs will be full of error messages about bad `ipset add` invocations. Finally, because of the discarded stderr, when IPv4 `ipset add` commands occasionally fail, there is no way to know why.

Rather than trying to address these issues in the current design or doing the work myself before I submit it to you for review, I propose that AGH adopt https://github.com/ti-mo/netfilter and interface with a netlink kernel socket directly in pure go. I'm happy to reimplement the current `ipset add` functionality in terms of that library if you'd be willing to accept the new dependency.

Finally, please let me know if you'd like issues opened for any of the bugs or my dependency proposal. I am currently running this patchset and it is successfully keeping my projector from talking to anyone except NetFlix :-).

### DIFF
--- a/dnsforward/ipset.go
+++ b/dnsforward/ipset.go
@@ -89,6 +89,31 @@ func (c *ipsetCtx) getIP(rr dns.RR) net.IP {
 	}
 }
 
+// Find the ipsets for a given host (accounting for subdomain wildcards)
+func (c *ipsetCtx) getIpsetNames(host string) ([]string, bool) {
+	var ipsetNames []string
+	var found bool
+
+	// search for matching ipset hosts starting with most specific subdomain
+	i := 0
+	for i != -1 {
+		host = host[i:]
+
+		ipsetNames, found = c.ipsetList[host]
+		if found {
+			break
+		}
+
+		// move slice up to the parent domain
+		i = strings.Index(host, ".")
+		if i != -1 {
+			i++
+		}
+	}
+
+	return ipsetNames, found
+}
+
 func addToIpset(host string, ipsetName string, ipStr string) {
 	code, out, err := util.RunCommand("ipset", "add", ipsetName, ipStr)
 	if err != nil {
@@ -115,7 +140,7 @@ func (c *ipsetCtx) processMembers(ctx *dnsContext, addMember func(string, string
 	host := req.Question[0].Name
 	host = strings.TrimSuffix(host, ".")
 	host = strings.ToLower(host)
-	ipsetNames, found := c.ipsetList[host]
+	ipsetNames, found := c.getIpsetNames(host)
 	if !found {
 		return resultDone
 	}
--- a/dnsforward/ipset.go
+++ b/dnsforward/ipset.go
@@ -122,15 +122,17 @@ func (c *ipsetCtx) processMembers(ctx *dnsContext, addMember func(string, string
 
 	log.Debug("IPSET: found ipsets %v for host %s", ipsetNames, host)
 
-	for _, it := range ctx.proxyCtx.Res.Answer {
-		ip := c.getIP(it)
-		if ip == nil {
-			continue
-		}
+	if ctx.proxyCtx.Res != nil {
+		for _, it := range ctx.proxyCtx.Res.Answer {
+			ip := c.getIP(it)
+			if ip == nil {
+				continue
+			}
 
-		ipStr := ip.String()
-		for _, name := range ipsetNames {
-			addMember(host, name, ipStr)
+			ipStr := ip.String()
+			for _, name := range ipsetNames {
+				addMember(host, name, ipStr)
+			}
 		}
 	}
 
--- a/dnsforward/ipset.go
+++ b/dnsforward/ipset.go
@@ -89,8 +89,22 @@ func (c *ipsetCtx) getIP(rr dns.RR) net.IP {
 	}
 }
 
-// Add IP addresses of the specified in configuration domain names to an ipset list
-func (c *ipsetCtx) process(ctx *dnsContext) int {
+func addToIpset(host string, ipsetName string, ipStr string) {
+	code, out, err := util.RunCommand("ipset", "add", ipsetName, ipStr)
+	if err != nil {
+		log.Info("IPSET: %s(%s) -> %s: %s", host, ipStr, ipsetName, err)
+		return
+	}
+	if code != 0 {
+		log.Info("IPSET: ipset add:  code:%d  output:'%s'", code, out)
+		return
+	}
+	log.Debug("IPSET: added %s(%s) -> %s", host, ipStr, ipsetName)
+}
+
+// Compute which addresses to add to which ipsets for a particular DNS query response
+// Call addMember for each (host, ipset, ip) triple
+func (c *ipsetCtx) processMembers(ctx *dnsContext, addMember func(string, string, string)) int {
 	req := ctx.proxyCtx.Req
 	if !(req.Question[0].Qtype == dns.TypeA ||
 		req.Question[0].Qtype == dns.TypeAAAA) ||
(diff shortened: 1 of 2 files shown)


ameshkov:
Thank you for the pull request!

> Rather than trying to address these issues in the current design or doing the work myself before I submit it to you for review, I propose that AGH adopt https://github.com/ti-mo/netfilter and interface with a netlink kernel socket directly in pure go. 

Yeah, it makes perfect sense to me.

> I'm happy to reimplement the current ipset add functionality in terms of that library if you'd be willing to accept the new dependency.

We'll be happy to, thank you!

ainar-g:
<p>Hi,&#xA0;and&#xA0;thank you for&#xA0;your PR!&#x2003;I&#xA0;will try
to&#xA0;review it&#xA0;shortly.</p>

dnsforward/ipset.go
```diff
@@ -89,10 +89,49 @@ func (c *ipsetCtx) getIP(rr dns.RR) net.IP {
 	}
 }
 
-// Add IP addresses of the specified in configuration domain names to an ipset list
-func (c *ipsetCtx) process(ctx *dnsContext) int {
+// Find the ipsets for a given host (accounting for subdomain wildcards)
```
ainar-g, line 92: It's better to start doc comments for unexported identifiers just like one would for exported, that is:

```go
// getIpsetNames finds the ipstes […]
```

That applies to the other doc comments below.

dnsforward/ipset.go
```diff
@@ -89,10 +89,49 @@ func (c *ipsetCtx) getIP(rr dns.RR) net.IP {
...
+func (c *ipsetCtx) getIpsetNames(host string) ([]string, bool) {
+	var ipsetNames []string
+	var found bool
```
ainar-g, line 95: Seems like a perfect place to use named returns, like we recommend in our recently-added `HACKING.md` file.

dnsforward/ipset.go
```diff
@@ -89,10 +89,49 @@ func (c *ipsetCtx) getIP(rr dns.RR) net.IP {
...
+
+func addToIpset(host string, ipsetName string, ipStr string) {
+	code, out, err := util.RunCommand("ipset", "add", ipsetName, ipStr)
+	if err != nil {
+		log.Info("IPSET: %s(%s) -> %s: %s", host, ipStr, ipsetName, err)
+		return
+	}
+	if code != 0 {
+		log.Info("IPSET: ipset add:  code:%d  output:'%s'", code, out)
+		return
```
ainar-g, line 125: We're trying to switch away from the practice of logging errors at the error site and return a properly wrapped error instead.  So perhaps `addToIpset` should return an error.

dnsforward/ipset.go
```diff
@@ -89,10 +89,49 @@ func (c *ipsetCtx) getIP(rr dns.RR) net.IP {
...
+	}
+	log.Debug("IPSET: added %s(%s) -> %s", host, ipStr, ipsetName)
+}
+
+// Compute which addresses to add to which ipsets for a particular DNS query response
+// Call addMember for each (host, ipset, ip) triple
+func (c *ipsetCtx) processMembers(ctx *dnsContext, addMember func(string, string, string)) int {
```
ainar-g, line 132: It's better to make the type of `addMember` a newtype.  And also document, at the site of the newtype's declaration, what the arguments are.

dnsforward/ipset_test.go
```diff
@@ -1,21 +1,100 @@
...
+
+type Binding struct {
+	host  string
+	ipset string
+	ipStr string
+}
+
+var b map[Binding]int
+
+func setup() {
```
ainar-g, line 24: We're trying to eschew global variables as much as possible.  Could you please rewrite the tests using the subtest feature instead?

dnsforward/ipset_test.go
```diff
@@ -1,21 +1,100 @@
...
+		},
+	}
+}
+
+func makeReqA(fqdn string) *dns.Msg {
+	return makeReq(fqdn, dns.TypeA)
+}
+
+func makeReqAAAA(fqdn string) *dns.Msg {
+	return makeReq(fqdn, dns.TypeAAAA)
```
ainar-g, line 61: I don't think that adding another level of indirection makes any difference here.  If anything, I would say that it hurts readability.  Perhaps inline the body of `makeReq` here and above?

dnsforward/ipset_test.go
```diff
@@ -1,21 +1,100 @@
...
+}
+
+func addToBindings(host string, ipset string, ipStr string) {
+	binding := Binding{host, ipset, ipStr}
+	count := b[binding]
+	b[binding] = count + 1
+}
+
+func doProcess(t *testing.T) {
+	assert.Equal(t, resultDone, c.processMembers(ctx, addToBindings))
```
ainar-g, line 92: Again, I don't think this indirection is necessary or helpful.

ainar-g (changes requested):

ainar-g:
Sorry for the long response.  We'll probably go with #2179.  I'll try to review, improve, and merge it this week.

ihipop:
seems to be merged at
 https://github.com/AdguardTeam/AdGuardHome/commit/510573a904e0fd184d6a2ccedf9ebe7b7ed81465
?

PR #2179, opened 2020-10-12
ipset netlink rewrite

dsheets:
This PR supersedes #2165 and implements the proposed rewrite of the ipset functionality on top of netlink sockets.

This change improves 'difficult' (10 addresses to be inserted) request performance by 50x, removes the need for root permissions to invoke the `ipset` executable (replacing it with a new requirement for `CAP_NET_ADMIN`), quiets the logspam from ipv4/ipv6 ipset mismatches, fixes a couple of semantics issues, and fixes a crashing race condition. Additionally, this changeset paves the way for some exciting future ipset-based features and, as far as I know, makes AGH the best choice for personal DNS-based firewalls.

Unfortunately, nothing is free. I've introduced some new dependencies, made the ipset tests harder to run, and I don't know what the cross-platform story is:

1. **New dependencies** [mdlayher/netlink](https://github.com/mdlayher/netlink), [ti-mo/netfilter](https://github.com/ti-mo/netfilter), and [digineo/go-ipset](https://github.com/digineo/go-ipset) are all necessary for this functionality. [vishvananda/netns](https://github.com/vishvananda/netns) has been added to the tests in order to run the ipset creation/query/addition/destruction operations in a network namespace to keep the root network namespace clean. As creating a network namespace requires root permissions, I'm not sure this is worth the extra hassle and I'm considering removing this functionality.

2. **Integration tests** I converted all of the ipset tests to integration tests now as they do I/O during initialization to check whether the referenced ipsets exist and what their properties are. Various things could be done to _unintegrate_ the tests so that at least some of them can be run without `CAP_NET_ADMIN` and without risking leaking test ipsets into the user's ipset namespace. I don't really know what to do here as I don't know what your test infrastructure looks like. I think at the least there will need to be some kind of privileged test run in CI that uses the appropriate build flags to pick up the tests.

3. **Cross-platform** I have no idea... I only have access to Linux.

Please let me know how you'd like to proceed. I'm happy to do some more work to get this accepted. I haven't tested this in 'production' yet but that will probably happen in the next couple days.

Thanks!


ameshkov:
> replacing it with a new requirement for CAP_NET_ADMIN

Hm, I will add this to Wiki then.

> I have no idea... I only have access to Linux

We'll see how Travis tests go. In theory, ipset functionality was not supposed to work outside of Linux, we just need to exclude it properly.

ainar-g:
Apologies for the long wait.  We've rebased your changes and mostly checked them in.  We'll readapt your integration test in a following change, probably sometime during the v0.106.0 cycle.

PR #2182, opened 2020-10-13
ipset comment and timeout support

dsheets:
This patchset adds support for AGH to populate the 'comment' and 'timeout' fields of address entries with the Q domain name and RR TTL respectively. Should the operator provide ipsets with these features on, they will be used. If either or both are not on (default), nothing different happens.

The comment feature allows the operator to audit their AGH ipsets to understand which domain resolution caused each address to be inserted.

The timeout feature allows the operator to construct a DNS-based firewall such that routing policy can be lifted entirely into the DNS namespace. That is, with appropriate firewall rules, the timeout feature enables a LAN administrator to prohibit connections that do not originate via an administrator-controlled DNS server -- DoH, DoQ, DoTLS and friends can be entirely shutdown for purposes of address resolution. This is exactly what one desires when sequestering 'smart' things on VLANs -- no smuggling of IP addresses or extraneous caching or reuse of addresses allowed. This functionality also means that operator ipsets do not grow without bound. A 1s extra slop period has been added to the TTL for the ipset entry timeout.

These features take AGH's ipset support beyond dnsmasq's.

This patchset builds on #2179 but requires digineo/go-ipset#6 which is still pending review.


ameshkov:
Awesome, thank you!

I am going to do a thorough review a little bit later - I'd like to release v0.104 before merging it.

dsheets:
You may want to cherry-pick the wildcard stuff and/or the race condition fix or back out the ipset feature entirely if you're about to release. If you let me know what you'd like in the release, I can prepare a modified ipset feature with that in it.

ameshkov:
Hmm, I guess we need to solve this. I'll take a closer look at #2165 so that we could at least solve the race issue.

dsheets:
I think https://github.com/AdguardTeam/AdGuardHome/pull/2182/commits/81abd6e13cb8101e98c619b8c1c7a726dedc0c2e is what you need but I only discovered the race late so it's in #2179.

ameshkov:
@dsheets sorry for the late reply, I've been really busy with adblockerdevsummit so I only had some free time to do my dnscrypt implementation.

Meanwhile, I just upgraded urlfilter to the new version with your changes and fixed the race condition: https://github.com/AdguardTeam/AdGuardHome/commit/0b43bf4cd9c1cfafd906593bc75b415f8e1c2b6f https://github.com/AdguardTeam/AdGuardHome/commit/e37ccdbb702c537dfba4538274949fbacac7ab99

Hopefully, we'll finally release v0.104 now and get to the other changes.

dsheets:
That's great, thank you! Everything on my end is groovy so don't feel the need to rush on my behalf. I'm happily running with my patches now and enjoying better control over my IoT subnet. I'm also drafting some blog posts about it all for consumption by HN or the twitterati or whomever but that is a bit slow so it's OK for it to be awhile before the next release with the changes (or revisions thereof).

I'm also contemplating ipset feature downgrading so if a user e.g. swaps a timeout+comment ipset for a timeout-only ipset, AGH detects it, logs the downgrade, but still inserts. Right now, it just logs the insertion error and carries on but that's a bit silly.

I don't know what's going on with digineo/go-ipset#6... I might just fork so its delay doesn't block this PR.

Do you think you'll release my patches before the new year? Any particular suggestions to consider for any blog posts I do?

ameshkov:
> Do you think you'll release my patches before the new year?

I really, really hope so. There's been some changes in the team, and we'll now be able to devote more time to AGH development so things should go faster. Also, v0.105 is supposed to be a relatively small update, the most important feature there is DNSCrypt server-side support, but 95% of the work is already done since it was added to [dnsproxy](https://github.com/AdguardTeam/dnsproxy/commit/39abe8047a59281128fa40ef879061ce25792d66) already.

> Any particular suggestions to consider for any blog posts I do?

Well, it's your blog posts, I guess you should tell what's important to you, not me:)

But if you ask what's important to me, then I'd say these points:

1. DNS encryption on both ends, I find these features important.
2. Flexible filtering rules. Yes, DNS is not as powerful as a proxy, but this does not mean we should limit ourselves to using hosts files.

Also if you need any assistance, please let me know, I'll be happy to help.

ainar-g:
Hello and sorry for the wait.  I can see that the upstream feature hasn't been merged yet.  That means we'll probably not be able to merge this in the v0.105 cycle, apologies.

dsheets:
The upstream maintainers seem to be AWOL so forking with the patch applied may be the most expedient option.

ihipop:
> The upstream maintainers seem to be AWOL so forking with the patch applied may be the most expedient option.

@dsheets @ainar-g 

suggest to switch to https://github.com/gonetx/ipset with full featured ipset binding，including  comment and ttl out of the box

ihipop:
Suggest use https://github.com/vishvananda/netlink/blob/master/ipset_linux.go for full featured ipset library
- it use netlink,so has no depend on ipset binary or library
- it use netlink,so it's fast enough for ADG useage
- it full featured for ADG,including the support for timeout and ttl

ainar-g:
Thanks, and sorry for the long wait time.  Unfortunately, we still have a lot on our hands at the time.  I still hope to review, improve, and merge this in the v0.107 cycle.  Again, apologies.

PR #2283, opened 2020-11-08
+ dnsfilter, client: Add MS telemetry to blocked services

devinslick:
I'm not sure about the SVG generation process for Icons.js, so I'd appreciate a hint on how to go about that correctly.

Blocking Microsoft FQDNs could introduce some risk of blocking their other services, so I understand if you'd rather not add MS as an option.   Either way, I'd like to understand how you went about creating the SVGs.

The FQDNs I've suggested in this PR should probably be supplemented with some from this list:
https://github.com/crazy-max/WindowsSpyBlocker/blob/master/data/dnscrypt/spy.txt

What do you think?


KizunaH:
@devinslick I think that's not suitable for blocked services component and you may need to consider for open separate feature request then it'll be new component in adguardhome for telemetry/tracking protection. Let's see what'll @ameshkov say 😊

ameshkov:
Yeah, to be honest, this a blocklists' territory. "Services" are supposed to block access to specific companies, not trackers or ads.

PR #2299, opened 2020-11-12
Fix french traduction

Lunik:
In French "Encryption" translate to "Chiffrement".
"Cryptage" doesn't exists.


ameshkov:
Could you please do this on Crowdin?
https://github.com/AdguardTeam/AdGuardHome#help-with-translations

Translations are imported from there automatically

PR #2312, opened 2020-11-14
Prometheus DNS metrics

### LINKED ISSUE #516: Metrics endpoint for Prometheus
Let's please add an `/metris` endpoint to service up metrics to Prometheus. We should be able to expose all of the internal stats already exposed in the UI plus all the standard Go runtime stats.

radhus:
This PR introduces an **unauthenticated** `/metrics` endpoint, if enabled, which presents Prometheus metrics.

Adds two metrics, besides the default ones:
* `adguard_dns_request_duration_second`
   Histogram with buckets based on CoreDNS `coredns_dns_request_duration_seconds` metric.

* `adguard_requests_total`
   Counter for every requests, with result labels.

I chose not to include a `client` label on `adguard_requests_total` because that would probably lead to [too many time series](https://prometheus.io/docs/practices/naming/#labels).

Future work could be to expose metrics for the HTTP endpoints as well.

Fixes #516.

### DIFF
--- a/internal/prometheus/prometheus.go
+++ b/internal/prometheus/prometheus.go
@@ -1,0 +1,93 @@
+package prometheus
+
+import (
+	"net"
+	"net/http"
+	"strconv"
+
+	"github.com/AdguardTeam/AdGuardHome/internal/stats"
+	"github.com/AdguardTeam/golibs/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Config - Prometheus configuration
+type Config struct {
+	Enabled   bool   `yaml:"enabled"`
+	BindHost  string `yaml:"bind_host"`
+	BindPort  int    `yaml:"bind_port"`
+	Namespace string `yaml:"namespace"`
+}
+
+type Server struct {
+	conf Config
+	mux  *http.ServeMux
+
+	dnsRequests *prometheus.CounterVec
+	dnsDuration prometheus.Histogram
+}
+
+func Create(config Config) *Server {
+	s := Server{}
+	s.conf.Enabled = config.Enabled
+	s.conf.BindHost = config.BindHost
+	s.conf.BindPort = config.BindPort
+	s.conf.Namespace = config.Namespace
+
+	if !s.conf.Enabled {
+		return &s
+	}
+
+	s.mux = http.NewServeMux()
+	s.mux.Handle("/metrics", promhttp.Handler())
+
+	s.dnsRequests = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name:      "requests_total",
+			Subsystem: "dns",
+			Namespace: s.conf.Namespace,
+			Help:      "Counter of DNS requests made per result type.",
+		},
+		[]string{
+			"result",
+		},
+	)
+	s.dnsDuration = promauto.NewHistogram(prometheus.HistogramOpts{
+		Name:      "request_duration_seconds",
+		Subsystem: "dns",
+		Namespace: s.conf.Namespace,
+		Buckets:   prometheus.ExponentialBuckets(0.00025, 2, 16), // from 0.25ms to 8 seconds
+	})
+
+	return &s
+}
+
+// Update counters
+func (s *Server) Update(e stats.Entry) {
+	if !s.conf.Enabled {
+		return
+	}
+
+	labels := prometheus.Labels{
+		"result": e.Result.String(),
+	}
+
+	s.dnsRequests.With(labels).Inc()
+	s.dnsDuration.Observe(float64(e.Time) / 1000000)
+}
+
+// Start server
+func (s *Server) Start() {
+	if !s.conf.Enabled {
+		return
+	}
+
+	port := strconv.Itoa(s.conf.BindPort)
+	addr := net.JoinHostPort(s.conf.BindHost, port)
+	go func() {
+		if err := http.ListenAndServe(addr, s.mux); err != nil {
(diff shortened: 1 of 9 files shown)


ameshkov:
Cool, thank you for the PR! We'll take a closer look during the week after v0.104.2 is released.

internal/prometheus/prometheus.go
```diff
@@ -0,0 +1,76 @@
...
+	if !s.conf.Enabled {
+		return &s
+	}
+
+	if s.conf.Namespace == "" {
+		s.conf.Namespace = "adguard"
+	}
+
+	// Use http.Handle directly to be unauthenticated.
+	http.Handle("/metrics", promhttp.Handler())
```
ameshkov, line 41: It'd be better to have it listen on a different port & create server mux here instead of registering the handler globally.

internal/stats/stats.go
```diff
@@ -76,10 +76,29 @@ const (
 	rLast
 )
 
+func (r Result) String() string {
+	switch r {
+	case RNotFiltered:
+		return "NotFiltered"
```
ameshkov, line 82: let's use `lower_case` everywhere

internal/prometheus/prometheus.go
```diff
@@ -0,0 +1,76 @@
...
+	Enabled   bool   `yaml:"enabled"`
+	Namespace string `yaml:"namespace"`
+
+	AnonymizeClientIP bool `yaml:"-"` // anonymize clients' IP addresses - set from DNS config
+}
+
+type Server struct {
+	conf Config
+
+	dnsRequests *prometheus.CounterVec
```
ameshkov, line 23: These two metrics are kinda not enough to create a good grafana dashboard:)

Any ideas what else could be added?

internal/prometheus/prometheus.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Config - Prometheus configuration
+type Config struct {
+	Enabled   bool   `yaml:"enabled"`
+	Namespace string `yaml:"namespace"`
+
+	AnonymizeClientIP bool `yaml:"-"` // anonymize clients' IP addresses - set from DNS config
```
ameshkov, line 17: I don't see how it's used further

ameshkov (changes requested):

radhus:
Thanks for reviewing, I will try to take time soon this week and update the PR :)

internal/prometheus/prometheus.go
radhus: I agree, although I aimed for having smallest possible scope in this PR and presenting the same statistics as the start page in AdGuard.

I'm not sure what other DNS metrics could be added without having too high cardinality labels (e.g. counters per client IPs or target hostnames). Perhaps the list that caused the block/filter could be added as a label?

Other than that, I guess we could add metrics for other parts of the application, e.g. last time filters were updated?

internal/prometheus/prometheus.go
radhus: Oops, a left over from when developing which should've been cleaned up :)

internal/stats/stats.go
radhus: Fixed!

internal/prometheus/prometheus.go
radhus: Fixed!

internal/prometheus/prometheus.go
ameshkov: Well, we could add some simple stuff like `build_info`, see how it's done in CoreDNS:
```
var buildInfo = promauto.NewGaugeVec(prometheus.GaugeOpts{
	Namespace: plugin.Namespace,
	Name:      "build_info",
	Help:      "A metric with a constant '1' value labeled by version, revision, and goversion from which CoreDNS was built.",
}, []string{"version", "revision", "goversion"})
```

This should also be an example of how to update prom metrics from a different part of the program.

ameshkov (changes requested):

internal/prometheus/prometheus.go
hadret: As I have personal interest in having this available, I wanted to provide a small suggestion. There's a project here on GitHub called [ebrianne/adguard-exporter](https://github.com/ebrianne/adguard-exporter). I think it can provide some additional ideas on what could be potentially exposed + there's a Grafana dashboard ready for taking (at least as long as the metrics names are compatible) 😊

ameshkov: Oh, thank you! We'll mention it in README then

k1ng440:
please merge it

ameshkov:
Assigned to a milestone just in case

radhus:
Sorry, I've had no time to spend on this PR 😢 

I can try to rebase tonight, but perhaps we can merge it as is and someone else could implement any other metrics as new PRs? WDYT @ameshkov ?

ameshkov:
@radhus no worries, thank you! I think we can continue the work ourselves at some point. However, since we're going to make some serious changes to AGH internals, it'd be better to not overcomplicate things for now and keep the PR open until we're ready to work on it.

As a temporary solution, there's a Prometheus exporter for AGH:
https://github.com/ebrianne/adguard-exporter

ReSearchITEng:
Hi @ameshkov,
unfortunately the exporter project disappeared, and the previous docker-hub images are not work (at least for latest adguard fail).
Missing: [Prometheus exporter for AdGuard Home](https://github.com/ebrianne/adguard-exporter)  by @ebrianne ..

As @radhus mentioned, adding p8s metrics would be a natural thing, especially that this project is around infrastructure.

As this was closed over 2 years back, can anyone confirm there is a plan to take exporting p8s metrics into consideration?

PR #2392, opened 2020-12-04
Add security to openapi spec

shopeonarope:
I went to create a client with the swagger codegen client (never tried it before) and after digging around in the generated code it looked like the auth settings were missing. After referencing [this](https://swagger.io/docs/specification/authentication/basic-authentication/) I got the codegen client working after adding those elements to the spec.

### DIFF
--- a/openapi/openapi.yaml
+++ b/openapi/openapi.yaml
@@ -12,6 +12,9 @@
 'servers':
 - 'url': '/control'
 
+'security':
+- 'basicAuth': []
+
 'tags':
 - 'name': 'clients'
   'description': 'Clients list operations'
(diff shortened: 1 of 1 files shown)


ainar-g:
<p>Hi,&#xA0;thank you for&#xA0;the&#xA0;change!&#x2003;We will review
it&#xA0;shortly.</p>

openapi/openapi.yaml
```diff
@@ -12,6 +12,9 @@
 'servers':
 - 'url': '/control'
 
+'security':
+  - 'basicAuth': []
```
ainar-g, line 16: There shouldn't be an indentation here, according to our [*YAML* style].  Other than that, it's good to go.

[*YAML* style]: https://github.com/AdguardTeam/AdGuardHome/blob/master/HACKING.md#yaml

ainar-g (changes requested):

openapi/openapi.yaml
shopeonarope: fixed

ainar-g (approved):
Thanks!  We'll merge shortly.

PR #2397, opened 2020-12-05
Add DNS rebinding protection

### LINKED ISSUE #102: Provide an option for blocking internal IP addresses
https://forum.adguard.com/index.php?threads/adguard-dns-discuss.18080/

> Cannot block internal IP addresses - DNS responses containing IP addresses listed in RFC1918 should be filtered out to mitigate DNS Rebinding attacks. For example, if badstuff.attacker.com points to 192.168.1.1 then that response should be filtered out. (OpenDNS allows this option and adguard should too)

juniorz:
The implementation of IP blocking is based on `dnsmasq` in regard to which CIDRs to block.

Closes #102

### DIFF
--- a/client/src/__locales/en.json
+++ b/client/src/__locales/en.json
@@ -586,5 +586,13 @@
     "port_53_faq_link": "Port 53 is often occupied by \"DNSStubListener\" or \"systemd-resolved\" services. Please read <0>this instruction</0> on how to resolve this.",
     "adg_will_drop_dns_queries": "AdGuard Home will be dropping all DNS queries from this client.",
     "client_not_in_allowed_clients": "The client is not allowed because it is not in the \"Allowed clients\" list.",
-    "experimental": "Experimental"
+    "experimental": "Experimental",
+    "rebinding_title": "DNS Rebinding Protection",
+    "rebinding_desc": "Here you can configure protection against DNS rebinding attacks",
+    "rebinding_protection_enabled": "Enable protection from DNS rebinding attacks",
+    "rebinding_protection_enabled_desc": "If enabled, AdGuard Home will block responses containing host on the local network.",
+    "rebinding_allowed_hosts_title": "Allowed domains",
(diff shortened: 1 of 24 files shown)


ameshkov:
Wow that's massive, thank you!

ArtemBaskal:
Great, but there is a need to add some logic to handle the new DNS rebinding protection request in the query log correctly

juniorz:
@ArtemBaskal could you give some pointers on what's missing specifically? Thanks for reviewing.

ameshkov:
@juniorz the problem is that this new status is not handled by the query log.

1. There's no human-readable name for it.
2. There's no way I can find it using any search criteria. I suggest reusing "filteringStatusBlockedSafebrowsing" for it  -- see `searchcriteria.go`. Let it find both queries blocked by Safe Browsing web service AND by DNS rebinding protection (and in the future, we may extend it with other security features).
3. I also suggest reusing the same color indication as for `FilteredSafeBrowsing`.

juniorz:
Hi @ameshkov , thank you for the feedback. I believe I have found my way through the stats and logging subsystems and I was able to make the requests blocked to be displayed to the end-user.

Could you please review and provide feedback on the choice of colors/categorization?

I have marked responses blocked due to DNS rebinding attempts contribute to "Blocked by Filters" stats. In the logs, they are in the "Filtered" category but I have also added a "Blocked reminding" category.

juniorz:
Also, lint failed but the Github check does not give any meaningful output so I don't really know what needs to be fixed. And running `make go-deps go-tools go-lint VERBOSE=2` does not show anything either.

ameshkov:
@juniorz got it, thank you! We'll take care of it

Dynasty-Dev:
When will this be merged?

relvinarsenio (approved):

Sil3ntVip3r:
I checked the linked issue and current product state. #102 was closed as
completed in v0.107.67 through the vetted HaGeZi DNS Rebind Protection
filter, and the issue discussion confirms normal user-rule allowlisting for
private domains.

This older built-in implementation now conflicts across the DNS and UI
architecture and is no longer the production path. It can be closed as
superseded by the shipped solution.

PR #2594, opened 2021-01-23
Update fr.json

jerometerrier:
replace seveur by serveur in the locale fr traduction


Aikatsui:
@jerometerrier you need to fix this through https://crowdin.com/project/adguard-applications/en#/adguard-home

ameshkov:
@jerometerrier thanks for the contribution! However, @Aikatsui is right, we accept translations changes via Crowdin: https://github.com/AdguardTeam/AdGuardHome#help-with-translations

All translations are imported from there automatically.

PR #3077, opened 2021-05-05
Disabled autocapitalize on username field in mobile browsers

jankais3r:
The username is case sensitive. Since most usernames are not capitalized, it is annoying that mobile browsers automatically capitalize the first letter in the login username input field. This pull request adds the autocapitalize="none" property which disables such behavior.


ainar-g:
Thanks for the contribution!  We will review this shortly.

PR #3245, opened 2021-06-08
Add links to all the dashboard things

markhicken:
I found myself living on the Query Log instead of on the Dashboard. This PR makes the Dashboard more useful by adding links on each stat card and each table row item to the filtered Query Log for that item. 

Let me know what you think. I'm happy to make changes if needed.

Screen Shot 2021-06-08 at 3 06 57 PM

### DIFF
--- a/client/src/components/Dashboard/Counters.js
+++ b/client/src/components/Dashboard/Counters.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import propTypes from 'prop-types';
+import { Link } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


ainar-g:
Thanks for your contribution!  We have merged an updated version that fits a bit better with the current design in 232cd38.

Zarawee:
Hi, in which version, this feature is available?

ainar-g:
@Zarawee, currently, on the beta channel of the v0.107 version.  The final release has been unfortunately postponed a few times, but we currently plan to release it within a couple of weeks.

Issue #3366, opened 2021-07-20
Adguard Home DHCP Issue


Saku241:
### Issue Details

* **Version of AdGuard Home server:**
  * v0.106.3
* **How did you install AdGuard Home:**
  * `curl -sSL https://raw.githubusercontent.com/AdguardTeam/AdGuardHome/master/scripts/install.sh | sh`
* **How did you setup DNS configuration:**
  * Load-balancing
* **If it's a router or IoT, please write device model:**
  * Raspberry Pi 3 Model B
* **CPU architecture:**
  * ARM64
* **Operating system and version:**
  * Raspbian 10

### Actual Behavior

I got a problem with the DHCP of Adguard Home. I disabled my router's DHCP and Adguard DHCP seems to work fine with WIFI devices because they are granted an IP BUT for Wired devices (RJ45) they never get an IP. 
I actually have to add every MAC addresses of wired devices to manually assign IP to each. It kinda works but sometimes when I turn on my computer, it takes a while to get an IP (approx. 30s-2m).  

My IP Range is 192.168.1.2 to 192.168.1.126, subnet 255.255.255.0 and lease 86400sec

### Additional Information

I added the MAC addresses of my devices in client section but I cant determine the device if it's an other IP, I thought it would be able to identify with only the MAC Address

Saku241:
> Hello. Are you sure that your router processes both wired and wireless connections using a single DHCP server? 

Not sure but my router got basic dhcp options 

I’ll try the verbose and update ticket

ainar-g:
Please make sure that AdGuardHome is actually up.  For example that the web interface is shown and that you can see DNS queries on the Query Log page.

Saku241:
Adguard is on but there's no query as devices didnt get a correct local ip. If I put back verbose false and log_file "", Adguard start working as normal.. Anything wrong with the log_file directory ?

ainar-g:
Are you sure that `/home/pi/logs` is a **file** and that AGH can access it?

Saku241:
> Are you sure that `/home/pi/logs` is a **file** and that AGH can access it?

Oof I thought it was only a directory not a file...

So here's verbosse logs :

- https://pastebin.com/raw/tBFgcYrt Here, I got my local ip (192.168.1.22) with static lease in approx 30-45sec
[...]

PR #3458, opened 2021-08-16
feat: support unicast for dhcp replies

### LINKED ISSUE #3053: DHCP Issue again.
Hi, after few test, I have some trouble again with the DHCP. Some of my devices can't obtain an IP and I need to set a static lease if I want it to work. 

I also encounter some of disconnection with few other. In the end, my log file are flooded by message about normalisation error. 

I run now the 1.0.6.1 b1

Best regards 

[AdGuard.log](https://github.com/AdguardTeam/AdGuardHome/files/6409005/AdGuard.log)

_Originally posted by @Black8all in https://github.com/AdguardTeam/AdGuardHome/issues/2978#issuecomment-830565777_

### LINKED ISSUE #3443: DHCP server should take request's broadcast flag into account
DHCP Discover requests may have a broadcast flag.  As Wikipedia [states](https://en.wikipedia.org/wiki/Dynamic_Host_Configuration_Protocol#Operation):

> Additionally, there is a BROADCAST flag (1 bit in 2 byte flags field, where all other bits are reserved and so are set to 0) the client can use to indicate in which way (broadcast or unicast) it can receive the DHCPOFFER [...]

Responding to the broadcast address may sometimes be necessary for some clients so the flag should be handled properly.

Additionally, we should respect the behaviour described in the RFC regarding the sending of addresses depending on what information we have about the client (e.g. when we only have a hardware address).

We should consider writing these messages to all interfaces where we can.  A separate issue will be filled regarding the mental model of it all.

anbraten:
closes #3443
closes #3053
and maybe addresses some of the latest DHCP bug issuses and maybe also
addresses #3366 

# Background
By using Wireshark I noticed most clients requests Unicast responses instead of Broadcasts from DHCP servers and these well-known / battle-tested DHCP servers respect these requirements. Some device fallback and also allow Broadcast responses, but some are not (like ESP32, ESP8266, my HP-printer, Chromecast, Smart-TV etc.).

# Changes
This PR adds support for unicast responses inspired / using code from [coredhcp](https://github.com/coredhcp/coredhcp/blob/master/server/sendEthernet.go). This code allows to send a response to a mac address by using [gopacket](https://github.com/google/gopacket) to create and send a raw network packet.

# TODO
- [x] detect if request wanted old broadcast response and properly answer if so
- [x] test if dhcp requests with broadcast flag are handled correctly

# Tests
- I tested it with some device I had lying around and it worked for all of them so far
  - Android
  - Linux PC
  - ESP32 / ESP8266
  - Chromecast
  - Alexa Echo Dot

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/go-ping/ping v0.0.0-20210506233800-ff8be3320020
 	github.com/google/go-cmp v0.5.5
+	github.com/google/gopacket v1.1.19
 	github.com/google/renameio v1.0.1
 	github.com/insomniacslk/dhcp v0.0.0-20210310193751-cfd4d47082c2
 	github.com/kardianos/service v1.2.0
--- a/go.sum
+++ b/go.sum
@@ -94,6 +94,8 @@ github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-github v17.0.0+incompatible/go.mod h1:zLgOLi98H3fifZn+44m+umXrS52loVEgC2AApnigrVQ=
 github.com/google/go-querystring v1.0.0/go.mod h1:odCYkC5MyYFN7vkCjXpyrEuKhc/BUO6wN/zVPAxq5ck=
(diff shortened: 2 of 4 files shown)


ainar-g:
Hello, and thank you for this contribution.  Please make sure that your code conforms to our [Go code guidelines][go].  Unfortunately, we're currently very busy with the ongoing release stuff, so it might take us a while to perform a more thorough review.

[go]: https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Go.md

EugeneOne1:
@anbraten, hello again.  In the meantime we've created our own implementation of unicasting to hardware addresses in da86620288256df7b0c51b0b413ded2a09aae940.  It's inspired by your work but also enhances the approach you suggest.  However, we're honored by your contribution and will mention you in the notes to the next release.

The implementation is also available in latest edge builds.  Could you please check if it works properly?

ainar-g:
We'll close this PR for now.  Once again, thank you for the contribution!

PR #3717, opened 2021-10-08
Fix unsupported regex for QQ blocked rules

Hill-98:
The source rule uses regex that Golang doesn't support and is now changed to a normal rule.


Hill-98:
Can it be merged?

Hill-98:
Can it be merged?

ainar-g:
Apologies for the late response!  We'll review and commit this shortly.

Hill-98:
You are welcome. and thank you for such an awesome project.

PR #3745, opened 2021-10-18
Cleaned up the list

bruvv:
Since SPAM404 is not maintained anymore and is a huge full spam list now of it's own (https://github.com/Spam404/lists/pull/18) it is save to remove it I would say.
Also removed BarbBlock since it is also unmaintained (Last Update: 2019-05-26)
Added Netherlands blocking list

I also would like to suggest [OISD](https://oisd.nl/) as a list since it is the only list you basically need in my opinion what do you think ?

### DIFF
--- a/client/src/helpers/filters/filters.json
+++ b/client/src/helpers/filters/filters.json
@@ -60,12 +60,6 @@
             "homepage": "https://github.com/crazy-max/WindowsSpyBlocker",
             "source": "https://raw.githubusercontent.com/crazy-max/WindowsSpyBlocker/master/data/hosts/spy.txt"
         },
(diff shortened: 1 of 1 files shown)


bruvv:
thanks @ameshkov what did you think about the OISD list that I mentioned?

mokide:
I would suggest this list as a replacement for Spam404 https://gitlab.com/curben/phishing-filter it's from the same author of Online Malicious URL. There's also this one which could be selected as an extra addition to the security https://gitlab.com/curben/pup-filter

Both of these have been recently added to uBlock Origin since Spam404 is dead since months by now.

PR #3792, opened 2021-11-01
Updated PT locale - "Repor" replaced with "Actualizar"

eSkiSo:
Updated PT locale
TLDR: "Repor" means "Reset" and "Actualizar" means "Refresh"
Under **refresh_statistics** "Repor" was used in translation which confuses users that may think that button is used to reset the statistics instead of updating them.


ainar-g:
Thank you for your contribution, but we keep all of our translations on Crowdin: https://crowdin.com/project/adguard-applications/en#/adguard-home.  We update the locale texts before every beta and final release, and we'll be happy to see your contributions there!

@el-termikael, please take these changes into account.

PR #3933, opened 2021-12-12
Prevent spurious diffs in config file by sorting Client objects before writing

hnefatl:
This PR aims to prevent occasional reordering diffs like the following from appearing in the `AdGuardHome.yaml` config file:
```
 clients:
-- name: device 1
-  ids:
-  - x.x.x.x
-  ...
 - name: device 2
   ids:
   - y.y.y.y
   ...
+- name: device 1
+  ids:
+  - x.x.x.x
+  ...
```

This is problematic when including the config file in version control, since occasional no-op changes pop up and require reverting/committing.

I think the root cause of these reorders is that Go doesn't guarantee a consistent iteration order over maps (see the "Iteration order" section of https://go.dev/blog/maps), so e.g. different executions of an adguard docker container/daemon will produce different orders. By sorting the list of objects we create from the map before outputting them, we guarantee clients will be ordered by name.

I've only noticed this being an issue for the `clients` section of the config file: I haven't done a deep dive into the other config file fields, but from a quick skim of my config it looks they're probably not stored in maps so don't see this behaviour.

### DIFF
--- a/internal/home/clients.go
+++ b/internal/home/clients.go
@@ -222,6 +222,9 @@ func (clients *clientsContainer) addFromConfig(objects []clientObject) {
 // WriteDiskConfig - write configuration
 func (clients *clientsContainer) WriteDiskConfig(objects *[]clientObject) {
 	clients.lock.Lock()
+	defer clients.lock.Unlock()
+
+	clientObjects := []clientObject{}
 	for _, cli := range clients.list {
 		cy := clientObject{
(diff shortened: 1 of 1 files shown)


ainar-g:
Thanks for the contribution!  We'll try to review it shortly.

ainar-g:
Looks great!  We've made some additional adjustments and merged it.  Thanks again for the contribution!

PR #4157, opened 2022-01-17
Add DOH bootstraps

tomitheninja:
If the ISP blocks (or redirects to an overloaded server) communication on port 53, adguardhome will not be able to resolve the ip address of the dns over https servers.
For this reason cloudflare (and many others) provide dns over https service on their ip addresses, without a domain name.

I tested this change by blocking everything with ufw on my local machine, and then I added https://1.1.1.1/dns-query to the list of bootstrap servers and it seems to work.

### DIFF
--- a/internal/dnsforward/dnsforward.go
+++ b/internal/dnsforward/dnsforward.go
@@ -41,7 +41,7 @@ const (
 var defaultDNS = []string{
 	"https://dns10.quad9.net/dns-query",
 }
-var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "2620:fe::10", "2620:fe::fe:10", "https://1.1.1.1/dns-query"}
+var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "https://9.9.9.10/dns-query", "2620:fe::10", "2620:fe::fe:10"}
 
 // Often requested by all kinds of DNS probes
 var defaultBlockedHosts = []string{"version.bind", "id.server", "hostname.bind"}
--- a/internal/dnsforward/dnsforward.go
+++ b/internal/dnsforward/dnsforward.go
@@ -41,7 +41,7 @@ const (
 var defaultDNS = []string{
 	"https://dns10.quad9.net/dns-query",
 }
-var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "2620:fe::10", "2620:fe::fe:10"}
+var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "2620:fe::10", "2620:fe::fe:10", "https://1.1.1.1/dns-query"}
 
 // Often requested by all kinds of DNS probes
 var defaultBlockedHosts = []string{"version.bind", "id.server", "hostname.bind"}


internal/dnsforward/dnsforward.go
```diff
@@ -41,7 +41,7 @@ const (
 var defaultDNS = []string{
 	"https://dns10.quad9.net/dns-query",
 }
-var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "2620:fe::10", "2620:fe::fe:10"}
+var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "https://9.9.9.10/dns-query", "2620:fe::10", "2620:fe::fe:10"}
```
emlimap, line 44:
```suggestion
var defaultBootstrap = []string{"9.9.9.10", "149.112.112.10", "https://9.9.9.10/dns-query", "2620:fe::10", "2620:fe::fe:10", "https://[2620:fe::fe]/dns-query"}
```

IPv6 address is accessible over DoH as well

agneevX: This is getting a little overcrowded since it's just used to resolve hostnames for the upstreams. 

Ideally, there should be at most three servers here, secure or IPv6.

emlimap: I agree. Personally I would say we should just have DoH & DoT addresses for bootstrap as well and avoid plain text DNS. This helps the original issue this PR is trying to fix where ISP can block or redirect without much effort.

The only issue this might cause is if people run Adguard home on devices such as Raspberry Pi or routers with RTC that depend on NTP to set correct time on boot, might end up in a catch 22 scenario, where you need DNS up and running to get time but DNS doesn't work because the time isn't correct to negotiate TLS or HTTPS.

emlimap: > My ISP does this, and while I use DoH servers via IP only, I don't think it's much of an issue?

The issue is as a user I can't choose which DNS server to use. This makes it impossible to run recursive unbound server bypassing public DNS servers, almost same as running your own public DNS server at home.
 
I don't see the benefit of them doing this other than another revenue stream for them through serving advert pages for non-existing domain lookups or selling their customers DNS usage to data brokers. 

The only legitimate use case I can think for hijacking is if subscriber utilizes parental control feature offered by ISP in which case it should be done only for subscribers who subscribe to it.

> If they poison their cache for the hostnames, that's possible but extremely rare given it can bring forward lawsuits.

It depends on the law's of the country and how accessible it is. Also, attacks on less popular domains or targetting for brief period might make it harder to be discovered.

> I actually do this. I'm sure there's race conditions involved, but so far there's been no issue and the time has been correct by the time I do a check after startup.

> If I used DNSSEC validation, it could be an issue.

Good to know

tomitheninja:
ping

ainar-g:
@ameshkov, what do you think about adding a DoH default bootstrap?

ameshkov:
I think we should either have ONLY encrypted bootstraps there, or have no encrypted bootstraps at all.

nep-0:
> If the ISP blocks (or redirects to an overloaded server) communication on port 53, adguardhome will not be able to resolve the ip address of the dns over https servers. For this reason cloudflare (and many others) provide dns over https service on their ip addresses, without a domain name.
> 
> I tested this change by blocking everything with ufw on my local machine, and then I added https://1.1.1.1/dns-query to the list of bootstrap servers and it seems to work.

This is how I solved this problem: I set the bootstrap to `127.0.0.1` and add a DNS rewrite rule for each upstream.

PR #4235, opened 2022-01-31
Remove `CMD` host flag from Dockerfile

### LINKED ISSUE #4231: Cannot set `bind_host` in AdGuardHome.yaml (docker version)
### Issue Details

* **Version of AdGuard Home server:**
  * v0.107.3 (docker)
* **How did you install AdGuard Home:**
  * docker
* **How did you setup DNS configuration:**
  * Here is my config file: https://github.com/pexcn/docker-images/blob/master/ext/adguardhome/adguardhome-data/conf/AdGuardHome.yaml
  * Use docker-compose, here is my compose config: https://github.com/pexcn/docker-images/blob/master/ext/adguardhome/docker-compose.yml
* **CPU architecture:**
  * AMD64
* **Operating system and version:**
  * OpenWrt

Hi guys:
Now I am using v0.107.3 version, I want to change `bind_host` to `192.168.1.1`, but after restarting adguardhome, the value in the configuration file will automatically change to `0.0.0.0`.

Maybe it has something to do with this line of the Dockerfile https://github.com/AdguardTeam/AdGuardHome/blob/v0.107.3/scripts/make/Dockerfile#L59, can it be fixed in the future? :smile:

agneevX:
This flag served no real purpose other than listening to all interfaces, despite selecting a specific interface during onboarding. 

Also use long flags instead of shorthands

Closes #4231


ainar-g:
Thanks for the contribution!  We'll probably merge it this week, but the final decision is yet to be made.  In any way, this is going to be in v0.108.0, since this will definitely break some people's installations if we're not careful.

ainar-g (changes requested):
@‌ameshkov thinks that this change as-is could break too many installations.  To prevent a breaking change like that, could you add the `-h` back, but put its value into an `ENV` instruction?  I.e. `ENV AGH_BIND_HOST=0.0.0.0`.  We'll then update the documentation to mention that people who want to bind to a specific address inside their containers should use `--env AGH_BIND_HOST=192.168.1.1`.

agneevX:
OK, but exec form does not allow for shell expansion.

Shell form can be used but does have [caveats](https://docs.docker.com/engine/reference/builder/#entrypoint):

> The shell form prevents any CMD or run command line arguments from being used, but has the disadvantage that your ENTRYPOINT will be started as a subcommand of /bin/sh -c, which does not pass signals. This means that the executable will not be the container’s PID 1 - and will not receive Unix signals - so your executable will not receive a SIGTERM from docker stop <container>.

ainar-g:
*Sigh.*  Docker being Docker, heh.  Something like this should probably work, albeit quite a hack:

```dockerfile
ENV AGH_BIND_HOST='0.0.0.0'

ENTRYPOINT [ \
	"/bin/sh", \
	"-c", \
	"exec /opt/adguardhome/AdGuardHome --host \"$AGH_BIND_HOST\" \"$@\"", \
	"--" \
]

CMD [ \
	"--config", "/opt/adguardhome/conf/AdGuardHome.yaml", \
	"--no-check-update", \
	"--work-dir", "/opt/adguardhome/work" \
]
```

agneevX:
Ehh that could work but do you want to have your Dockerfile looking like that?

I'd say supporting environment variables is something that the binary itself should understand, without the need for flags in the middle.

This does not feel like the correct way to do this. Since this is env vars, it might not be fair to restrict it to a single _item_.

ainar-g:
I think, we've had several issues about using the environment more actively.  We'll need to think about it.  But until then, this PR will remain open, if you don't mind.

Chinaza12 (approved):

jonathanjsimon:
Necrobumping

With the `-h 0.0.0.0` in the Dockerfile, AdGuard becomes unusable in my use case as I have two interfaces and port 80 is in use on one of them. Specifying the other interface's IP is the only way to use AdGuard *and* have individual client IPs listed in the interface.

jonathanjsimon:
Also, since this setting will have previously configured a user's `AdGuardHome.yaml` file with `0.0.0.0`, how does removing the flag break an installation? It's already configured to listen on all interfaces. For further protection, couldn't the application itself be altered to presume `0.0.0.0` if no other value is present from switches or config files?

jonathanjsimon:
For anyone wanting to do this now, https://hub.docker.com/r/jsimon0/adguardhome

Mizzick:
Hello,

With the upcoming command-line flags changes, we are going to deprecate the `-h` flag. So I think now, we are ready to merge this pull request.

Mizzick (approved):

Mizzick:
Hello again, 

We have implemented the requested changes in https://github.com/AdguardTeam/AdGuardHome/commit/e54fc9b1e9ad9f344edad86a2991308e6c5c29d0.

Thank you for you help!

PR #4376, opened 2022-03-08
Use directory with upstream_dns_file

hellodword:
How do you think about this?

```
$ tree rules
rules
├── 01-default.txt
└── 02-use-doh.txt
└── 03-use-system.txt
└── 04-ads.txt
```

### DIFF
--- a/internal/dnsforward/config.go
+++ b/internal/dnsforward/config.go
@@ -4,9 +4,11 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"io/fs"
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


ainar-g:
Hello again.  Feature requests should generally be discussed in Issues first.

@ameshkov, do we want this feature?

ainar-g:
@ameshkov?

ameshkov:
@hellodword could you please explain the need for this? IMO, it complicates things a little, but I don't see any direct benefit.

hellodword:
> @hellodword could you please explain the need for this? IMO, it complicates things a little, but I don't see any direct benefit.

Sorry this is not a well-thought-out idea when I think it again now. I think it should be closed.

PR #4387, opened 2022-03-12
correct openapi schema

bakito:
This PR fixes some issues in the openapi scheme.

- RemoveUrlRequest is missing property 'whitelist'
- FilterStatus is missing property 'whitelist_filters'
- FilterSetUrl property 'data' is set to type Filter instead of inline object
- 'last_updated' of Filter is not required. At least it comes back as an empty string from the API

It also sets the filter 'last_updated' property to omitempty, to make sure the Filter json and scheme are valid.

### DIFF
--- a/internal/home/controlfiltering.go
+++ b/internal/home/controlfiltering.go
@@ -316,7 +316,7 @@ type filterJSON struct {
 	URL         string `json:"url"`
 	Name        string `json:"name"`
 	RulesCount  uint32 `json:"rules_count"`
-	LastUpdated string `json:"last_updated"`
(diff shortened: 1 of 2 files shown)


bakito:
Any update here?

ainar-g:
Apologies for this slipping through the cracks.  We'll try to review it for the next patch release.

ainar-g:
Once again, apologies for the long wait.  We've made some additional changes and merged the PR.  Thanks for the contribution!

bakito:
Thank you very much

PR #4417, opened 2022-03-22
update go.mod to go1.18

Potterli20:


ainar-g:
@Potterli20, thanks for the suggestion, but please discuss such changes in an issue or discussion before committing them.  I will have to decline this pull requests and the pull requests in other repositories for the following reasons:

1.  In the v0.108 cycle we have only deprecated Go 1.17, not removed it.  We'll be able to remove it during the v0.109 cycle.  See the changelog.

2.  Updates to our build and test infrastructure need to be made to support Go 1.18.

3.  Besides our own, there are people building their own AdGuard Home and also supporting OS packages.  They will need to update as well.

We will make proper updates on our own schedule.

Potterli20:
> @Potterli20, thanks for the suggestion, but please discuss such changes in an issue or discussion before committing them. I will have to decline this pull requests and the pull requests in other repositories for the following reasons:
> 
>     1. In the v0.108 cycle we have only deprecated Go 1.17, not removed it.  We'll be able to remove it during the v0.109 cycle.  See the changelog.
> 
>     2. Updates to our build and test infrastructure need to be made to support Go 1.18.
> 
>     3. Besides our own, there are people building their own AdGuard Home and also supporting OS packages.  They will need to update as well.
> 
> 
> We will make proper updates on our own schedule.

okay

PR #4431, opened 2022-03-26
Case-insensitive username check

### LINKED ISSUE #2581: Username should not be case sensitive
During login, the username seems to be case sensitive.

in my Case: admin will work, Admin not.

Bug or Feature?

mdawsonuk:
Resolves #2581 by ensuring that auth checks for the username are performed by doing a lower-case comparison.

A test case was also added in `auth_test.go` to ensure that the case-insensitive check is correctly performed.


mdawsonuk:
Not sure why this is being marked as wontfix

Issue #613, opened 2019-03-02
Feature Request: Optional Dark Theme


Ultrabenosaurus:
Please can you add an optional dark theme of some sort? There are plenty of examples online to take inspiration from like community-made Atom themes hosted on Github. I have a browser extension that allows me to turn any website "dark" but it basically just inverts CSS colours, so the colour-coded bits in AdGuard Home and the logo in the top-left don't really work. 

<details><summary>Screenshot:</summary>

![image](https://user-images.githubusercontent.com/1901532/53680286-17677b80-3cd1-11e9-9b00-26f3d5bfcafd.png)

</details>

### Your environment

|   Description  |     Value    |
| -------------- | ------------ |
| Version of AdGuard Home server:| 0.93
| How did you setup DNS configuration:| Primary DNS for my Router
| If it's a router or IoT, please write device model:| Raspberry Pi 3 Model B 
| Operating system and version:| latest Raspbian (I updated last night)

### EDIT

For those interested I have published my latest dark theme CSS on userstyles.org for Stylus users: 

https://userstyles.org/styles/177418/adguard-home-dark 

I recommend using the "Customize Settings" button to set the IP address or local domain of your AdGuard Home interface before installing the theme, so that it works straight away. I've also added a setting so you can specify a main font; personally, I quite like [Lato](https://fonts.google.com/specimen/Lato) at the moment. Thanks to @carlbennett for the initial CSS and inspiration to tinker with this.

ameshkov:
Well, the UI is not our top priority at the moment, but we'll take this feature request into account once we start working on the final design.

Ultrabenosaurus:
That's fine, I thought that'd be the case for now :)

frenck:
It would look amazing though...

![image](https://user-images.githubusercontent.com/195327/57011125-5c8c2900-6c00-11e9-83f7-9d31048383c9.png)

carlbennett:
I'm making a userstyle theme for those interested. This is just a work in progress with what I have right now. But for me, this is so much nicer already.

Get the Stylus extension: https://add0n.com/stylus.html

Add this code:
```css
@-moz-document domain("192.168.1.22") {
body { background-color: #222; color: #eee; }
a { color: #aaffaa; }

.header { background-color: #336633; color: #eee; }
.mobile-menu { background-color: inherit; }

.footer { background-color: #113311; }
.footer a:not(.btn) { color: #aaffaa; }

.card { background-color: #333; }
[...]

Issue #2554, opened 2021-01-13
AdGuard redesign tracking issue


ameshkov:
It may be necessary to consult this issue: https://github.com/AdguardTeam/AdGuardHome/issues/2075

This issue just contains a checklist of pages that we implement (see "client2").

* [X] Initial setup
* [ ] Login
* [ ] Dashboard
* [ ] Setup guide - part 1 (before review)
* [ ] Setup guide - part 2 (after review)
* [ ] DNS blocklists
* [ ] DNS whitelists
* [ ] DNS rewrites
* [ ] Blocked services
* [ ] Custom filtering rules
* [ ] Settings -> General settings
* [ ] Settings -> DNS settings
* [ ] Settings -> Encryption
* [ ] Settings -> Client settings
* [ ] Settings -> DHCP
* [ ] Query log
* [ ] Global notifications
    * [ ] New version is available
    * [ ] Certificate is about to expire

Blakhard:
For now, we can not make date picker on Dashboard due to API.

ameshkov:
@Blakhard yep, np, this feature will be moved to a future version.

Blakhard:
1) Now "Blocked ads" and "Blocked trackers" are combined to "Blocked by filters". 
2) No chart data for "Average processing time"
3) "Internal server statistic" block can not be done, appear in next API releases.
4) "Server uptime" in header, unclear is it server uptime, or machine uptime.

ameshkov:
@Blakhard we'll have to implement these in this version so some changes to the backend.

WildByDesign:
@ameshkov Does this redesign implementation allow the potential for adding Dark Mode to UI elements?

Since you are at the beginning of the implementation and design stage for this, I figured I should bring this up. Dark Mode continues to be one of the most up-voted feature requests.

Thank you for all of the great work that your team does. Despite the whole Covid-19 slowing down much of the world, your team kept on rocking it without slowing down. Cheers!

ameshkov:
Yep, I guess once the redesign is completed, we'll be able to bring the dark mode soon after that

Aikatsui:
@ameshkov Is it possible to try https://github.com/AdguardTeam/AdGuardHome/issues/481#issuecomment-546758541 ?
(I know, already has many issues to address but very difficult to work with query log to view time series data)

ameshkov:
Well, I'd better do that later, we already planned too much on this redesign.

ainar-g:
[...]

PR #4527, opened 2022-04-26
added quick dark theme

Albirew:
First PR for #613
Automatic selection (for now) using prefers-color-scheme: dark.

### DIFF
--- a/client/src/components/Header/Header.css
+++ b/client/src/components/Header/Header.css
@@ -205,3 +205,9 @@
     padding: 0.35em 0.5em;
     line-height: 10px;
 }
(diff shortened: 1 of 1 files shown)


ainar-g:
Thanks for the contribution, but as you can see that issue is already marked to be resolved in v0.108.  We're actually planning a full redesign in that version, see #2554.

In the future, please discuss if a PR is necessary before sending one to prevent duplication of effort.

Albirew:
Oh, sorry about that.
BTW, this snippet of CSS is pretty universal so it can act as a pre-theme before making a true dark theme.

Issue #4520, opened 2022-04-23
API to get all service options available for setting blocked_services


RoboMagus:
Have a question or an idea? Please search it [on our forum](https://github.com/AdguardTeam/AdGuardHome/discussions) to make sure it was not yet asked. If you cannot find what you had in mind, please [submit it here](https://github.com/AdguardTeam/AdGuardHome/discussions/new).

### Prerequisites

Please answer the following questions for yourself before submitting an issue. **YOU MAY DELETE THE PREREQUISITES SECTION.**

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Problem Description

I'm trying to create an automatic schedule that will block / unblock specific services through the AdGuardHome API using `blocked_services/set`. I've been working on getting this API access in a python module so I can then call this through HomeAssistant on demand and on schedule. This works quite well, given that I've first taken a look at the blocked_services page of my local instance.
However when the set of available services would change (#3092, #3814), there is no automated way of telling.

### Proposed Solution

I would propose adding another function to the blocked_services API that allows anyone using the API to get an extensive list of services available on this interface.
E.g. `GET` `/blocked_services/services`. This could then either just return a list of service names compatible with the `/blocked_services/set` function, or a structure that contains their ID, name, (and icon).

### Alternatives Considered

Hard coding the currently available list and updating this every time the interface is updated. However that's not ideal

### Additional Information

Would like to eventually further integrate AdGuardHome into HomeAssistant, by extending the [python wrapper](https://github.com/frenck/python-adguardhome). Then having all allowed data that can be passed over the API available programmaticaly becomes even more important.

ainar-g:
Merged in #4535.

NebzHB:
Dears,
About https://github.com/AdguardTeam/AdGuardHome/pull/4535 , Is it possible to have the full name in that service list too ? example : 'ok' is displayed 'OK.ru' in the web interface ('disneyplus' : 'Disney+', ...)

PR #4535, opened 2022-04-28
Blocked services API: List available services

RoboMagus:
Related to FR #4520 

This PR adds `GET` `/blocked_services/services` that will return a full list of the service names that can be blocked using the blocked_services API.

It seemed quite easy to implement, though I must declare that I am __NOT__ proficient in Go...


ainar-g:
Hello and thank you for that contribution!  We've actually planned to postpone that until a proper API refactoring that we planned to do later in the v0.108 cycle, but since you've already provided an implementation, we'll try to review your contribution next week.  Be aware, though, that when the refactoring comes, it is likely that this API will be rewritten.

RoboMagus:
Thanks for taking a look into this. It's good to hear that these api changes are already on the roadmap. In that case it might be best to first iron out what the api should look like.

Is there an estimated time frame for when v0.108 is to be released? Cause then I'll postpone work on my integration that interacts with the api until it arrives.

frenck:
This feature is very much wished for, for the Home Assistant project as well. It would allow us to create switches in Home Assistant to automate on. Right now, we can't, as we can't get a list of services.

ainar-g:
@RoboMagus, @frenck, we have decided to merge this into v0.107.12.  Thank you for the contribution!

NebzHB:
Dears,
Is it possible to have the full name in that service list too ? example : 'ok' is displayed 'OK.ru' in the web interface ('disneyplus' : 'Disney+', ...)

PR #4586, opened 2022-05-23
Added duplicate rewrite entry check (partially fixes #4585)

### LINKED ISSUE #4585: Add some sanity checking and functionality for rewriteAdd and rewriteRemove
Have a question or an idea? Please search it [on our forum](https://github.com/AdguardTeam/AdGuardHome/discussions) to make sure it was not yet asked. If you cannot find what you had in mind, please [submit it here](https://github.com/AdguardTeam/AdGuardHome/discussions/new).

### Prerequisites

Please answer the following questions for yourself before submitting an issue. **YOU MAY DELETE THE PREREQUISITES SECTION.**

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

### Problem Description

When adding a new rewrite through /control/rewriteAdd it always adds it no matter what. Even if the same domain:answer rule already exists.
Also when removing a rule, /control/rewriteDelete always expects a full domain:answer struct.

### Proposed Solution

It would be much better to do atleast a bit of sanitychecking. Like `if domain:answer exists return "error: already exists"` instead of adding a duplicate rule.
Also a /control/rewriteDelete with only a domain entry should remove all rules that share the domain; no matter the attached "answers". Same goes for /control/rewriteDelete with only a answer. This should delete all rules where the answer matches, no matter the attached domains.
Bonus points if the response of those contains the added/removed entries for clientside sanity checking without the need for additional requests.

[ ] Add duplicate checking for /control/rewriteAdd (return duplicate error and dont save rule)
[ ] Return added entries for /control/rewriteAdd 

[ ] Allow requests to /control/rewriteDelete to bulk delete entries for only the specified domain/answer
[ ] Return deleted entries for /control/rewriteDelete

Bluscream:
Loop over existing entries and return an error if the entry already exists

### DIFF
--- a/internal/filtering/rewrites.go
+++ b/internal/filtering/rewrites.go
@@ -277,6 +277,15 @@ func (d *DNSFilter) handleRewriteAdd(w http.ResponseWriter, r *http.Request) {
 	}
 
 	d.confLock.Lock()
+
+	for _, ent := range d.Config.Rewrites {
+		if ent.Domain == rw.Domain && ent.Answer == rw.Answer {
+			aghhttp.Error(r, w, http.StatusBadRequest, "duplicate rewrite")
+			d.confLock.Unlock()
+			return
+		}
+	}
+
 	d.Config.Rewrites = append(d.Config.Rewrites, rw)
 	d.confLock.Unlock()
 	log.Debug("rewrite: added element: %s -> %s [%d]", rw.Domain, rw.Answer, len(d.Config.Rewrites))


ainar-g:
Thank you, but I'd like to put this on hold, since the HTTP API and its code will be fully rewritten during the v0.108 cycle.  Also, correct me if I'm wrong, but this won't check for duplicates in the data that is currently being sent?

Bluscream:
This only accounts for sending one single rewrite because the current implementation doesn't work for multiple rewrites.

ainar-g:
Ah, you're right.  I've been thinking ahead of myself :sweat_smile:.

Sil3ntVip3r:
I retested the underlying case on current `master`: adding the same rewrite
twice is still accepted, and `rewrite.DefaultStorage.Add` still has an
explicit duplicate-handling TODO. A review-only current-master HTTP
regression test expecting the second add to fail reproduced the defect.
After removing that temporary test, the existing rewrite HTTP suite passes
under `-race` and the checkout is clean.

This patch itself now conflicts and predates the current rewrite storage/API,
so it should not be revived as-is. The actionable current fix belongs in
`internal/filtering/rewrite/storage.go` plus HTTP/storage regression tests
under #4585. This PR can be closed as an obsolete implementation without
treating the underlying issue as fixed.

Sil3ntVip3r:
I prepared and verified the smallest current-master salvage for the still-open
duplicate-rewrite part of #4585.

The baseline was current `master`
`369d47616342c8be82262bf798f2fed2f64b697f`.  Test-first regressions confirmed
both current behaviors:

- `POST /control/rewrite/add` accepted and persisted an existing domain/answer
  pair and fired the configuration-save callback.
- `rewrite.DefaultStorage.Add` accepted the same item twice instead of
  returning `errors.ErrDuplicated`.

The focused fix changes four files:

- `internal/filtering/rewritehttp.go` performs the duplicate check while
  holding `confMu`, before appending.  A duplicate returns HTTP 400 with
  `duplicate rewrite` and does not invoke the configuration modifier.
- `internal/filtering/rewrite/storage.go` performs the equivalent check under
  its storage mutex and returns `errors.ErrDuplicated`.
- The corresponding HTTP and storage tests verify the error, unchanged list
  length, and absence of the configuration-save callback.

Validation completed successfully:

```text
go test -race -count=10 ./internal/filtering/rewrite \
  -run '^TestDefaultStorage_CRUD$'

go test -race -count=10 ./internal/filtering \
  -run '^TestDNSFilter_HandleRewriteHTTP_errors$'

go test -race -count=3 \
  ./internal/filtering ./internal/filtering/rewrite

go vet ./internal/filtering ./internal/filtering/rewrite
make go-check
git diff --check
```

All commands passed.  The complete `make go-check` gate included the
repository Go linter, vulnerability scan, race-enabled tests, and coverage
run.

The patch is 28 insertions and 2 deletions across those four files.  No
frontend files, API schema, dependencies, or unrelated rewrite behavior are
changed.

Exact PR head `0acb75e087c62e84f8ec7cb665822bd369e7fdff` still predates the current
rewrite architecture and should not be merged as-is.  It can be closed as an
obsolete implementation while this tested current-master version moves
forward under #4585.

PR #4657, opened 2022-06-10
korean translate fix typo

green1052:
fix typo


ainar-g:
Thank you for the contribution, but we only accept internationalization improvements through Crowdin.  Please send your improvement here: <https://crowdin.com/project/adguard-applications/en#/adguard-home>.  Thanks!

PR #4686, opened 2022-06-23
feat: add dns.ipset_file setting

hellodword:
just like the upstream_dns_file

### DIFF
--- a/internal/dnsforward/config.go
+++ b/internal/dnsforward/config.go
@@ -131,6 +131,10 @@ type FilteringConfig struct {
 	//   DOMAIN[,DOMAIN].../IPSET_NAME
 	//
 	IpsetList []string `yaml:"ipset"`
+
+	// IpsetListFileName, if set, points to the file with ipset configuration.
+	// The format is the same as in IpsetList.
+	IpsetListFileName string `yaml:"ipset_file"`
 }
 
 // TLSConfig is the TLS configuration for HTTPS, DNS-over-HTTPS, and DNS-over-TLS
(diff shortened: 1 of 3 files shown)


zhudan:
watch!!

ainar-g:
Good day and apologies for the long wait.  It's generally better to propose a feature in an issue first, but since it's already done, can I ask, what is the problem you are trying to solve?  Is the ipset configuration really getting this large?

hellodword:
> Good day and apologies for the long wait. It's generally better to propose a feature in an issue first, but since it's already done, can I ask, what is the problem you are trying to solve? Is the ipset configuration really getting this large?

For example:

https://github.com/hezhijie0327/GFWList2AGH/tree/main/gfwlist2adguardhome

hellodword:
> I can see that these are lists of upstreams, correct? How is it related to ipsets?

For most usage in my country, `upstream_dns_file` protects our DNS, `ipset` define the whitelist or blacklist to pass through the firewall via VPN.

So the `ipset` often has simliar content and size with the `upstream_dns_file`.

hellodword:
For example:

`upstream_dns_file`:
```
https://doh.pub:443/dns-query
[/github.com/google.com/]tls://8.8.4.4
```

`ipset`:
```
ipset:
- github.com,google.com/ipset-for-vpn
```

ainar-g:
Hm, I see.  We'll review the change then.

internal/dnsforward/config.go
```diff
@@ -129,7 +129,8 @@ type FilteringConfig struct {
 	//
 	//   DOMAIN[,DOMAIN].../IPSET_NAME
 	//
-	IpsetList []string `yaml:"ipset"`
+	IpsetList         []string `yaml:"ipset"`
+	IpsetListFileName string   `yaml:"ipset_file"`
```
ainar-g, line 133: Please add documentation.  Something like:

> IpsetListFileName, if set, points to the file with ipset configuration.  The format is the same as in IpsetList.

internal/dnsforward/dnsforward.go
```diff
@@ -480,7 +481,21 @@ func (s *Server) Prepare(config *ServerConfig) error {
...
+
+		log.Debug("dns: using %d ipset list from file %s", len(ipsets), s.conf.IpsetListFileName)
+	} else {
+		ipsets = s.conf.IpsetList
+	}
+
+	err := s.ipset.init(ipsets)
 	if err != nil {
 		return err
 	}
```
ainar-g, line 501: This increases the [cyclomatic complexity][1] of method `Prepare` too much.  On `master`, we have finally reduced the complexity of all methods in package `dnsforward` to 10 and below.  Please put what is now on lines 484 to 501 to a separate method.  Do not forget about error context.

[1]: https://en.wikipedia.org/wiki/Cyclomatic_complexity

ainar-g (changes requested):
Please also:

1.  Merge the current `master` branch or rebase your branch on top of it, optionally with a squash.

2.  Describe the changes in `CHANGELOG.md`.

ainar-g:
Thanks for the contribution!

PR #4710, opened 2022-07-05
Add dependabot

wrt54g:
I added dependabot for Actions.


ainar-g:
Apologies for the long wait.  Thanks for the contribution, but we manage dependencies manually, both for the backend and the frontend.  (Although, to be fair, the frontend has been slacking with updates, heh.)

Issue #2926, opened 2021-04-06
Listening on IPv6 LLA is not possible


lars-1:
### Issue Details

* **Version of AdGuard Home server:**
  * v0.106.0-b.1
* **How did you install AdGuard Home:**
  * snap
* **How did you setup DNS configuration:**
  * LXD
* **CPU architecture:**
  * AMD64
* **Operating system and version:**
  * Ubuntu 20.04

I am trying to listen on an IPv6 LLA address, but AdGuard won't start if I add it to the configuration yaml.

### Expected Behavior
The configuration should accept IPv6 LLA adresses:
```yaml
bind_hosts:
  - 10.10.10.10
  - fe80::10%eth0
```

### Actual Behavior
For IPv6 LLA you need to specify an interface (e.g. eth0) but AdGuard does not start:
`Job for snap.adguard-home.adguard-home.service failed because the control process exited with error code.`

I tried escaping the percent sign but that resulted in `Couldn't parse config file: invalid IP address: fe80::10%%eth0`.

lars-1:
Bug still exists in v0.106.3.

jdeluyck:
Facing the same issue when trying to get Adguardhome to listen on an ULA, in my case fddf:xxx:xxx::x/64 -> same error. Probably same fix.

EugeneOne1:
@lars-1, hello and sorry for the long response.  We've pushed the latest edge build that should now be able to serve DNS on link-local IPv6 addresses.  Could you please try it and tell if it works properly within your network?

Chialya:
The use of IPv6 LLA addresses is wrong and should not be used! According to the specifications for IPv6, ULA addresses are used for this purpose!
Even if you define the network interface for LLA, end devices that can only reach the DNS via multiple hops can still not communicate to the DNS.
The use of ULA addresses works perfectly.

EugeneOne1:
@lars-1, hello again.  We'll close the issue for now if you don't mind.  It works properly on our test machines, but feel free to reopen the issue if any problem occured with the implementation.

PR #4759, opened 2022-07-18
rfc: wip: switch to netip for go 1.18 and v0.109.0

MartB:
First pull request to this repo, looking forward to your feedback :tea:

This fixes the issue with link local addresses (LLA) not working due to a limitation in go that discards the zone information and does not consider anything with `fe80:: .... %ethX` a valid IP address (see https://github.com/AdguardTeam/AdGuardHome/issues/2926)

Sadly I did not find a way to obtain LLAs from the interface list in "zoned" format, but specifying one for `BindHosts` in the aforementioned format works fine for binding the DNS resolver now! 

I also lifted the artificial limitation in `CheckPort` for "udp" and "tcp" so ipv6 only listening with "udp6" and "tcp6" can also be tested.

Note: This commit will be squashed and rebased in the end.

### DIFF
--- a/internal/home/controlinstall.go
+++ b/internal/home/controlinstall.go
@@ -108,7 +108,7 @@ type checkConfResp struct {
 func (req *checkConfReq) validateWeb(uc aghalg.UniqChecker) (err error) {
 	defer func() { err = errors.Annotate(err, "validating ports: %w") }()
 
-	port := int(req.Web.Port)
+	port := req.Web.Port
 	addPorts(uc, tcpPort(config.BetaBindPort), tcpPort(port))
 	if err = uc.Validate(aghalg.IntIsBefore); err != nil {
 		// Avoid duplicating the error into the status of DNS.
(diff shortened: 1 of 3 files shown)


ainar-g:
Hello, thank you for the contribution, and apologies for the delay.

It's generally better to find or file an issue to ask if a PR is necessary.  Because often there is already a WIP branch somewhere.  A lot of this code will be fully rewritten for the v0.108 cycle, so I'll be honest and say that this PR is likely to be overwritten by our own work.  I'll add it to the milestone, but unfortunately I can't give any time frames.

MartB:
@ainar-g Hey Ainar! Thanks for your reply :)

I made this PR bcs i urgently needed this change for my home network (dynamic IPv6 prefixes, only stable addresses are LLA). 
So i had to port it anyway and it served as a learning opportunity to get to know AdGuardHomes code base and the new go IP implementation.

It is perfectly fine if it gets overwritten in a future release as theres no work "wasted" here. Just wanted to give back after fixing my issue :) If you guys end up having a somewhat production ready WIP branch, just drop me a hint here 👍

Issue #4843, opened 2022-08-16
Update Snap to Ubuntu Core 22 (UC22)


jslawler:
### Prerequisites

- [X] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [X] I have searched other issues and found no duplicates

- [X] I want to request a feature or enhancement and not ask a question

### Description

Ubuntu Core 22 was released [21 APRIL 2022](https://ubuntu.com/blog/what-youre-missing-out-if-you-dont-try-ubuntu-core-22).

Note: UC22 is backwards compatible with UC20, maintaining the same partition layout, so there should be little to no issues with this upgrade.

PR #4844, opened 2022-08-16
Update Snap to Ubuntu Core 22 #4843

jslawler:


ainar-g:
Thanks for the contribution!

Issue #4846, opened 2022-08-18
AdGuard Home fails to migrate config: "unexpected type of dns: map[string]interface {}"


FabioEight:
### Prerequisites

- [X] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [X] I have searched other issues and found no duplicates

- [X] I want to report a bug and not ask a question

### Operating system type

Linux, Other (please mention the version in the description)

### CPU architecture

AMD64

### Installation

GitHub releases or script from README

### Setup

On a router, DHCP is handled by the router

### AdGuard Home version

0.107.10

#### What did you do?
I tried to install AdGuard using the script:

`
curl -s -S -L https://raw.githubusercontent.com/AdguardTeam/AdGuardHome/master/scripts/install.sh | sh -s -- -v
`

#### Expected result
I got:

```

root@FakePi3:~# curl -s -S -L https://raw.githubusercontent.com/AdguardTeam/AdGuardHome/master/scripts/install.sh | sh -s -- -v
starting AdGuard Home installation script
channel: release
operating system: linux
cpu type: amd64
AdGuard Home will be installed into /opt/AdGuardHome
checking tar
script is executed with root privileges
no need to uninstall
downloading package from https://static.adtidy.org/adguardhome/release/AdGuardHome_linux_amd64.tar.gz -> AdGuardHome_linux_amd64.tar.gz
successfully downloaded AdGuardHome_linux_amd64.tar.gz
unpacking package from AdGuardHome_linux_amd64.tar.gz into /opt
successfully unpacked, contents: 
total 35388
-rwxrwxrwx 1 root root 36122624 Aug 17 17:44 AdGuardHome
-rw-rw-rw- 1 root root      587 Aug 17 17:44 AdGuardHome.sig
-rw-r--r-- 1 root root    47183 Aug 17 17:44 CHANGELOG.md
-rw-r--r-- 1 root root    35149 Aug 17 17:44 LICENSE.txt
-rw-r--r-- 1 root root    22788 Aug 17 17:44 README.md
2022/08/18 15:04:58 [info] service: control action: install
2022/08/18 15:04:58 [info] service: started
2022/08/18 15:04:58 [info] Almost ready!
AdGuard Home is successfully installed and will automatically start on boot.
There are a few more things that must be configured before you can use it.
[...]

PR #4848, opened 2022-08-18
regression test for unexpected type of dns: map[string]interface

marco-m:
I stumbled upon the same problem of #4846:

```
go build . &&  ./AdGuardHome -c AdGuardHome.yaml
2022/08/18 14:17:16.611279 [info] AdGuard Home, version
2022/08/18 14:17:16.611523 [info] Upgrade yaml: 12 to 13
2022/08/18 14:17:16.611552 [fatal] unexpected type of dns: map[string]interface {}; want: map[interface {}]interface {}
```

I was not sure how to fix, but after having added a regression test, I saw the fix 4a7b4d03a1d81 by @ainar-g.

The interest of this PR is in the regression test, that AdGuardHome currently doesn't have.

If you disagree, feel free to close.

Thanks for AdGuardHome!

This validates 4a7b4d03a1d81, fix for #4846

### DIFF
--- a/internal/home/testdata/12to13-regression.yaml
+++ b/internal/home/testdata/12to13-regression.yaml
@@ -1,0 +1,3 @@
+schema_version: 12
+dns:
+  port: 53
--- a/internal/home/upgrade.go
+++ b/internal/home/upgrade.go
@@ -240,8 +240,9 @@ func upgradeSchema3to4(diskConf yobj) error {
 
 // Replace "auth_name", "auth_pass" string settings with an array:
 // users:
-// - name: "..."
-//   password: "..."
+//   - name: "..."
+//     password: "..."
+//
 // ...
 func upgradeSchema4to5(diskConf yobj) error {
 	log.Printf("%s(): called", funcName())
@@ -288,16 +289,18 @@ func upgradeSchema4to5(diskConf yobj) error {
 
 // clients:
 // ...
-//   ip: 127.0.0.1
-//   mac: ...
+//
+//	ip: 127.0.0.1
+//	mac: ...
 //
 // ->
 //
 // clients:
 // ...
-//   ids:
-//   - 127.0.0.1
-//   - ...
+//
+//	ids:
+//	- 127.0.0.1
+//	- ...
 func upgradeSchema5to6(diskConf yobj) error {
 	log.Printf("%s(): called", funcName())
 
@@ -355,19 +358,21 @@ func upgradeSchema5to6(diskConf yobj) error {
 }
 
 // dhcp:
-//   enabled: false
-//   interface_name: vboxnet0
-//   gateway_ip: 192.168.56.1
-//   ...
+//
+//	enabled: false
+//	interface_name: vboxnet0
+//	gateway_ip: 192.168.56.1
+//	...
 //
 // ->
 //
 // dhcp:
-//   enabled: false
-//   interface_name: vboxnet0
-//   dhcpv4:
-//     gateway_ip: 192.168.56.1
-//     ...
+//
+//	enabled: false
+//	interface_name: vboxnet0
+//	dhcpv4:
+//	  gateway_ip: 192.168.56.1
+//	  ...
 func upgradeSchema6to7(diskConf yobj) error {
 	log.Printf("Upgrade yaml: 6 to 7")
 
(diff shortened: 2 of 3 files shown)


internal/home/upgrade.go
```diff
@@ -240,8 +240,9 @@ func upgradeSchema3to4(diskConf yobj) error {
 
 // Replace "auth_name", "auth_pass" string settings with an array:
 // users:
-// - name: "..."
-//   password: "..."
+//   - name: "..."
+//     password: "..."
```
ainar-g, line 244: Just to be clear, these format changes have been made by the Go 1.19 comment formatter as opposed to manual reformatting, right?

internal/home/upgrade.go
```diff
@@ -720,7 +719,7 @@ func upgradeSchema12to13(diskConf yobj) (err error) {
 	var dns yobj
 	dns, ok = dnsVal.(yobj)
 	if !ok {
-		return fmt.Errorf("unexpected type of dns: %T", dnsVal)
+		return fmt.Errorf("unexpected type of dns: %T; want: %T", dnsVal, dns)
```
ainar-g, line 722: There are a few other similar errors throughout the recent migrations, and it would be nice to have them all in a unified format.

internal/home/upgrade_test.go
```diff
@@ -558,6 +560,18 @@ func TestUpgradeSchema12to13(t *testing.T) {
 	}
 }
 
+func TestUpgradeSchema12to13Regression(t *testing.T) {
```
ainar-g, line 563: I feel like a better regression test would try to migrate a configuration file going all the way back to version `1`.  Please add a TODO [assigned to me][1] about that.

[1]: https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Text.md#li-2249ae11

internal/home/upgrade_test.go
```diff
@@ -558,6 +560,18 @@ func TestUpgradeSchema12to13(t *testing.T) {
...
+	body, err := os.ReadFile("testdata/12to13-regression.yaml")
+	assert.NoError(t, err)
```
ainar-g, line 565: This one should be a `require`, and also split from the rest by an empty line.

internal/home/upgrade_test.go
```diff
@@ -558,6 +560,18 @@ func TestUpgradeSchema12to13(t *testing.T) {
...
+	diskConf := yobj{}
+	err = yaml.Unmarshal(body, &diskConf)
+	assert.NoError(t, err)
+
+	err = upgradeSchema12to13(diskConf)
+
+	assert.NoError(t, err)
```
ainar-g, line 572: We usually group `assert`s and `requires` together with the statements where the data was acquired, so please remove the empty line above.

internal/home/testdata/12to13-regression.yaml
```diff
@@ -0,0 +1,3 @@
+schema_version: 12
```
ainar-g, line 1: Here, too, a TODO should be added to make it a more faithful representation of a `config.yaml` at version `12`.  With YAML, you can never be too defensive, heh.

ainar-g (changes requested):
Thanks for the contribution!  We have some comments regarding formatting and comments.

internal/home/upgrade.go
marco-m: As sorry I missed this :-(
Yes, it is the Go 1.10 comment formatter :-/.
I can remove this if you prefer?

internal/home/upgrade.go
marco-m: I had added this to help me understand better what the code expected, but I agree it is more a debug printf.

I am not sure I understand your comment thou: do you want me to revert this change or to extend it to all the similar "unexpected ..." ?

internal/home/upgrade.go
ainar-g: No, you don't need to, I'm just making sure.

internal/home/upgrade.go
ainar-g: I mean that it's better to replace *all* instances of errors like `unexpected type of %s: %T` with this new form that includes the expected type as opposed to just here.

marco-m:
Closing since no time to dedicate to it. Feel free to resurrect if you want. Thanks for Adguard!

Issue #4366, opened 2022-03-06
AdGuard Home logo for light/dark themes


scripthunter7:
Hello!

The README currently uses a logo image that is dark in all cases, so this logo will not look good if the dark mode of GitHub is used:

![Current status](https://i.ibb.co/nMSRzNm/aghome-readme-issue.png)

However, the good news is that GitHub has been [supporting a feature](https://twitter.com/github/status/1463556704436637707) that allows you to display an image for a specific color theme since last November.

It can be used as in the example below:
- `IMAGE_URL#gh-light-mode-only`
- `IMAGE_URL#gh-dark-mode-only`

- For light mode (current dark logo):
  ```html
    <p align="center">
      <img src="https://cdn.adguard.com/public/Adguard/Common/adguard_home.svg#gh-light-mode-only" width="300px" alt="AdGuard Home" />
    </p>
  ```
- For dark mode (white logo):
  ```html
    <p align="center">
      <img src="https://cdn.adguard.com/public/Adguard/Common/adguard_home_white.svg#gh-dark-mode-only" width="300px" alt="AdGuard Home" />
    </p>
  ```

  - You can even use this svg:
    ```html
    <?xml version="1.0" encoding="utf-8"?>
    <svg width="164" height="41" viewBox="0 0 164 41" xmlns="http://www.w3.org/2000/svg">
      <g fill-rule="evenodd">
[...]

PR #4860, opened 2022-08-24
Color scheme based logo

scripthunter7:
https://github.com/AdguardTeam/AdGuardHome/issues/4366

Preview: https://github.com/scripthunter7/AdGuardHome/blob/fix/4366/README.md


ainar-g:
Thanks for doing this!

@ameshkov, as far as I know, we usually keep logos and such on our CDNs, while this PR adds the two versions of the logo into the repository itself.  What do you think?

ainar-g:
Thanks for the contribution!

PR #4879, opened 2022-08-30
Add More Disney Service

eduardoaddad:


ainar-g:
We already have Hulu as a separate blockable service.  I'm not sure if we want to add it to Disney, to be honest.

@ameshkov, am I correct that we typically prefer to avoid rule collisions in blocked services?

ameshkov:
I suppose Hulu should be a separate service, not just a part of Disney

PR #4891, opened 2022-09-04
Add some domains

MkQtS:
Added "b23.tv, bilicdn1.com, bilicdn2.com" for bilibili, "douyinpic.com, douyinstatic.com, douyinvod.com" for tiktok.
Removed one redundant "||ixigua.com^" which appeared twice, also sorted tiktok.


ainar-g:
Thank you for the contribution!

PR #4912, opened 2022-09-12
disable notify in forked repository

hellodword:
This is for us who fork this repository, and when we push a commit, we will always get two errors in github actions.

For example:

https://github.com/hellodword/AdGuardHome/actions?query=is%3Afailure


hellodword:
Thanks!

PR #4936, opened 2022-09-20
Fix #4925: SSL Lab result capped to B

ssrahul96:
Issue https://github.com/AdguardTeam/AdGuardHome/issues/4925

The Cipher `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA` needs to be supported in order to pass `Forward Secrecy` test of [SSL Labs](https://www.ssllabs.com/ssltest/analyze.html)

SSL Lab result after supporting this Cipher

image

Ref : https://www.namecheap.com/support/knowledgebase/article.aspx/9653/38/how-to-check-whether-the-server-supports-forward-secrecy/

This Cipher is allowed in [d.adguard-dns.com](d.adguard-dns.com) as well, 
SSL Labs report : [link](https://www.ssllabs.com/ssltest/analyze.html?d=d.adguard%2ddns.com&s=94.140.14.49&latest)

PR #4941, opened 2022-09-22
Added support for hsts

ssrahul96:
HSTS is added to get `A+` in SSL Lab Report

image

based on the feedback from https://github.com/AdguardTeam/AdGuardHome/pull/4936#issuecomment-1253259770


internal/home/control.go
```diff
@@ -320,6 +320,16 @@ func handleHTTPSRedirect(w http.ResponseWriter, r *http.Request) (ok bool) {
...
 	}
 
+	// Add headers for HSTS
+	// This informs browsers that the site should only be accessed using HTTPS,
+	// and that any future attempts to access it using HTTP should automatically be
+	// converted to HTTPS.
+	//
+	// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
+	if config.TLS.ForceHTTPS {
+		w.Header().Add("Strict-Transport-Security", "max-age=63072000; includeSubDomains")
```
ainar-g, line 330: 1.  From what I've read in the MDN article and in some recommendations, HSTS headers should only be set in HTTPS responses.  It looks like this code will set it regardless of whether or not the request is HTTPS?

2.  Let's limit max-age to 1 year for now.

ainar-g (changes requested):

internal/home/control.go
ssrahul96: > 1. From what I've read in the MDN article and in some recommendations, HSTS headers should only be set in HTTPS responses.  It looks like this code will set it regardless of whether or not the request is HTTPS?
> 2. Let's limit max-age to 1 year for now.

@ainar-g , ForceHTTPS, validation is added, which could be enabled, only if https is enabled, and force http -> https redirection is done in AG

Will change it to 1 year

ainar-g: @ssrahul96, consider the following case: if a client accesses AGH using plain HTTP, they are redirected, but the redirection response will contain the header, thus transferring it over plain HTTP.  This is the situation I'd like to avoid.

ssrahul96: this will occur for the very 1st api call, from next call onwards hsts would come into picture, anyhow will see, if i can handle it

ssrahul96: moved the hsts header block after `https` redirection, that should resolved this

Harvester57:
Ping @ainar-g

Harvester57:
Ping @ainar-g

Harvester57:
Up !

PR #4964, opened 2022-09-29
Add Warner Bros. Discovery Service

eduardoaddad:
Add HBO Max, Discover+, CNN GO and TNT GO

### DIFF
--- a/client/src/helpers/constants.js
+++ b/client/src/helpers/constants.js
@@ -347,6 +347,10 @@ export const SERVICES = [
         id: 'youtube',
         name: 'YouTube',
     },
+    {
+        id: 'warnerbrosdiscovery',
+        name: 'Warner Bros. Discovery',
+    },
 ];
 
 export const SERVICES_ID_NAME_MAP = SERVICES.reduce((acc, { id, name }) => {
--- a/internal/filtering/blocked.go
+++ b/internal/filtering/blocked.go
@@ -379,6 +379,20 @@ var servicesData = []svc{{
 		"||dreamcast.hk^",
 		"||hdslb.com^",
 	},
+}, {
+	name: "warnerbrosdiscovery",
+	rules: []string{
+		"||hbomax.com^",
+		"||hbonow.com^",
+		"||hbonow.com^",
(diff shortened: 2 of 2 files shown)


internal/filtering/blocked.go
```diff
@@ -379,6 +379,20 @@ var servicesData = []svc{{
...
+		"||hbomax.com^",
+		"||hbonow.com^",
+		"||hbonow.com^",
+		"||hbomax-images.warnermediacdn.com^",
+		"||art-gallery-latam.api.hbo.com^",
+		"||artist.api.cdn.hbo.com^",
+		"||hbogoasia.id^",
+		"||tntgo.tv^",
+		"||discoveryplus.com^",
+		"||go.cnn.com^",
```
ainar-g, line 394: Some older rule sets aren't alphabetically sorted, unfortunately, but we strive to keep rules sorted.

client/src/helpers/constants.js
```diff
@@ -347,6 +347,10 @@ export const SERVICES = [
         id: 'youtube',
         name: 'YouTube',
     },
+    {
+        id: 'warnerbrosdiscovery',
+        name: 'Warner Bros. Discovery',
+    },
```
ainar-g, line 353: `client/src/components/ui/Icons.js` should be changed as well.

ainar-g (changes requested):
Thanks for the contribution!  Here are some comments:

internal/filtering/blocked.go
eduardoaddad: done

ainar-g (changes requested):
See the comment about icons.

eduardoaddad:
> See the comment about icons.

How do I add icons?

ainar-g:
Apologies for the long wait.  We've moved to using the same list of blockable services that AdGuard DNS uses, https://github.com/AdguardTeam/HostlistsRegistry.  Please open a PR there, and once accepted, it will be loaded to the next AdGuard Home release as well as in AdGuard DNS.  Again, thanks for the contribution and apologies for not reacting quicker.

PR #4990, opened 2022-10-04
Fix #4925: Allow users to choose TLS ciphers

### LINKED ISSUE #4925: ssllab result capped to B
### Prerequisites

### Operating system type

Linux, OpenWrt

### CPU architecture

AMD64

### Installation

Docker

### Setup

On one machine

### AdGuard Home version

v0.108.0-b.15

#### What did you do?

SSL Lab test result is `B`

ref : https://blog.qualys.com/product-tech/2018/02/02/forward-secrecy-authenticated-encryption-and-robot-grading-update

Seems like additional config changes at go http listener

#### Screenshots (if applicable)

image

#### Additional information

Reference Article to check : https://www.namecheap.com/support/knowledgebase/article.aspx/9653/38/how-to-check-whether-the-server-supports-forward-secrecy/

tried `openssl s_client -connect example.com:443 -cipher "DHE, EDH"` , the request was failing,

I dont ve idea about these ciphers. I tried for few sites with ssl lab report `A+` the above command was passing

ssrahul96:
Issue https://github.com/AdguardTeam/AdGuardHome/issues/4925

The Cipher `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA` needs to be supported in order to pass `Forward Secrecy` test of [SSL Labs](https://www.ssllabs.com/ssltest/analyze.html)

SSL Lab result after supporting this Cipher

image

Ref : https://www.namecheap.com/support/knowledgebase/article.aspx/9653/38/how-to-check-whether-the-server-supports-forward-secrecy/

This Cipher is allowed in [d.adguard-dns.com](d.adguard-dns.com) as well, 
SSL Labs report : [link](https://www.ssllabs.com/ssltest/analyze.html?d=d.adguard%2ddns.com&s=94.140.14.49&latest)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,18 @@ and this project adheres to
 
 See also the [v0.107.16 GitHub milestone][ms-v0.107.15].
 
-[ms-v0.107.16]:   https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=1
--->
+[ms-v0.107.16]:   https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=
 
+### Added
+
+- The new optional `tls.override_tls_ciphers` property list, which can be set in
+  the configuration file.  It allows overriding TLS Ciphers that are used for
+  https listeners ([#4925])
+
+[#4925]: https://github.com/AdguardTeam/AdGuardHome/issues/4925
+
+
+-->
 
 
 ## [v0.107.15] - 2022-10-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,6 @@ and this project adheres to
 See also the [v0.107.16 GitHub milestone][ms-v0.107.15].
 
 [ms-v0.107.16]:   https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=
-
 -->
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ and this project adheres to
 
 ## [Unreleased]
 
+
+### Added
+
+- The new optional `tls.override_tls_ciphers` property list, which can be set in
+  the configuration file.  It allows overriding TLS Ciphers that are used for
+  https listeners ([#4925])
+
+[#4925]: https://github.com/AdguardTeam/AdGuardHome/issues/4925
+
 <!--
 ## [v0.108.0] - TBA (APPROX.)
 -->
@@ -23,15 +32,6 @@ and this project adheres to
 See also the [v0.107.16 GitHub milestone][ms-v0.107.15].
 
 [ms-v0.107.16]:   https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=
-
-### Added
-
-- The new optional `tls.override_tls_ciphers` property list, which can be set in
-  the configuration file.  It allows overriding TLS Ciphers that are used for
-  https listeners ([#4925])
-
-[#4925]: https://github.com/AdguardTeam/AdGuardHome/issues/4925
-
 
 -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to
 
 See also the [v0.107.16 GitHub milestone][ms-v0.107.15].
 
-[ms-v0.107.16]:   https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=
+[ms-v0.107.16]: https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=1
 -->
 
 
--- a/internal/aghtls/aghtls.go
+++ b/internal/aghtls/aghtls.go
@@ -34,11 +34,14 @@ func SaferCipherSuites() (safe []uint16) {
 	return safe
 }
 
-func UserPreferredCipherSuites(ciphers []string) (userCiphers []uint16) {
+// ParseCipherIDs returns a set of cipher suites with the cipher names provided
+func ParseCipherIDs(ciphers []string) (userCiphers []uint16) {
 	for _, s := range tls.CipherSuites() {
 		if slices.Contains(ciphers, s.Name) {
 			userCiphers = append(userCiphers, s.ID)
 			log.Debug("user specified cipher : %s, ID : %d", s.Name, s.ID)
+		} else {
+			log.Error("unknown cipher : %s ", s)
 		}
 	}
 
--- a/internal/aghtls/aghtls.go
+++ b/internal/aghtls/aghtls.go
@@ -3,9 +3,7 @@ package aghtls
 
 import (
 	"crypto/tls"
-
-	"github.com/AdguardTeam/golibs/log"
-	"golang.org/x/exp/slices"
+	"fmt"
 )
 
 // SaferCipherSuites returns a set of default cipher suites with vulnerable and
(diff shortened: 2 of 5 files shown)


ssrahul96:
fix for https://github.com/AdguardTeam/AdGuardHome/issues/4925

internal/aghtls/aghtls.go
```diff
@@ -28,3 +33,14 @@ func SaferCipherSuites() (safe []uint16) {
 
 	return safe
 }
+
+func UserPreferredCipherSuites(ciphers []string) (userCiphers []uint16) {
```
ainar-g, line 37: 1.  The name of this function isn't really informative.  I feel like a better name would be something like `ParseCipherIDs`.

2.  This function requires documentation.

internal/aghtls/aghtls.go
```diff
@@ -28,3 +33,14 @@ func SaferCipherSuites() (safe []uint16) {
...
+	for _, s := range tls.CipherSuites() {
+		if slices.Contains(ciphers, s.Name) {
+			userCiphers = append(userCiphers, s.ID)
+			log.Debug("user specified cipher : %s, ID : %d", s.Name, s.ID)
+		}
+	}
```
ainar-g, line 43: Currently, this function ignores unknown ciphers, which seems like a mistake.  It should return an informative error instead.

internal/dnsforward/config.go
```diff
@@ -165,6 +165,9 @@ type TLSConfig struct {
 	cert tls.Certificate
 	// DNS names from certificate (SAN) or CN value from Subject
 	dnsNames []string
+
+	// ciphers specified by user
+	TLSCiphers []string `yaml:"tls_ciphers" json:"-"`
```
ainar-g, line 170: 1.  Please follow the [Go conventions on documenting](https://go.dev/doc/comment).  Unfortunately, a lot of older code doesn't follow them, but we strive to fix that and write all new code with the conventions in mind.

2.  The documentation here should mention that if the slice is empty or nil, the default ciphers are used.

3.  I feel like a better name would communicate that this field should generally not be used.  Something like `override_tls_ciphers`.

internal/home/web.go
```diff
@@ -57,6 +57,9 @@ type webConfig struct {
 	firstRun bool
 
 	serveHTTP3 bool
+
+	// ciphers specified by user
+	tlsCiphers []string
```
ainar-g, line 62: It's better to [parse structured data as soon as possible](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), so it's better to have `[]uint16` here as well.

Also, the documentation should be improved, like in the other cases, and the slice should be moved higher to prevent the structure from having suboptimal field alignment.

ainar-g (changes requested):
1.  Please document the changes in `CHANGELOG.md`.

2.  I don't see the user-overriden ciphers being used in package `dnsforward`.  Is this intentional?

ssrahul96:
> 
This change mainly focused on `http` server only, i am not sure, whether the same is applicable for DOT as well, let me know if its required.

internal/aghtls/aghtls.go
ssrahul96: logged unknown ciphers

internal/dnsforward/config.go
ssrahul96: added a function `getTLSCiphers()`

internal/aghtls/aghtls.go
ssrahul96: changes as per the recommendation

internal/home/web.go
ssrahul96: changes done

CHANGELOG.md
```diff
@@ -22,9 +22,18 @@ and this project adheres to
...
--->
+[ms-v0.107.16]:   https://github.com/AdguardTeam/AdGuardHome/milestone/52?closed=
+
+### Added
+
+- The new optional `tls.override_tls_ciphers` property list, which can be set in
+  the configuration file.  It allows overriding TLS Ciphers that are used for
+  https listeners ([#4925])
 
+[#4925]: https://github.com/AdguardTeam/AdGuardHome/issues/4925
```
ainar-g, line 33: New changes are added to the Unreleased section.  Please see the [KeepAChangelog recommendations](https://keepachangelog.com/en/1.0.0/).

internal/aghtls/aghtls.go
ainar-g: Logging is not strict enough.  Please return an error and process it making AGH quit if the user has entered an incorrect cipher ID.

internal/aghtls/aghtls.go
```diff
@@ -28,3 +33,17 @@ func SaferCipherSuites() (safe []uint16) {
 
 	return safe
 }
+
+// ParseCipherIDs returns a set of cipher suites with the cipher names provided
+func ParseCipherIDs(ciphers []string) (userCiphers []uint16) {
+	for _, s := range tls.CipherSuites() {
+		if slices.Contains(ciphers, s.Name) {
+			userCiphers = append(userCiphers, s.ID)
+			log.Debug("user specified cipher : %s, ID : %d", s.Name, s.ID)
```
ainar-g, line 42: I'm not sure that logging belongs in here at all.  It's better to log custom ciphers, if they are present, before parsing them.

internal/home/home.go
```diff
@@ -887,3 +888,16 @@ type jsonError struct {
 	// Message is the error message, an opaque string.
 	Message string `json:"message"`
 }
+
+// getTLSCiphers check for overriden tls ciphers, if the slice is
+// empty, then default safe ciphers are used
+func getTLSCiphers() []uint16 {
+	var cipher []uint16
```
ainar-g, line 895: We prefer to use named returns in such cases.  See the [guidelines item](https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Go.md#li-da463e1c).

ainar-g (changes requested):

internal/aghtls/aghtls.go
ainar-g: I think that logging should not be present in what is essentially a general-purpose function.  For some context, the current plan is that most of package `aghtls` will be imported into our `golibs` module one day.

ainar-g:
After looking deeper into it, refactoring of our TLS stuff was required.  It is now done, and your contribution is merged.  Thanks!

Harvester57:
What is the expected format needed to pass to the tls.override_tls_ciphers option ?

Harvester57:
Okay, it is documented in the Wiki, so for those wondering, the option need an array of TLS ciphers. For example:

```
override_tls_ciphers:
    - TLS_AES_256_GCM_SHA384
    - TLS_CHACHA20_POLY1305_SHA256
    - TLS_AES_128_GCM_SHA256
    - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
    - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
    - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
    - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
    - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
    - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
```

Available strings are documented in the Go documentation: https://pkg.go.dev/crypto/tls#pkg-constants

Now another question, isn't the server supposed to respect the array order ? This is not the case for the array provided in example, SSL Server Test does return the following:

<html>
<body>


# TLS 1.3 (suites in server-preferred order)
--
TLS_AES_128_GCM_SHA256 (0x1301)   ECDH x25519 (eq. 3072 bits RSA)   FS | 128
TLS_AES_256_GCM_SHA384 (0x1302)   ECDH x25519 (eq. 3072 bits RSA)   FS | 256
TLS_CHACHA20_POLY1305_SHA256 (0x1303)   ECDH x25519 (eq. 3072 bits RSA)   FS | 256P


</body>
</html>

ainar-g:
@Harvester57, the Go implementation of TLS [uses its own automatic ordering](https://go.dev/blog/tls-cipher-suites).

Harvester57:
Thank you for this information ! Makes sens !

PR #4993, opened 2022-10-04
Fix #4986 Web UI HTTP/3 not working

### LINKED ISSUE #4986: Web UI HTTP/3 not working
### Prerequisites

### Operating system type

Linux, Other (please mention the version in the description)

### CPU architecture

AMD64

### Installation

### Setup

On one machine

### AdGuard Home version

v0.108.0-b.17

#### What did you do?

Enabled HTTP/3 support for the web UI with `serve_http3` set to `true`, and checked with Google Chrome stable (v106) and canary (v108) with `Experimental QUIC support` set to `Enabled` in `chrome://flags` if it is working.

#### Expected result

Request should show `h3` in the Dev Tools > Network > Protocol column, but it is still showing `h2`

#### Actual result

Request showing the protocol used is `h2`

#### Screenshots (if applicable)

Not working
H3

Same web browser, when connecting to youtube.com
image

#### Additional information

HTTP/3 for upstream servers seems to be working fine, I was able to force the Google DNS with `h3://8.8.8.8/dns-query`

ssrahul96:
Fix https://github.com/AdguardTeam/AdGuardHome/issues/4986
Even though the udp listener was listening on :443, the traffic was not sent to it, reason being the browser does not know that the website supports `h3`. 

reference : [link](https://www.smashingmagazine.com/2021/09/http3-practical-deployment-options-part3/#alt-svc)

So adding [Alt-Svc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Alt-Svc) in the header, let the browser know that http3 is enabled

After adding the header 
image

`h3` is used
image

### DIFF
--- a/internal/home/control.go
+++ b/internal/home/control.go
@@ -320,6 +320,16 @@ func handleHTTPSRedirect(w http.ResponseWriter, r *http.Request) (ok bool) {
 		return false
 	}
 
+	// Let the browser know that server support HTTP/3
+	// max-age is set to default (24 hourd)
+	//
+	// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Alt-Svc
+	// TODO: take max-age from config and set
+	if config.DNS.ServeHTTP3 {
+		altSvc := fmt.Sprintf(`h3=":%[1]v";`, config.TLS.PortHTTPS)
+		w.Header().Set("Alt-Svc", altSvc)
+	}
+
 	if r.TLS == nil && web.forceHTTPS {
 		hostPort := host
 		if port := web.conf.PortHTTPS; port != defaultPortHTTPS {
@@ -345,11 +355,6 @@ func handleHTTPSRedirect(w http.ResponseWriter, r *http.Request) (ok bool) {
 	originURL := &url.URL{
 		Scheme: aghhttp.SchemeHTTP,
 		Host:   r.Host,
-	}
-
-	if config.DNS.ServeHTTP3 {
-		altSvc := fmt.Sprintf(`h3=":%[1]v"; ma=2592000,h3-29=":%[1]v"; ma=2592000,h3-Q050=":%[1]v";ma=2592000,h3-Q046=":%[1]v"; ma=2592000,h3-Q043=":%[1]v"; ma=2592000,quic=":%[1]v"; ma=2592000; v="46,43"`, config.TLS.PortHTTPS)
-		w.Header().Set("Alt-Svc", altSvc)
 	}
 
 	w.Header().Set("Access-Control-Allow-Origin", originURL.String())
--- a/internal/home/control.go
+++ b/internal/home/control.go
@@ -348,7 +348,8 @@ func handleHTTPSRedirect(w http.ResponseWriter, r *http.Request) (ok bool) {
 	}
 
 	if config.DNS.ServeHTTP3 {
-		w.Header().Set("Alt-Svc", `h3=":443"; ma=2592000,h3-29=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"`)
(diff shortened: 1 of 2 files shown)


flavorgold1:
What if the port has been set to something other than 443?

ssrahul96:
> 

`443` is the standard port for `https`. we can get the port from config file and set, but do we really need to do it?

ishanjain28:
Hey @ssrahul96 Please read the port from config file. Since agh supports this, People may be accessing web ui from non-standard ports and we should try to make http3 work for every one.

ssrahul96:
@ishanjain28 @flavorgold1 , changes done to refer https port from config file

flavorgold1:
Also, HTTP3 might be used without SSL/HTTPS.

And, AdGuardHome may reside behind a reverse-proxy that is accessing it via plain HTTP3 (non-SSL).

So, we must ensure compatibility with that scenario, unless I'm wrong about something here?

flavorgold1:
@ssrahul96, you're right, I'm wrong.

https://www.cloudflare.com/learning/performance/what-is-http3/

It seems that SSL/TLS/encryption is required by HTTP3.

Thanks.

Harvester57:
@ainar-g could you check this merge request please ? :)

7ArxS2PrqAUF9Gom:
I assume the port can be custom, but protocol has to be UDP?

ssrahul96:
> I assume the port can be custom, but protocol has to be UDP? How exactly is alt-svc added to browser header? Firefox already has "network.http.http3.enable" option.

@7ArxS2PrqAUF9Gom , `network.http.http3.enable` option in firefox or `enable-quic` flag in chrome is on the client side, so that it can use the `h3` protocol, 

But for the client to know that the server supports `h3`, `Alt-Svc` is set in header,  

Refer the article in description for more details

7ArxS2PrqAUF9Gom:
> > I assume the port can be custom, but protocol has to be UDP? How exactly is alt-svc added to browser header? Firefox already has "network.http.http3.enable" option.
> 
> @7ArxS2PrqAUF9Gom , `network.http.http3.enable` option in firefox or `enable-quic` flag in chrome is on the client side, so that it can use the `h3` protocol,
> 
> But for the client to know that the server supports `h3`, `Alt-Svc` is set in header,
> 
> Refer the article in description for more details

I don't understand where to place alt-svc header and how... I just want to use HTTP3 for WebUI

flavorgold1:
TPTB have to approve the merge request in order for this functionality to be fully active.

internal/home/control.go
```diff
@@ -346,6 +346,12 @@ func handleHTTPSRedirect(w http.ResponseWriter, r *http.Request) (ok bool) {
 		Scheme: aghhttp.SchemeHTTP,
 		Host:   r.Host,
 	}
+
+	if config.DNS.ServeHTTP3 {
+		altSvc := fmt.Sprintf(`h3=":%[1]v"; ma=2592000,h3-29=":%[1]v"; ma=2592000,h3-Q050=":%[1]v";ma=2592000,h3-Q046=":%[1]v"; ma=2592000,h3-Q043=":%[1]v"; ma=2592000,quic=":%[1]v"; ma=2592000; v="46,43"`, config.TLS.PortHTTPS)
```
ainar-g, line 351: 1.  This string is against our [line length guidelines](https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Go.md#li-6e457108).

2.  Do we really need to enumerate all these drafts?  Wouldn't just `h3` suffice?

3.  I'd prefer `ma` to be set by default to a lower value, for example 24 h.  A TODO about considering making it configurable would be nice as well.

4.  Shouldn't this header be set on every response?  Because the way this is written, it is not set when redirecting to the HTTPS version.

ainar-g (changes requested):
Please also document the change in `CHANGELOG.md`.

ainar-g:
> @ainar-g , will work on the fetching `max-age` value from config and raise a separate PR

That is not necessary for now, I think.  You may open a new issue about that though.  We'll see if people actually need that first.

ssrahul96:
track https://github.com/AdguardTeam/AdGuardHome/issues/5011 for making `max-age` configurable

ainar-g:
@ssrahul96, thanks for the contribution!  We've made additional improvements and merged this PR.

GY8VSdYYzvL8-K6T:
I still don't understand this... I enable HTTPS/3 in Firefox and test that it works on sites like YouTube by pressing F12 and viewing Network tab, but it doesn't when I try to login to AGH UI. It just uses HTTPS/2. In AGH YAML file I made sure "serve_http3: true". What else am I supposed to do? I don't understand where to edit the Alt-Header information? In AdGuardHome.yaml file? In Firefox URL?

PR #5033, opened 2022-10-12
Update filters.json

clem9669:
Hi 👋 

I have added a FR specific list for blacklisting domains freshly registered to AFNIC (.fr), mostly phishing domains.

Have a look on https://red.flag.domains for details 

Cheers

### DIFF
--- a/client/src/helpers/filters/filters.json
+++ b/client/src/helpers/filters/filters.json
@@ -96,6 +96,12 @@
             "homepage": "https://github.com/DandelionSprout/adfilt",
             "source": "https://raw.githubusercontent.com/DandelionSprout/adfilt/master/NorwegianExperimentalList%20alternate%20versions/NordicFiltersAdGuardHome.txt"
         },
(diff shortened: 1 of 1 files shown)


ameshkov:
@clem9669 hi, the right way would be to add this list here instead: https://github.com/AdguardTeam/HostlistsRegistry

AdGuard Home soon to start using that repository as a source.

clem9669:
@ameshkov @ainar-g I will push to the other repo. 

Should I add in this file https://github.com/AdguardTeam/HostlistsRegistry/blob/main/assets/filters.json ? 
Should I create a new folder for France https://github.com/AdguardTeam/HostlistsRegistry/tree/main/filters/regional ?

ainar-g:
@clem9669, from what I understand, `assets/filters.json` shouldn't be directly modified, but I may be wrong.

@ameshkov, I feel like a better contribution guide is required in that repo.  There is enough info about *what* to add, but not a lot about *how* to add it.

Issue #2290, opened 2020-11-10
Reconsider the query log format


ainar-g:
<p>The&#xA0;current way in&#xA0;which we handle query logs has several issues:
it's racy, files can get pretty large, the&#xA0;search algorithm is
too&#xA0;complex, and&#xA0;we can only do basic filtering and&#xA0;very little
aggregation.&#x2003;But&#xA0;we also&#xA0;don't have a&#xA0;lot
of&#xA0;alternatives.&#x2003;Ideally we would like to&#xA0;have something
like&#xA0;an&#xA0;SQL database, <i>à&#xA0;la</i> <i>SQLite</i>, but&#xA0;that
would require <code>cgo</code>, which&#xA0;comes with&#xA0;its own bag
of&#xA0;issues.&#x2003;Package
<a href="https://pkg.go.dev/modernc.org/sqlite"><code>modernc.org/sqlite</code></a>
seems interesting, but&#xA0;it's a&#xA0;work-in-progress and&#xA0;currently
doesn't support all our platforms.</p>

<p>I&#xA0;will update the&#xA0;issue when&#xA0;we'll think of&#xA0;something
better.&#x2003;Perhaps <i>SQLite</i> is not&#xA0;completely
out&#xA0;of&#xA0;the&#xA0;question if&#xA0;we find a way to cross-compile it
without&#xA0;too much pain for&#xA0;the&#xA0;developers.</p>

ainar-g:
<p>Package&#xA0;<a href="https://modernc.org/ql"><code>modernc.org/ql</code></a>
is also an&#xA0;interesting&#xA0;alternative.</p>

ameshkov:
Tbh, there are a couple of advantages provided by using simple text-based storage:

1. Logs can be easily rotated (not like now, though, but anyway)
2. Logs can be loaded and analyzed by other software (ELK, Splunk, etc)

They should also be considered when we choose what to do next.

ZeroClover:
Perhaps consider using [Blevesearch](https://github.com/blevesearch/bleve) to index BoltDB to enhance queries (more filters) and [speed up queries](https://github.com/blevesearch/bleve-bench).

Blevesearch is a native Golang application.

However, it still does not help with write performance and consumes more disk space.

Using TSDB may help performance, but it can add significant complexity.

ainar-g:
@ameshkov,

[...]

PR #5099, opened 2022-11-04
QueryLog Interface

elee1766:
I have been looking for a while to reduce the footprint of the logs on my machine, while still keeping them. 

this pr is the start of that work, and does the following:

1. splits querylog package into two, "logs", and "jsonfile". jsonfile contains things specifically related to the existing on-disk format, while logs includes http handler, interfaces, and shared structs for a "logs.Api implementation" to use.
2. changes the HTTP to use a properly tagged struct with omitempty tags which then marshals to json, instead of populating a map. 
3. Extends the "Store" interface to include methods "ApplyConfig", "ConfigInfo", and "Clear", interface "Searcher" which uses struct "SearchParams" to get result "LogsPayload", and interface "Api", a Search + Store. 

my idea is that one could implement the "Store" and "Searcher" interfaces for different sorts of storage and indexing backends. 

This would allow people to create their own log storage & search implementations.  looking at #2290, it seems that people are eager to have custom log sinks, but the existing implementation makes it hard to add a new one. 

I think more thought needs to be put into an actual interface, along with the configuration format, but I do not feel like the qualified person to do that. so my goal in this PR was mostly to keep compatibility with the existing logging implementation, keeping the changes simple. hopefully we can move from here to something better. that said, things have to start somewhere, so I thought I would make the PR anyways. 

new to this project, so if i did anything wrong, please let me know. happy to make any changes / take any suggestions


Sil3ntVip3r:
I compared this branch with current `master` and #2290. The goal remains
relevant, but this 33-file split is built around the pre-current query-log
layout and now conflicts throughout. Current master already exposes a
narrower `querylog.QueryLog` service boundary, while storage/search
interchangeability and an indexed backend are being discussed under #2290.

The interfaces here are useful design history, but the implementation should
not be rebased mechanically; a current proposal needs to start from the
existing service boundary, preserve the HTTP/config contract, and include
migration and backend-parity tests. This PR can be closed as superseded by
the current architecture/discussion.

elee1766:
oh lord it's been nearly four years.

yeah, this should be redone from scratch xd

Issue #3972, opened 2021-12-21
Start using HostlistsRegistry as a source of vetted lists


ameshkov:
We now have a separate repo with vetted hosts blocklists: https://github.com/AdguardTeam/HostlistsRegistry
We should sync AdGuard Home's code with that repo: https://adguardteam.github.io/HostlistsRegistry/assets/filters.json

`filters.json` may need to be slightly transformed to the format that we use in AdGuard Home.

Additionally, we should update README and link the new repo (instead of the old file with meta).

ainar-g:
This has been partially done in 2a85d7d.  The next step is switching the blocked services.

PR #5108, opened 2022-11-06
Add `amazon.com.be` to the filter list.

fernvenue:
For https://github.com/AdguardTeam/AdGuardHome/issues/5090.


ainar-g:
Thank you for the contribution, but please see the comment on top of `servicelist.go` as well as #3972.  You can open pretty much the same PR in the [Hostlists Registry repo][1] and add the rule to `assets/services.json`, and it should land in AdGuard Home with the next update.

[1]: https://github.com/AdguardTeam/HostlistsRegistry

PR #5142, opened 2022-11-13
fix: invaild List-KR DNS

green1052:


ainar-g:
Hello and apologies for the long wait.  It seems like the issue with the URL of the list has already been fixed in v0.107.20?

PR #5295, opened 2022-12-26
Update servicelist.go

rootmelo92118:
TLD `onion` is only used in tor. And tor DNS won't be regulated by AdGuard Home. It doesn't make sense to add this kind of domain to this file. By the way, the domain `one.one` is not belonging to Cloudflare, but the `one.one.one` is.

https://bgp.he.net/dns/one.one
https://bgp.he.net/dns/one.one.one


EugeneOne1:
See the [comment](https://github.com/AdguardTeam/AdGuardHome/pull/5305#issuecomment-1365945448).

PR #5296, opened 2022-12-26
Update servicelist.go

rootmelo92118:
resource : https://github.com/v2fly/domain-list-community/blob/master/data/bilibili


EugeneOne1:
See the [comment](https://github.com/AdguardTeam/AdGuardHome/pull/5305#issuecomment-1365945448).

PR #5392, opened 2023-01-22
Update index.css

PVasileff:
Clear view for dns rewrites in query log - dark mode


Sil3ntVip3r (approved):
Verified exact head `05931a927a4e40ef9a538fc558d8852c6542f5f2`
against current `master`. The legacy query-log CSS still uses this variable,
and a three-way application is clean.

I also checked the relevant foreground/background contrast using the WCAG
relative-luminance formula. The current blue is 3.64:1 against the log text
color and 4.04:1 against white; the proposed color is 14.32:1 and 15.89:1.
This is a focused readability improvement with no behavior or API impact.

PR #5393, opened 2023-01-23
Added Minecraft (Bedrock) endpoints to Minecraft service block list

JakeShirley:
Minecraft dev here! Thought I would add a few more URLs to the Minecraft service list.  

This PR adds two URLs to the Minecraft block list:
- The (production) Minecraft PlayFab endpoint which handles marketplace, featured 3rd party server lists, and any sort of "store" stuff
- The general, Minecraft authored services which handle things like Realms

Minecraft will still launch and function but its backing services won't connect.

### DIFF
--- a/internal/filtering/servicelist.go
+++ b/internal/filtering/servicelist.go
@@ -1401,6 +1401,8 @@ var blockedServices = []blockedService{{
 		"||minecraft.net^",
 		"||minecraftservices.com^",
 		"||mojang.com^",
+		"||20ca2.playfabapi.com^", // Bedrock PlayFab Services (Login, Marketplace, etc.)
+		"||minecraft-services.net^", // Bedrock Services (Realms)
 	},
 }, {
 	ID:      "netflix",


Sil3ntVip3r:
I checked this against current `master` and the current blocked-service
source of truth. `internal/filtering/servicelist.go` is generated by
`scripts/blocked-services/main.go` from HostlistsRegistry, so a direct edit
here would be overwritten. I ran the current generator in a disposable
worktree: the registry already supplies `||minecraft-services.net^`, but it
does not currently supply `||20ca2.playfabapi.com^`.

The remaining PlayFab rule therefore needs to be proposed and validated in
HostlistsRegistry, followed by the normal AdGuard Home regeneration. The
endpoint evidence in this PR is still useful, but the implementation path has
moved upstream.

Sil3ntVip3r:
I completed the upstream validation for the remaining PlayFab rule.

The exact HostlistsRegistry change is a single source-line addition to
`services/minecraft.yml`:

```diff
   - '||minecraft-services.net^'
+  - '||20ca2.playfabapi.com^'
   - '||cubecraft.net^'
```

Evidence that this is still current rather than a stale 2023 endpoint:

- `20ca2.playfabapi.com` currently resolves over both IPv4 and IPv6 through
  Azure Front Door and accepts HTTPS.
- Current PrismarineJS authentication code still names
  `https://20ca2.playfabapi.com/Client/LoginWithXbox` as Minecraft's PlayFab
  login endpoint:
  https://github.com/PrismarineJS/prismarine-auth/blob/b795199dc5fa26059655bb1bc91c7f7f2733b232/src/common/Constants.js#L31

I tested the patch on HostlistsRegistry `main` at
`0c02e722a69c1c1fa9a6434d4f2c71c6269a0f9e`:

```text
yarn install --frozen-lockfile  PASS
yarn lint                     PASS
yarn compose                  PASS
git diff --check              PASS
```

Composition adds the rule exactly once to Minecraft in
`assets/services.json`.  The compose run also refreshed unrelated dynamic
filter assets from their network sources; those are not part of this service
change and should not be included in a focused patch.

So the remaining implementation path is now concrete: apply the one-line
HostlistsRegistry source change, let that repository regenerate
`assets/services.json`, then run AdGuard Home's blocked-services generator.
The direct `internal/filtering/servicelist.go` edit in this PR should not be
merged because that file is generated.

PR #5634, opened 2023-03-24
Update ru.json

AuthFailed:
Correcting the words form


ainar-g:
Thanks for the contribution, but please see [this section in the README](https://github.com/AdguardTeam/AdGuardHome#translate).  We use CrowdIn to update the translations, and we do it before a beta or a release build.  If you want to add corrections, please do that on CrowdIn.

PR #5692, opened 2023-04-05
docker: Remove CAP_NET_BIND_SERVICE capability

xduugu:
The capability was added in commit 58868b7, but since then, the USER instruction was removed from the Dockerfile, so that Adguard Home runs as root by default.

Even if it is run as non-root user, the capability is not required in docker versions >= 20.10.0 (released on Dec 9, 2020), because of commit 888da28[1], which sets `net.ipv4.ip_unprivileged_port_start` to `0` by default. Users running older docker versions or using the `host` network mode can still configure this sysctl manually.

[1] https://github.com/moby/moby/commit/888da28d42f7d0f9fa250dd8a75d51c2a6cf3098

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,7 +21,7 @@ LABEL\
 	org.opencontainers.image.version=$VERSION
 
 # Update certificates.
-RUN apk --no-cache add ca-certificates libcap tzdata && \
+RUN apk --no-cache add ca-certificates tzdata && \
 	mkdir -p /opt/adguardhome/conf /opt/adguardhome/work && \
 	chown -R nobody: /opt/adguardhome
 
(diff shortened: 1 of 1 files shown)


ferferga (approved):

Sil3ntVip3r (changes requested):
I rechecked exact head `9ef9ccaf` against current master. Removing the file
capability is a compatibility break, not just image cleanup: a non-root
container using host networking, or any namespace where
`net.ipv4.ip_unprivileged_port_start` remains nonzero, loses permission to
bind 53/80/443 and can fail startup. The PR body acknowledges those users
would need a manual sysctl, but there is no support-policy or documentation
change and no integration coverage. Please retain `libcap`/`setcap`, or
explicitly define the new runtime/network requirements and add non-root
bridge and host-network low-port tests. The exact diff is clean; container
validation was unavailable locally, and this head has no CI.

Issue #3701, opened 2021-10-03
Support strict order for upstreams (fallback DNS servers)


timkgh:
I use 2 upstreams from different providers, Cloudflare and Quad9. Cloudflare is fast from my location, Quad9 is about 2x slower. So I'd like to use Cloudflare all the time, unless it's unavailable (which happened before). When it's not available I'd like AGH to switch to the second provider (Quad9 in this case), then periodically check the first provider and switch back to it when it is available again.

For now I can use parallel queries but it seems wasteful and also don't want both providers to have a log of my queries all the time.

This is somewhat related to #3672, if the load balancing feature worked correctly it would be close to equivalent to strict order, unless someone wanted the slower upstream to be the main one.

agneevX:
This is a good feature request... that way primary and backup servers can be used.

timkgh:
This seems very similar to the [dnsproxy](https://github.com/AdguardTeam/dnsproxy) `--upstream` and `--fallback` options. AGH uses `dnsproxy`, is it just a matter of exposing the config options in AGH?

ismay:
> This is a good feature request... that way primary and backup servers can be used.

I agree. This would also allow for upstream dns resolution to prefer dns over a vpn tunnel if the tunnel is up, where it only falls back to a public dns server if the tunnel is down. So say:

```
10.1.1.1
8.8.8.8
```

If 10.1.1.1 is reachable, only that gets used. Otherwise 8.8.8.8 will be used as a fallback.

timkgh:
@ameshkov How difficult would it be to expose the dnsproxy feature through AGH? Thanks!

ameshkov:
Not too complicated, UI changes is what slows this down.

komic:
Hi there, is there any update regarding the implementation of this option? 🙏

ainar-g:
@komic, it's still unplanned, as you can see by the absence of a set milestone.  Please use the :+1: reaction to show your support for the feature.

agneevX:
Hello @ainar-g and team, today I faced an issue where this feature request would've come in really handy and potentially mitigated the issue.

Google DNS was returning SERFAIL responses for a specific domain while other servers were responding properly.

I'd like to therefore ask that this feature be prioritized.

![Screenshot (183)](https://user-images.githubusercontent.com/19761269/157240148-3df73217-c52f-4da5-b4ba-51036b96f4b0.png)

[...]

PR #5731, opened 2023-04-14
feature: support fallback upstream

rrfeng:
simply use the dnsproxy option --fallback

#3701


rrfeng:
@ainar-g would you like to take a look at this tiny patch? If this is ok, I will spend some time to update the UI.

ainar-g:
Apologies for the very belated response and thank you for the contribution.  We have implemented fallback upstreams in a different manner in #3701.

PR #5750, opened 2023-04-19
add /control/dns_config to expectsLargerRequests list

xbugio:
上游DNS服务器配置那里，有需求会配置一大堆的域名来通过另一个域名服务器来解析。
本来这个工作是交给DNSMASQ做的，后来看AdguardHome有这个功能，就想把DNSMASQ去掉。
结果添加了一堆后报错了。
希望这个接口也允许可以POST略大的数据

### DIFF
--- a/internal/home/middlewares.go
+++ b/internal/home/middlewares.go
@@ -45,7 +45,8 @@ func expectsLargerRequests(r *http.Request) (ok bool) {
 
 	p := r.URL.Path
 	return p == "/control/access/set" ||
-		p == "/control/filtering/set_rules"
+		p == "/control/filtering/set_rules" ||
+		p == "/control/dns_config"
 }
 
 // limitRequestBody wraps underlying handler h, making it's request's body Read


ainar-g:
Good day.  The current recommended way to handle large upstream lists is by using `dns.upstream_dns_file` in `AdGuardHome.yaml`.  Is there a need for this change besides the upstream list?

xbugio:
> Good day. The current recommended way to handle large upstream lists is by using `dns.upstream_dns_file` in `AdGuardHome.yaml`. Is there a need for this change besides the upstream list?

I didn't know that i could just use a dns.upstream_dns_fie.
May i know if i need to restart the process after i modify the upstream file?
Early I tried to modify the AdGuardHome.yaml and restart the service every time I needed to set an upstream server for a new domain.
After sevral times, I decided to change the code so i can simply do that on the admin page.
It's so convenient to make me think maybe I should make a pull request.

ainar-g:
>   May i know if i need to restart the process after i modify the upstream file?

You should stop AdGuard home whenever you change the file manually and start it once you're done.  We document that, see for example our [FAQ page about enabling verbose logs](https://github.com/AdguardTeam/AdGuardHome/wiki/FAQ#verboselog).  You can see the full reference of the configuration file on [its Wiki page](https://github.com/AdguardTeam/AdGuardHome/wiki/Configuration) as well.

Since there seems to be no need for this in the API, I'll close this PR, if you do not mind.

PR #5763, opened 2023-04-22
Feature: Blocking Mode - Blocking Page

reqed:
You could maybe add a custom blocking-page as redirect like on this site http://176.103.130.133/


ainar-g:
This doesn't seem like an actual PR.  Please use Issues to report bugs and send feature requests (provided it's not a dup).  As for the custom blocking pages, see the [FAQ entry](https://github.com/AdguardTeam/AdGuardHome/wiki/FAQ#customblock).

PR #5841, opened 2023-05-22
dhcpd: fix malformed link-layer address packet

### LINKED ISSUE #5840: Malformed Router Advertisment packet in DHCPv6 (with patch)
### Prerequisites

### Operating system type

Linux, Other (please mention the version in the description)

### CPU architecture

AMD64

### Installation

### Setup

On one machine

### AdGuard Home version

v0.107.29

## What did you do?
I am using AdGuard for a couple DNS rewrites and as a DHCP server in order to provide AdGuard as the DNS server (my router has no options to override the DNS server). For IPv4, this worked perfectly, and my server was set as the DNSv4 nameserver. For IPv6, the router was still used as the DNS server. I used wireshark to analyze the packets and discovered that the packets are malformed due to the MAC address being padded by zeros (See images)

## Original packet:
Router Advertisement packet from Adguard. Note that the MAC address is suffixed by 2 empty bytes before the next ICMPv6 option (which is the DNS option). (MAC address has been replaced with XX)
```
0070   xx xx xx xx xx xx 00 00 19 03 00 00 00 00 0e 10
0080   fe 80 00 00 00 00 00 00 cc 88 e2 8d 78 03 44 3f
```
The packet after the Source Link-layer Address packet is then malformed since it begins with zeros. 
image

## Edited packet:

If I edit that packet to remove those zeros, the packet is detected properly:
```
0070   xx xx xx xx xx xx 19 03 00 00 00 00 0e 10 fe 80
0080   00 00 00 00 00 00 cc 88 e2 8d 78 03 44 3f
```
image

## Additional information
I believe that the cause of this is [`hwAddrToLinkLayerAddr`](https://github.com/AdguardTeam/AdGuardHome/blob/master/internal/dhcpd/routeradv.go#LL43C8-L44C1) which converts MAC addresses to 8 bytes. While the RFC says that the length of the packet must be 8 bytes, I believe that is: Option Type (1 byte) + Length (1 byte) + address (6 bytes) = 8 bytes. Based on my inspection of Router Advertisement packets of my router, the address is 6 bytes. The [RFC](https://datatracker.ietf.org/doc/html/rfc4861#section-4.6.1) linked at the start of the function matches this interpretation:
```
Length         The length of the option (including the type and
                     length fields) in units of 8 octets.
```

I have written a pull request that fixes this issue: #5841

yamanq:
closes #5840

See issue for details on bug and approach for this fix.


snachx:
I think we should also modify the Length field of the source link-layer address option based on the length of the return result of hwAddrToLinkLayerAddr, instead of always setting it to 1. https://github.com/AdguardTeam/AdGuardHome/blob/ae7bd79a6c2dba0061db4d5d46b9069e4bd3f22c/internal/dhcpd/routeradv.go#LL179C11-L179C11

snachx:
You are right, AdGuard Home's DHCP server does indeed have several other issues. For example, the parameters for Router Advertisement (RA) messages are not configurable, and the most severe problem is that the default interval for sending RA messages is 1 second, which is too short. The recommended maximum interval according to the RFC is 600 seconds, and the minimum interval is 200 seconds. Additionally, it lacks the functionality to respond in real-time to received Router Solicitation (RS) messages, among other things. There is still much work to be done to use AdGuard Home as a DHCPv6 server.

Sil3ntVip3r:
This is now implemented on current master by commit `74640fb0`: the source
link-layer option length includes the type/length bytes, rounds to an
eight-octet unit, and pads after the actual hardware address. The current
regression decodes the packet with gopacket and asserts the six-byte MAC; it
passed 20/20 under `-race` in my fresh check. Exact head `e9e353dd` now
conflicts with that implementation, so this PR and linked #5840 appear
superseded and can be closed.

PR #5959, opened 2023-07-03
fix error days

Jiraiya8:
String "number_of_dns_query_days": "The number of DNS queries processed for the last {{count}} day",

Mistakenly displayed in milliseconds。


ainar-g:
Thanks for the contribution!

Issue #5720, opened 2023-04-13
Possibility for wildcards in "ignored domains"


RealKev79:
### Prerequisites

- [X] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [X] I have searched other issues and found no duplicates

- [X] I want to request a feature or enhancement and not ask a question

### Description

Within the "ignored domains" in logs-configuration and statistics-configuration I want to add several domains with wildcards because I do not want to "see" every single subdomain or similar from a specific domain. For example I want to add following domain:
*.blabla.org
and this should trigger on "www.blabla.org" and also on "blablablabla.blabla.org"

I have a working laptop and when I am working at home I did not want to see the whole specific company domains.

At the moment I added these into the "disallowed domains" in the DNS-Settings. 

Regards
Kevin

davralin:
In the same situation myself, I don't want to see my monitoring setup's continous query of my internal hosts - so I want to exclude my entire local DNS-name from the query log.

@RealKev79 
How does it work with having those DNS-records in `Disallowed domains` ? The way I understand the help-text, those domains wont be resolved at all?

RealKev79:
@davralin 
Yes the domains entered in "disallowed domains" are completly ignored and are not resolved. Because of that "issue" it will be an effort if the domains can entered in the  "ignored domains" within the "general settings" / "logs configuration" as I mentioned above.

ainar-g:
AdBlock syntax is going to be supported in v0.107.37.

whyisthisbroken:
So i can now put *in-addr.arpa in the Ignored Field ?
because I still see the entries.
Will there possibly be a UI added? Maybe in the same button where you can also block?

FFFcat:
> So i can now put *in-addr.arpa in the Ignored Field ? because I still see the entries. Will there possibly be a UI added? Maybe in the same button where you can also block?

AdBlock syntax such as `||in-addr.arpa`, it's just hidden

jeverley:
Is it expected that we should be able to use the AdBlock $client syntax in the rules?

||168.192.in-addr.arpa
||example.com^$client=192.168.1.6

With the above I successfully exclude the arpa queries, but I can still see entries for _example.com_ from the specified client IP.

gpz1100:
[...]

PR #6097, opened 2023-08-10
Added wildcards support for statistics and querylog ignore

sky96111:
Close: https://github.com/AdguardTeam/AdGuardHome/issues/5720

support prefix domain wildcard for statistics and querylog ignore (e.g. ``*.arpa``)


ainar-g:
Thanks for the contribution, but we'll probably use the same filtering engines we use for filtering and host blocking.  They support wildcards and regexps.

PR #6284, opened 2023-10-06
ci & build: build, tag and push for ghcr.io

### LINKED ISSUE #2482: Add ghcr.io as an option to pull the image
docker hub has pull limits in place. ghcr does not currently. should be a couple extra tasks in CI job to login via token and push to ghcr.io

### LINKED ISSUE #5975: Unable to Find ghcr.io Entry for Geoipupdate
### Prerequisites

### The problem

I see that there is a docker file for this project, but I cannot find the GitHub Container Registry for this project. I do see it on Docker Hub, but I no longer use that as my default due to licensing issues.

### Proposed solution

Use GitHub actions to build a Docker container and then add it to GHCR.

There is no guarantee that Docker won't take down the free container registry at some point, so this essentially is the backup option.

ferferga:
Fixes #5975, fixes #2482 (which was closed for some reason)

Ensure tags for ghcr.io are also created when building Docker images. These should be harmless in local environments as well.

The only requirement for this before merge is to have the ``GHCR_TOKEN`` added so GitHub Actions can login into the registry.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,6 +100,12 @@
       'uses': 'docker/setup-qemu-action@v1'
     - 'name': 'Set up Docker Buildx'
       'uses': 'docker/setup-buildx-action@v1'
+    - 'name': Login to GitHub Container Registry 🔑
+      'uses': 'docker/login-action@v3.0.0'
+      'with':
+        'registry': ghcr.io
+        'username': ${{ github.repository_owner }}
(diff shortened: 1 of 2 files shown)


ainar-g:
Thank you for the contribution, but we have no plans to support this.  See https://github.com/AdguardTeam/AdGuardHome/issues/5975#issuecomment-1785007264.

PR #6398, opened 2023-11-08
Clarify the README comparison with pi-hole.

TimTheBig:
Just a small clarification, because pi-hole can block adult domains with external block lists.


ainar-g:
Thanks for the contribution!

PR #6410, opened 2023-11-10
feat: Enable dependabot in this repository to make it easier to upgrade dependencies that are being used by this project

wmundev:
This change enables dependabot which will hopefully make it easier to keep dependencies up to date, specifically the following dependencies
- go dependencies
- client npm packages dependencies
- github actions dependencies

Hope it helps!

Here is a couple of screenshots to show you what dependabot will raise as PRs

image
image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,7 +24,7 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
   # Upgrade for npm packages, minor and patch versions only
-  - package-ecosystem: "gomod"
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


ainar-g:
Thanks for the contribution, but we have our own process for updating dependencies.  It is often not safe to update them automatically without some kind of testing.

wmundev:
no worries @ainar-g 
if you do want to reconsider however, enabling just the `- package-ecosystem: "github-actions"` and removing the other configuration with go andd npm is very safe as it only updates github related actions

e.g. `actions/checkout@v2` will be updated to `actions/checkout@v4` for example in your github actions actions

PR #6451, opened 2023-11-23
home/config: improve error message when validating bind hosts

ishanjain28:
Hey, This PR improves error message when the program sees an invalid address. 

The `BindHosts` field in `Config` struct in the program is of type `netip.Addr`. `yaml` reads the (unquoted)value `2a0a:6040:4050::` incorrectly and it's set to 'Invalid IP' by `netip`.  

```
dns:
  bind_hosts:
    - 10.1.1.3
    - 2a0a:6040:4050::
  port: 53
  anonymize_client_ip: false
  ratelimit: 200
  ratelimit_subnet_len_ipv4: 24
  ratelimit_subnet_len_ipv6: 56
  ratelimit_whitelist: []
  refuse_any: true
```

The error message has been updated from 

```
2023/11/23 13:49:07.012617 [error] parsing configuration file: dns.bind_hosts at index 3 is not a valid ip address
```
to 
```
2023/11/23 13:48:35.042301 [error] parsing configuration file: dns.bind_host at index 3 with value 'invalid IP' is not a valid ip address. try quoting the value
```

There is no associated issue or wiki discussion since this is a relatively small change.

### DIFF
--- a/internal/home/config.go
+++ b/internal/home/config.go
@@ -463,7 +463,7 @@ func validateBindHosts(conf *configuration) (err error) {
 
 	for i, addr := range conf.DNS.BindHosts {
 		if !addr.IsValid() {
-			return fmt.Errorf("dns.bind_hosts at index %d is not a valid ip address", i)
+			return fmt.Errorf("dns.bind_hosts at index %d with value '%s' is not a valid ip address. try quoting the value", i, addr)
 		}
 	}
 


ainar-g:
Thanks for the contribution, but the problem here is deeper, it seems.  See go-yaml/yaml#1006.  Considering that that repo seems to be half-dead, we'll probably need to make a wrapper type that implements `YAMLUnmarshaler` for `netip.Addr` and actually validates the value to get a more obvious message on the decoding stage.

Please file an issue about invalid addresses not being properly handled, so that this doesn't get lost.

ainar-g (changes requested):
See comment above.

Sil3ntVip3r:
The underlying problem has since been fixed on `master`, so this PR can now be
closed as superseded.

Commit `9be4ca90ef632eb684aafe6beeced692e8860018` added the deeper handling
requested in the maintainer review.  Current `validateBindHosts` keeps the
generic invalid-address error, but when an invalid bind host is found it
re-parses the original YAML and detects the `{key: null}` shape produced by an
unquoted IPv6 address ending in a colon.  It then logs the targeted warning:

```text
quote addresses that end with a colon in 'dns.bind_hosts'
```

That preserves the raw-input evidence which is already lost by the time the
value becomes an invalid `netip.Addr`.  The original one-line change here
cannot distinguish that YAML case from unrelated invalid addresses and should
not be merged.

Current implementation:
https://github.com/AdguardTeam/AdGuardHome/blob/369d47616342c8be82262bf798f2fed2f64b697f/internal/home/config.go#L666-L789

PR #6453, opened 2023-11-23
update tracker list: porn sites

duhow:
Despite there is no category for "porn websites", it's more suitable to tag the websites with "pornversiting" category.


ainar-g:
Please file an issue in <https://github.com/AdguardTeam/companiesdb> instead, as this is where AGH gets the data.

PR #6564, opened 2023-12-20
correct schema type for client upstreams_cache_size

bakito:
The openapi schema has a wrong data type for the Client field upstreams_cache_size.

Currently in the schema it is defined as boolean, but in the struct an integer.


ainar-g:
Thank you for your contribution!  We'll review it shortly.

PR #6703, opened 2024-02-03
Mention AdGuardHome sync in usages

bakito:
Mention 'AdGuardHome sync' in Applications that use AdGuardHome.


ainar-g:
Thanks for the contribution!  We'll merge it soon.

PR #6748, opened 2024-02-18
Fix for privilege escalation vulnerability

go-compile:
### Description
Patch for a privilege escalation vulnerability in the service component of AdGuardHome. 

**Please note that this pull request does not come with a issue number because it was privately disclosed via email to @ameshkov and @ainar-g; due to the sensitive nature of security disclosure.**

### Patch Details
This patch sets the correct executable permissions for AdGuardHome when running the `AdGuardHome -s install` command. This includes setting the owner as root:root and permissions to 0755 (blocks write access to all users except root). Furthermore, the binary is moved to `/usr/bin/` to add it to the global path and vitally place it in a non-root read-only directory, otherwise the previous steps would not be enough to protect against this vulnerability.

### Patch Knock On Effects (all accounted for)
Additional changes include, disabling the installation of AdGuardHome as a service on Windows due the vulnerability being unable to be patched (Go lacks the proper API bindings to set file ownership and permissions on Windows). It is highly recommended a msi installer is created to install the service, rather than allow a user to try install it manually without any guidance. Moreover, even with guidance the process is complex and time consuming. Do note this can easily be changed in `internal/home/service.go` on line `698`.

Other alterations. There is a issue where AdGuardHome will write it's data to `/usr/bin/` if the executable is within that path. This is an inappropriate directory for data and thus should not be used. Instead if the program is within `/usr/bin/` AdGuardHome now assumes a new directory `/var/lib/AdGuardHome` (auto created recursively if it does not exist). Note, this does not interfere with the desired behavior of allowing AdGuardHome to use the current executable dir as it's workDir, it only redirects the workDir if the binary is within `user/bin` (essentially if AdGuardHome is running as a service it will store data in the new location, unless overwritten with the existing workdir CLI option).

### Testing
This PR has been tested on Ubuntu wsl.


internal/home/service.go
```diff
@@ -681,3 +687,55 @@ rc_bg=YES
...
 `
+
+// secureBinary is used before service.Install(). This function protects AdGuardHome from
+// privilege escalation vulnerabilities caused by writable files
+func secureBinary() error {
+	switch runtime.GOOS {
+	case "windows":
+		// TODO: support windows service support securely
+		// Set file owner to admin/system and public permissions read-only
+		return errors.Error("you currently cannot install adguardhome as a service on window")
```
ameshkov, line 698: This would definitely be too problematic to existing users and to Windows users in general.

I suggest instead printing a warning that explains the situation.

internal/home/service.go
```diff
@@ -681,3 +687,55 @@ rc_bg=YES
...
+
+	// Set permissions to root(read,write,exec), group(read,exec), public(read)
+	// This combined with changing the owner make the file undeletable without root privlages
+	// UNLESS THE PARENT FOLDER IS WRITABLE!
+	if err := os.Chmod(binary, 0755); err != nil {
+		return err
+	}
+
+	// Move binary to the PATH in a folder which is read-only to non root users
+	if err := os.Rename(binary, "/usr/bin/AdGuardHome"); err != nil {
```
ameshkov, line 727: Hard-coding the path here is problematic

ameshkov (changes requested):
Thank you for the submission!

Added some comments there. However, let's wait for @ainar-g's input before changing anything, he may add some corrections.

internal/home/service.go
go-compile: Would you prefer it be a CLI option called "install-dir" defined in `internal\home\options.go`  `var cmdLineOpts = []cmdLineOpt{{` with a default value of "/usr/bin".  Then the dir variable will have the executable name appended set as  "AdGuardHome" (which can be defined as a constant).

 Alternatively I can make it a constant defined at the top of the file.

internal/home/home.go
```diff
@@ -763,6 +763,15 @@ func initWorkingDir(opts options) (err error) {
 	if opts.workDir != "" {
 		// If there is a custom config file, use it's directory as our working dir
 		Context.workDir = opts.workDir
+	} else if !execDirAvaliable() {
+		// If running as a service and from /usr/bin/ use /var/lib for working dir instead of
+		// /usr/bin/data
+		Context.workDir = "/var/lib/AdGuardHome"
+
+		// Create dir if it does not already exist
+		if err := os.MkdirAll(Context.workDir, 0755); err != nil {
```
ainar-g, line 772: Here and in a few other places, prefer to not shadow the `err`.  See `make go-lint`.

internal/home/service.go
```diff
@@ -304,6 +305,11 @@ func handleServiceStatusCommand(s service.Service) {
 
 // handleServiceStatusCommand handles service "install" command
 func handleServiceInstallCommand(s service.Service) {
+	// Set the binary's permissions and move to /usr/bin (if on linux)
```
ainar-g, line 308: Is this also true for BSDs?

internal/home/service.go
```diff
@@ -681,3 +687,64 @@ rc_bg=YES
 
 rc_cmd $1
 `
+
+// secureBinary is used before service.Install(). This function protects AdGuardHome from
+// privilege escalation vulnerabilities caused by writable files
+func secureBinary() error {
+	switch runtime.GOOS {
```
ainar-g, line 694: In new code, we prefer to use build tags for OS-specific pieces of functionality.  See package `aghos`.  In fact, this function should probably be there.

internal/home/service.go
ainar-g: The warning should probably include the link to the new [DNS KB article](https://adguard-dns.io/kb/adguard-home/running-securely/).

internal/home/service.go
```diff
@@ -681,3 +687,64 @@ rc_bg=YES
...
+	case "windows":
+		// TODO: support windows service support securely
+		// Set file owner to admin/system and public permissions read-only
+		return errors.Error("you currently cannot install adguardhome as a service on window")
+	default:
+		return secureBinaryUnix()
+	}
+}
+
+func secureBinaryUnix() error {
```
ainar-g, line 704: Please document the function (or the exported API once moved to `aghos`).

See https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Go.md#li-4e61a581.

internal/home/service.go
```diff
@@ -681,3 +687,64 @@ rc_bg=YES
...
+	// Installalation can only be completed with root privileges, so check and handle if not
+	if os.Getuid() != 0 {
+		return errors.Error("permission denied. Root privileges required")
+	}
+
+	// Get current file path
+	binary, err := os.Executable()
+	if err != nil {
+		return err
```
ainar-g, line 713: Please add context to either the function, per https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Go.md#li-651fcd50.

internal/home/service.go
```diff
@@ -681,3 +687,64 @@ rc_bg=YES
...
+// If AdGuardHome is running as a service, it should not use the binary's location as a
+// workDir, thus this function will return false.
+func execDirAvaliable() bool {
+	binary, err := os.Executable()
+	if err != nil {
+		panic(err)
+	}
+
+	// If installed in /usr/bin do not use /usr/bin/data to store files
+	return filepath.Dir(binary) != "/usr/bin"
```
ainar-g, line 749: If the code needs to treat `/usr/bin` specially, it should be a constant in `aghos` with documentation on why it's treated specially.

internal/home/service.go
```diff
@@ -681,3 +687,64 @@ rc_bg=YES
...
+}
+
+// execDirAvaliable returns true if the executable's current folder is avaliable to be
+// used as a workDir.
+// If AdGuardHome is running as a service, it should not use the binary's location as a
+// workDir, thus this function will return false.
+func execDirAvaliable() bool {
+	binary, err := os.Executable()
+	if err != nil {
+		panic(err)
```
ainar-g, line 745: Please no panic here, as `os.Executable` routinely fails on macOS unless the binary is called from a particular location.  See `scripts/install.sh`.

ainar-g (changes requested):
Additional notes:

* It is not clear how this change affects the BSDs, and in particular macOS, which really wants applications to be installed in `/Applications`, for example.

* It seems like it's breaking Docker?  There are assumptions about AdGuard Home being installed to `/opt/` there.

* The install script should probably be changes as well to fully implement the change for new installations as well.

go-compile:
> It is not clear how this change affects the BSDs, and in particular macOS, which really wants applications to be installed in /Applications, for example.

Regarding macOS, I am very unfamiliar with the system. But if /Applications is owned by root and denies writes by non-root users, I don't see we can't just define `const serviceInstallDir = "/usr/bin/AdGuardHome"` to `const serviceInstallDir = "/Applications/AdGuardHome"`. Then also add`/Applications` to `protectedDirectories`. These variables could be moved to os_linux.go, os_darwin within the `aghos` package.

I will test AdGuardHome on a OpenBSD vm at some point and verify it works; currently it uses the same patch as the one for linux.

> It seems like it's breaking Docker? There are assumptions about AdGuard Home being installed to /opt/ there.

I don't see why it would break docker as Docker doesn't install the service and the checks added only run when `-s install` is ran?

> The install script should probably be changes as well to fully implement the change for new installations as well.

I am not as familiar with the script, and it is very large. But would redefining $agh_dir to equal the new location, obtained by the command `whereis` or hard coded work?

```bash
...

# Function install_service tries to install AGH as service.
install_service() {
	# Installing the service as root is required at least on FreeBSD.
	use_sudo='0'
	if [ "$os" = 'freebsd' ]
	then
		use_sudo='1'
	fi

	if ( cd "$agh_dir" && maybe_sudo ./AdGuardHome -s install )
	then
                $agh_dir = "/usr/bin"
		return 0
	fi

	log "installation failed, removing $agh_dir"

...
```

internal/home/service.go
go-compile: Yes

go-compile:
Do we have any updates for this issue?

Side note the CVE number is CVE-2024-36586 (reserved for now).

Issue #6634, opened 2024-01-11
Settings page stays blank if stats/queries purged recently


xRuffKez:
### Prerequisites

- [X] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) and found no answer

- [X] I have searched other issues and found no duplicates

- [X] I want to report a bug and not [ask a question or ask for help](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a)

- [X] I have set up AdGuard Home correctly and [configured clients to use it](https://github.com/AdguardTeam/AdGuardHome/wiki/Clients). (Use the [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) for help with installing and configuring clients.)

### Platform (OS and CPU architecture)

Linux, AMD64 (aka x86_64)

### Installation

GitHub releases or script from README

### Setup

Other (please mention in the description)

### AdGuard Home version

Multiple public servers with load balancer

### Action

```sh
nslookup -debug -type=a 'www.example.com' '$YOUR_AGH_ADDRESS'
```

### Expected result

Purging query logs/stats will load settings page correctly.

### Actual result

Purging query logs/stats will lead to settings page to stay blank and a restart is required.

### Additional information and/or screenshots

Only happens on huge query entries beeing purged in settings of frontend.

Mizzick:
@xRuffKez 
Please have a look, we have implemented the requested changes.
The new build version `v0.108.0-a.853+7f24fc7c` has been just published to the edge channel.

PR #6770, opened 2024-02-26
Fix blank settings page when access clients first

hoang-rio:
**Step to reproduce bug:**
1. Open dashboard
2. Open client setting
3. Open settings

**Actual:**
Blank page by queryLogs.ignored undefined error

**Expect:**
Settings page display normally

It may a fix for #6634


Mizzick:
I reproduced the issue and I confirm it's solved here.

PR #6789, opened 2024-03-01
Add build release for android arm64

JeelsBoobz:


hyprilyes (commented):
can you hepl me

sevouilyes:
tu parle fr

Sil3ntVip3r:
I validated exact head `568f86b4`; its one-line change still merges cleanly
into current master. An Android/arm64 build succeeds with the repository's
CGO-disabled release settings and produces an aarch64 Android ELF using
`/system/bin/linker64`. The release script's generic archive/version logic
would emit `AdGuardHome_android_arm64.tar.gz` and `download_android_arm64`. I
found no code-level blocker. Before treating Android as an officially
supported target, please link the approved support requirements and add an
Android runtime smoke test covering startup, DNS binding, configuration
persistence, and shutdown. Exact diff check passed; no CI is attached.

Issue #4680, opened 2022-06-19
agh_session cookie key conflict for multiple instances


ihotte:
It is recommended to provide command line parameters for modifying the agh_session cookie key, so that multiple instances can be logged in to the same host at the same time.

Issue #1196, opened 2019-11-23
Session cookie issue


Onepamopa:
When you have several websites on the same domain/ip where adguard UI is configured (I have directadmin on port 2222 and adguard on port 10000), logging into adguard is impossible (all the time 302 redirect back to /login.html). If I log out of DirectAdmin - I can login normally in AGH.

Suggest: rename "session" cookie to "agh_session" or add config option so we can name the cookie.

ameshkov:
Makes sense, thank you!

PR #7072, opened 2024-06-12
Support customizing session cookie name.

hsldymq:
What scenarios require custom session cookie names?

When multiple AdGuard Home instances are running on a host, having the same IP or domain name, only the port is different, you cannot log in to the web UI of two instances at the same time, because they use the same session cookie name. . After logging into instance A, logging in to instance B will cause the login status on instance A to be invalid.

#4680 and #1196 also mentioned this problem.

### DIFF
--- a/internal/home/authhttp.go
+++ b/internal/home/authhttp.go
@@ -22,8 +22,11 @@ import (
 // cookieTTL is the time-to-live of the session cookie.
 const cookieTTL = 365 * timeutil.Day
 
+// defaultSessionCookieName is the default name of the session cookie.
+const defaultSessionCookieName = "agh_session"
+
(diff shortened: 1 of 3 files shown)


hsldymq:
Hi @ainar-g , could you please have a look at this PR

ainar-g:
Good day.  As recommended in the [contribution guidelines](https://github.com/AdguardTeam/AdGuardHome/blob/d3cc2dc930d8e03bd72fc374edcbc5fe69f27740/CONTRIBUTING.md), it's better to discuss a change before contributing any code.  Currently, we think that it's the job of the reverse proxy (i.e. Nginx) to handle header mutations for unusual configurations, so we'd rather not add complexity for that.

PR #7290, opened 2024-09-26
snap: set restart-delay to make service more reliable

alexmurray:
This makes the service more reliable to start at boot on slower devices where it takes longer for networking to properly come up. Should help fix issues like #6983

### DIFF
--- a/snap/snap.tmpl.yaml
+++ b/snap/snap.tmpl.yaml
@@ -36,6 +36,7 @@
     - 'network-control'
     'daemon': 'simple'
     'restart-condition': 'always'
+    'restart-delay': '30s'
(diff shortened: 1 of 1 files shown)


Sil3ntVip3r (approved):
Reviewed exact head `33d3703f`. The one-line change still merges cleanly into
current master. `adguard-home` is already a `daemon: simple` with
`restart-condition: always`, and Snapcraft defines `restart-delay` as a
duration for daemon apps, so `30s` provides the intended retry spacing after
early network or bind failures. The template parses as YAML and the exact
diff is clean. I could not run a Snapcraft build locally because Snapcraft is
not installed; this head has no attached CI.

PR #7552, opened 2025-01-06
7421 add brave safe search

### LINKED ISSUE #7421: SafeSearch option for Brave engine
### Prerequisites

### The problem

Hello,

we don't have a safesearch option for Brave engine.

### Proposed solution

Please, add safesearch preset for Brave search engine. We just need to add CNAME record for:
`search.brave.com`
point to:
`safe.search.brave.com`

_No response_

FlorianLang06:
Fixes #7421 

I added the option to enable safe search for [brave search engine](https://search.brave.com/).


Sil3ntVip3r (changes requested):
I applied this exact head (`bd48661a9531fb093731f2d0f3d9e2375574fd21`)
to current master and ran the focused and package race tests.  The backend
rule works in isolation, but two merge blockers remain.

1. **Please complete the public API and current UI integration.**  The patch
   adds the backend config field, rule, and test, but `SafeSearchConfig` in the
   OpenAPI schema and generated `client_v2` model still omit `brave`.  The
   provider map and global settings UI omit it as well, and the per-client
   state/payload path does not preserve it.  Editing an existing per-client
   object with `brave=true` through `client_v2` can therefore drop the field on
   save.  Please add the field to OpenAPI and its changelog, regenerate the
   client, wire it into global and per-client settings, update the deprecated
   client fallback and legacy hint, and add API/UI round-trip coverage.

2. **Please use Brave's documented canonical target.**  The rule and test use
   `safesearch.brave.com`, while Brave's current administrator documentation
   specifies:

   ```text
   search.brave.com IN CNAME forcesafe.search.brave.com
   ```

   Source: https://safe.search.brave.com/help/safesearch

   The old alias currently redirects successfully in a live probe, so I am
   not claiming a present bypass.  It is nevertheless undocumented and may
   disappear; a temporary regression test expecting the documented target
   failed exactly on this mismatch.  AdGuard's HostlistsRegistry safe-search
   asset also carries the old alias, so that asset should be coordinated
   separately rather than copied as authority here.

After those changes, the useful regression matrix is: enabled A/AAAA queries
return `FilteredSafeSearch` with `forcesafe.search.brave.com`; disabled queries
remain unfiltered; GET/PUT and per-client add/edit/save preserve the field;
the deprecated fallback enables it consistently; and other providers remain
unchanged.  The focused Brave test, full safe-search package, and filtering,
safe-search, and home packages all passed under `-race` for the current patch.

FlorianLang06:
I am not that familiar with the codebase so I am not sure if I missed something which you mention in you comment. Please let me know if I still miss something.

Sil3ntVip3r (approved):
Re-reviewed exact head 05e86c4f8abeb84b366687674b9f82e5da880693 against current master. The previous blockers are resolved: the rule uses Brave’s documented search.brave.com -> forcesafe.search.brave.com target, and the setting now reaches the OpenAPI contract, regenerated current UI client, global settings, per-client add/edit/save path, deprecated fallback, and legacy API hint.

Validation completed locally: go test -race -count=1 ./internal/filtering/safesearch ./internal/filtering ./internal/home; make go-check; npm run api:generate with a clean generated diff; and npm run check (619 tests). Approved.

PR #7678, opened 2025-03-08
fix: QueryLogItem.client_proto in openapi schema

bakito:
This add the missing type to QueryLogItem.client_proto in the opanapi schema.


ainar-g:
Thank you for the contribution and apologies for the long wait!  We'll review this shortly.

PR #7738, opened 2025-03-30
Fixes #1709: Per Client Blocking Mode configuration 

### LINKED ISSUE #1709: Client Settings Blocking Mode
dont know if this been asked before,
is this even possible?
blocking mode per client setting?
if so, its a good feature to have

ishanjain28:
This PR closes https://github.com/AdguardTeam/AdGuardHome/issues/1709

This PR adds the following.

1. In the backend side, It adds per client `blocking_mode` configuration. 
2. On the frontend, It adds the UI to configure Blocking Mode for a client. 

Notes: 
* The default value of `blocking_mode`, `blocking_ipv4`, `blocking_ipv6` for a client is derived from the global values. 
* Previously, The blocking mode configuration was looked up in `*Server.DnsContext.BlockingMode` which returns the global values. Now, These values are looked up in the client context.

### DIFF
--- a/client/src/components/Settings/Clients/Form/components/MainSettings.tsx
+++ b/client/src/components/Settings/Clients/Form/components/MainSettings.tsx
@@ -166,7 +166,9 @@ export const MainSettings = ({ processingAdding, processingUpdating, safeSearchS
                         name="blocking_mode"
                         control={control}
                         render={({ field }) => (
-                            <Radio {...field} options={blockingModeOptions} disabled={processingAdding || processingUpdating} />
+                            <Radio {...field}
+                                options={blockingModeOptions}
+                                disabled={processingAdding || processingUpdating} />
                         )}
                     />
                 </div>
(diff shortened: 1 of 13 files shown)


Sil3ntVip3r (changes requested):
I reviewed exact head `d3d0de62`. The feature remains useful, but this
implementation needs a current-master rewrite before it is safe to merge.

Blocking issues:

- The new JSON fields are non-optional. Previously valid update bodies that
omit them decode to empty values and are silently converted to `default`, so
an unrelated legacy update can erase an existing per-client override.
Missing fields should preserve the previous values, with a regression test.
- `Persistent.validateBlockingMode` accepts unknown enum values. Such a
client later reaches response generation, logs an invalid mode, and returns
an unintended empty response. Unknown YAML and HTTP values should be
rejected.
- Non-custom modes serialize zero `netip.Addr` values as the literal
`"invalid IP"`. These should be empty unless custom mode has valid addresses.
- The new fields are absent from the OpenAPI `Client` and
`ClientFindSubEntry` schemas, so generated clients and API documentation
cannot round-trip the feature.

Current master also has the newer `client_v2` and substantially changed
client/API internals.

Local exact-head verification:

- `gofmt -d`: clean.
- `go test -race -count=1 ./internal/client ./internal/dnsforward
./internal/filtering ./internal/home`: all pass.
- `git diff --check`: pass.

The existing tests do not cover the compatibility, invalid-mode,
serialization, or OpenAPI cases above.

Sil3ntVip3r:
I converted the exact-head review blockers into a current-master proof on
`369d47616342c8be82262bf798f2fed2f64b697f` (fresh fetch: `0 0`
divergence from `origin/master`).  The prototype is intentionally
unpublished.

The current-master rewrite now:

- models an empty persisted mode as dynamic inheritance and the API literal
  `global` as an explicit reset;
- makes `blocking_mode` optional on updates so older update bodies preserve
  an existing override, while omitted add fields default to global;
- rejects unknown modes and invalid or wrong-family custom IPv4/IPv6 values at
  the HTTP, YAML, and persistent-client validation boundaries;
- never serializes zero `netip.Addr` values as `invalid IP`;
- seeds each request with the global response settings, then applies a
  per-client override independently of the existing
  `use_global_settings` protection-settings switch;
- generates filtered replies from the request's effective settings rather
  than reading the global mode again;
- adds the optional fields to the OpenAPI `Client` and
  `ClientFindSubEntry` schemas and regenerates one shared
  `ClientBlockingMode` model; and
- implements the feature in both the legacy React UI and active SolidJS UI,
  strips stale custom addresses from non-custom payloads, and validates
  custom addresses locally.

Regression coverage includes legacy update preservation, explicit global
reset, unknown and wrong-family input, JSON/YAML round trips, absence of
`invalid IP`, request-local effective settings, global/custom DNS response
generation, old API payload defaults, payload stripping/serialization, store
validation, and SolidJS rendering/editability.

Final local verification passed:

```text
go test -race -count=1 ./internal/client ./internal/filtering \
  ./internal/dnsforward ./internal/home
go test -race -count=3 ./internal/client ./internal/filtering \
  ./internal/dnsforward ./internal/home
make go-check
make md-lint

cd client
npm run lint
npm run typecheck
npm run test          # 4 files, 35 tests
npm run build-prod

cd ../client_v2
npm run lint
npm run typecheck
NODE_OPTIONS=--localstorage-file=<temporary-file> npm run test
                      # 59 files, 626 tests
npm run build-prod
npm run api:generate
git diff --check
```

Both production Webpack builds completed successfully.  API generation
reproduced the intended client-model changes.  Its unrelated, pre-existing
`dhcpStatus.ts` required/optional drift was excluded.

Known limitations: the unmodified Node 26 test command needs the temporary
local-storage file above because existing intl tests otherwise fail before
this feature is exercised.  Existing locale freshness/unused-key warnings are
also unrelated and no locale files change.  Playwright E2E, a live-browser
flow, and a live DNS runtime smoke test were not run; the feature is instead
covered at the API, persistence, effective-settings, DNS-message, store, and
component seams.

Two independent final reviews found no confirmed blocker.  No file was staged,
committed, or pushed.  If maintainers and the author want this approach, I can
provide the complete current-master diff rather than asking the old,
conflicting head `d3d0de62` to absorb piecemeal fixes.

PR #7783, opened 2025-04-22
Add bind-tools to container

beschoenen:
Bind tools would be very useful to have in a kubernetes setting where healthchecks could use `dig` or `nslookup` to check whether everything is still running as expected.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,7 +21,7 @@ LABEL \
 	org.opencontainers.image.version=$VERSION
 
 # Update certificates.
-RUN apk --no-cache add ca-certificates libcap tzdata && \
+RUN apk --no-cache add ca-certificates libcap tzdata bind-tools && \
 	mkdir -p /opt/adguardhome/conf /opt/adguardhome/work && \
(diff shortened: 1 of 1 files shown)


Jared0430:
This would be amazing!

Sil3ntVip3r:
I replayed exact head `6de66b43` onto current master. Git's rename handling
applies the change cleanly to `docker/build.Dockerfile`.

However, the project's [Docker documentation](https://github.com/AdguardTeam/AdGuardHome/wiki/Docker#healthcheck)
deliberately recommends a custom image for healthcheck tooling after the
built-in healthcheck caused cross-runtime problems. Adding `bind-tools` to
every production image increases the runtime image and attack surface for an
optional probe implementation.

Given that documented policy, I recommend closing this PR. Deployments
needing `dig` or `nslookup` can extend the official image, while the built-in
reserved healthcheck DNS name remains available.

If bundled tooling is reconsidered later, it should first have maintainer
design approval plus measured image-size and cross-architecture impact.

beschoenen:
Fair enough, thanks for the write up.

PR #7791, opened 2025-04-25
chore: Docker SemVer tags

### LINKED ISSUE #7226: Add docker tags for major/minor/patch versions
### Prerequisites

### Description

As of the point of writing this, `v0.107.52` is the latest AdGuardHome version.
As I'm manually updating my images and not using something like release, I have to change the image tag every new release. It would be nice if I wouldn't have to do that for patch releases (minor releases too).

Currently these tags are provided:
- `0.107.52` - Version 0.107.52
- `0.107.51` - Version 0.107.51
- ...
- `latest` - Version 0.107.52

### Proposed solution

I would propose a tag system like this:
- `0.107.52` - Version 0.107.52
- `0.107.51` - Version 0.107.51
- `0.107`      - Version 0.107.52
- `0`             - Version 0.107.52
- `latest`      - Version 0.107.52

This would allow pinning to specific version and allows the user to choose which semver types they are okay with auto updating.
I for example would have pinned to `0.107` and would have automatically gotten the hotfix to `0.107.52`.

_No response_

RoboMagus:
This PR Closes #7226.

When releasing new Docker images, additional [SemVer tags](https://medium.com/@mccode/using-semantic-versioning-for-docker-image-tags-dfde8be06699) are created providing users with more freedom of pinning their AdGuardHome instance to e.g. a specific `minor` version, while still being able to automatically pull in `patch` releases.

### DIFF
--- a/scripts/make/build-docker.sh
+++ b/scripts/make/build-docker.sh
@@ -59,7 +59,14 @@ readonly docker_output
 
 case "$channel" in
 'release')
-	docker_version_tag="--tag=${docker_image_name}:${version}"
+	V_MAJOR=$(echo "${version}" | awk -F'[. +-]' -vOFS='.' '{ print $1 }')
+	V_MINOR=$(echo "${version}" | awk -F'[. +-]' -vOFS='.' '{ print $1, $2 }')
+	V_PATCH=$(echo "${version}" | awk -F'[. +-]' -vOFS='.' '{ print $1, $2, $3 }')
+	docker_version_tag="\
+		--tag=${docker_image_name}:${version} \
+		--tag=${docker_image_name}:${V_PATCH} \
+		--tag=${docker_image_name}:${V_MINOR} \
+		--tag=${docker_image_name}:${V_MAJOR}"
 	docker_channel_tag="--tag=${docker_image_name}:latest"
 	;;
 'beta')


RoboMagus:
@ainar-g, This PR has been open for a while now. Mind taking a look at it?

It's a simple PR with limited scope and would help out greatly with version management in deployments.

RoboMagus:
@ainar-g, mind taking a look?

Sil3ntVip3r (changes requested):
I reviewed exact head
`94c67a12ef727fdda6c7783ed6fb52b2c527c11e` and reproduced the release
command on both its original code and a clean application to current master
`091d9287fc81613f08072a8d5381214f5422df87`.

There is one release-blocking rebase defect.  The PR stores four `--tag=...`
flags in one multiline `docker_version_tag` string.  On the old base that
string was expanded unquoted.  Current master now appends
`"$docker_version_tag"` as one positional argument at
`scripts/make/build-docker.sh:135`, so a release build receives:

```text
--tag=...:v0.107.80 --tag=...:v0.107.80 --tag=...:v0.107 --tag=...:v0
```

as a **single** argv value.  Docker 29.7.1 rejects the equivalent value with
`invalid reference format`, so the release image publication step cannot
complete.  A fake-Docker channel matrix reproduced this only for `release`:

```text
release:     2 tag argv, 1 argv containing four --tag= values
beta:        2 tag argv, 0 combined
edge:        1 tag argv, 0 combined
development: 0 tag argv, 0 combined
```

There is also a smaller duplicate: `V_PATCH` is identical to the existing
exact `version` value for a release, so `v0.107.80` is emitted twice.

Please rebase and keep the existing exact-version tag as one argument, derive
only the additional minor and major values, and append each nonempty tag as a
separate positional argument (a small helper or loop is fine).  Do not encode
multiple CLI flags in one scalar.

A focused regression should put a fake `docker` first on `PATH` and capture
argv.  For `VERSION=v0.107.80`, assert four unique release tags (`latest`,
`v0.107.80`, `v0.107`, `v0`) and one tag per argv.  Table-test beta, edge, and
development as well, with and without `DOCKER_PUSH`; fail if any argument
contains whitespace followed by another `--tag=`.  That test specifically
protects the current quoted-argv implementation.

Validation performed:

```text
PASS: clean application of the PR commit to current master
PASS: git diff --check
PASS: sh -n
PASS: ShellCheck 0.11.0
PASS: repository shfmt
PASS: make sh-lint
EXPECTED FAIL: Docker CLI rejects the combined release-tag argument
```

The static shell gates are clean but cannot detect this argv-boundary bug.
There is no newer competing PR for #7226.

PR #7824, opened 2025-05-15
home: add service ready notification support on Linux

alaviss:
On systemd-based Linux systems, signal service readiness once the web service is up and running. This allows depending services to accurately rely on AdGuardHome being active and ready to service.

This feature is particularly useful with Podman, as the container manager can verify that the service starts correctly after an auto-update operation, and rollback if it doesn't.

This new feature does not do anything on non-systemd Linux or other operating systems.

### DIFF
--- a/internal/home/ready_linux.go
+++ b/internal/home/ready_linux.go
@@ -1,0 +1,49 @@
+//go:build linux
+
+package home
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"time"
+)
+
+// Notifies the service manager that the program is ready to serve
+func notifyReady() error {
+	return sdNotify("READY=1")
+}
+
+// Notifies the service manager that the program is beginning to reload its
+// configuration
+func notifyReload() error {
+	now := time.Now().UnixMicro()
+	return sdNotify(fmt.Sprintf("RELOADING=1\nMONOTONIC_USEC=%v", now))
+}
+
+// Implements the sd_notify mechanism
+//
+// Reference: https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html
+func sdNotify(message string) error {
+	socketPath := os.Getenv("NOTIFY_SOCKET")
+	if socketPath == "" {
+		return nil
+	}
+	socketAddr := net.UnixAddr{
+		Name: socketPath,
+		Net:  "unixgram",
+	}
+
+	conn, err := net.DialUnix("unixgram", nil, &socketAddr)
+	if err != nil {
+		return fmt.Errorf("connecting to %q: %w", socketAddr.String(), err)
+	}
+	defer conn.Close()
+
+	_, err = conn.Write([]byte(message))
+	if err != nil {
+		return fmt.Errorf("sending %q: %w", message, err)
+	}
+
+	return nil
+}
--- a/internal/home/ready_others.go
+++ b/internal/home/ready_others.go
@@ -1,0 +1,14 @@
+//go:build !linux
+
+package home
+
+// Notifies the service manager that the program is ready to serve
+func notifyReady() error {
+	return nil
+}
+
+// Notifies the service manager that the program is beginning to reload its
+// configuration
+func notifyReload() error {
+	return nil
+}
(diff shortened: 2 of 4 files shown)


idaundaw-debug (approved):

Sil3ntVip3r (changes requested):
I reviewed exact head `ff1f0209` against current master.

The direction is useful, but there are two startup-protocol blockers:

1. `web.start` launches `ListenAndServe` asynchronously and sends `READY=1`
before inspecting the bind result. DNS startup is also asynchronous. An
occupied HTTP or DNS port can therefore produce a readiness notification
immediately before the process exits. Please establish a successful
listener/startup barrier first, then notify exactly once. A regression test
should assert that bind failures emit no readiness datagram.
2. `MONOTONIC_USEC` is populated with `time.Now().UnixMicro()`, which is
wall-clock epoch time. systemd requires a `CLOCK_MONOTONIC` timestamp. Please
use a monotonic Linux clock or omit this field until it can be implemented
and tested correctly.

Normal shutdown also currently passes through the `ErrServerClosed` branch
and emits `RELOADING=1`; reload and shutdown should be distinguished.

Current master has moved the service template to
`internal/ossvc/config_linux.go`, so the patch needs to be rebased around
that layout.

Local verification on this exact head:

- `gofmt -d` on changed Go files: clean.
- `go test -race -count=1 ./internal/home`: pass.
- Linux amd64 static test-binary cross-compilation: pass.
- `git diff --check`: pass.

Those checks cover the ordinary path; the failed-bind, shutdown, and
notification-datagram cases still need focused tests.

Sil3ntVip3r:
I completed the current-`master` rewrite for the startup-protocol blockers in my
review, based on `369d47616342c8be82262bf798f2fed2f64b697f`.

The replacement design:

- starts DNS synchronously, so a DNS bind failure terminates before readiness;
- pre-binds the plain HTTP listener and sends readiness only after that bind
  succeeds;
- serves on the already-bound listener, avoiding the prior
  notify-before-`ListenAndServe` race;
- emits exactly `READY=1` once, with no wall-clock `MONOTONIC_USEC` and no false
  `RELOADING=1` during shutdown;
- uses `Type=notify` in the current
  `internal/ossvc/config_linux.go` systemd template;
- does not re-notify when the HTTP listener is intentionally rebound;
- treats an empty `NOTIFY_SOCKET` as a no-op and supports both filesystem and
  Linux abstract (`@name`) systemd notification sockets.

The abstract-socket case was an additional defect found during independent
review: systemd encodes it with `@`, while Go's Unix networking API requires a
leading NUL. The corrected implementation translates that representation and
has a real datagram regression test.

Validation:

```text
focused readiness + filesystem/abstract socket tests x10 -race   PASS
internal/home and internal/ossvc race suites                     PASS
focused go vet                                                    PASS
Linux amd64 test-binary cross-compilation                         PASS
Linux arm64 tests in a local Linux VM                             PASS
make go-check after the abstract-socket fix                       PASS
git diff --check + gofmt                                          PASS
```

The Linux VM was returned to its prior stopped state. HTTPS and HTTP/3 are
intentionally not initial readiness barriers; the mandatory DNS and plain HTTP
listeners are. One final maintainer-side validation remains appropriate: a real
systemd process smoke test showing `systemctl start` remains activating until
bind succeeds, becomes active after `READY=1`, and fails without readiness when
a required port is occupied.

The original branch is conflicting and its startup layout is obsolete, so this
should replace rather than incrementally patch that implementation.

Sil3ntVip3r:
Follow-up: I completed the real systemd smoke test for the current-master prototype described above.

Environment: Ubuntu 22.04 container, systemd 249 running as PID 1, Linux arm64 build of the patched tree, and a real `Type=notify` unit with `NotifyAccess=main`.

Observed results:

- Normal first-start path: systemd logged `Starting`; AdGuard Home initialized and bound `127.0.0.1:3000`; only then did systemd log `Started` and report `ActiveState=active`, `SubState=running`, `Result=success`.
- Occupied-port path: a separate AdGuard Home process held `127.0.0.1:3000`; starting the unit returned nonzero; the main process exited with status 1 after `bind: address already in use`; systemd reported `ActiveState=failed`, `SubState=failed`, `Result=exit-code`. It was never marked ready.
- The abstract-`NOTIFY_SOCKET` behavior remains covered by the real Unix-datagram test added after the earlier review.

This closes the remaining Linux/systemd runtime gap in my earlier validation. The first-start smoke does not exercise configured DNS-port startup, so I am still treating the focused `internal/home` tests as the evidence for synchronous DNS startup; the systemd test specifically validates readiness timing and bind-failure propagation.

PR #7849, opened 2025-05-30
Adds `wget` to container

stavros-k:
See:
https://github.com/truenas/apps/issues/2421


stavros-k:
I can change it to add wget instead, as only the busybox's wget has the issue with zombie processes

PR #7878, opened 2025-06-14
Fix for #6320

### LINKED ISSUE #6320: Sorting by Answer column, in DNS Rewrites tab, doesn't work correctly if CNAME and IP addresses present.
### Prerequisites

- [X] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) and found no answer

- [X] I want to report a bug and not [ask a question or ask for help](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a)

- [X] I have set up AdGuard Home correctly and [configured clients to use it](https://github.com/AdguardTeam/AdGuardHome/wiki/Clients). (Use the [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) for help with installing and configuring clients.)

Linux, AMD64 (aka x86_64)

### Installation

Docker

### Setup

On one machine

### AdGuard Home version

v0.107.38

### Action

Inside the DNS Rewrites column, the table enables users to sort, which doesn't work properly.

Add 3 DNS Rewrite entries:
- a.example.com -> 10.1.1.3
- b.example.com -> 10.1.1.5
- c.example.com -> b.example.com
- d.example.com -> 10.1.1.1

Try sorting by the Answer column. The data is not sorted by .

image

### Expected result

The table contents sorted in consistent way. 

* CNAME entries sorted alphabetically
* IPv4 and IPv6 address sorted
* CNAME entries before or after IP addresses (doesn't matter which, as long as it is consistent)

### Actual result

The table contents is not sorted in any particular order. The IPv4 and CNAME addresses are mixed together. 

When trying to perform the sort, the JavaScript console complains:

image

I think the core of the issue is in here `helpers.js:getAddressesComparisonBytes` function. It tries to parse the answer either as CIDR or IP address, when some of the values are CNAME. This causes an exception and sorting is dropped.

grogi:
Closes #6320 

This introduces additional sorting function - `sortAddress` - that extends the behaviour of existing `sortIp`. There were two options effectively here 
- expand the sortIp and (although unlikely) potentially modify the existing behaviour
- create a new sorting function.

The second approach was selected. The unit tests for the previous implementation pass 100%, there is a new set of tests for the new function as well. 

image

I also cleaned up a bit the code around sorting, added some additional type safety and removed a need for es-lint overrides.

### DIFF
--- a/client/src/__tests__/helpers.test.ts
+++ b/client/src/__tests__/helpers.test.ts
@@ -355,7 +355,23 @@ describe('sortIp', () => {
 describe('sortAddress', () => {
     testBaseSortFunction(sortAddress);
 
-    describe('mixed_withString', () => {
+    describe('domain_names_sorting', () => {
+        test('widcard before other names', () => {
+            const arr = [
+                'home.fritz.box', 
+                '*.home.fritz.box', 
+                'adguard-home.fritz.box'
+            ]
+
+            const sortedArr = [
+                '*.home.fritz.box', 
+                'adguard-home.fritz.box', 
+                'home.fritz.box'
+            ];
+
+            expect(arr.sort(sortAddress)).toStrictEqual(sortedArr);
+        })
+
         test('only non-ip strings', () => {
             const arr = [
                 'fritz.box', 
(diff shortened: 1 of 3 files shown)


Sil3ntVip3r:
Issue #6320 is already closed, and the production `client_v2` rewrite table
uses generic string comparison rather than parsing every answer as an IP, so
mixed CNAME/IP values no longer trigger this failure mode. This PR also
changes only the legacy client and currently fails `git diff --check`.
Closing it as obsolete appears appropriate.

Issue #4127, opened 2022-01-12
AdGuardHome 0.107.3 fails to compile on NetBSD 9.x - again


bbartlomiej:
### Issue Details

* **Version of AdGuard Home server:**
  * 0.107.3
* **How did you install AdGuard Home:**
  * Built from source
* **How did you setup DNS configuration:**
  * n/a
* **If it's a router or IoT, please write device model:**
  * n/a
* **CPU architecture:**
  * AMD64
* **Operating system and version:**
  * NetBSD 9.2

### Expected Behavior
Code builds

### Actual Behavior
Code build fails with:
`# github.com/AdguardTeam/AdGuardHome/internal/aghnet
internal/aghnet/net.go:26:9: undefined: ifaceHasStaticIP
internal/aghnet/net.go:31:9: undefined: ifaceSetStaticIP
internal/aghnet/net.go:74:9: undefined: canBindPrivilegedPorts
internal/aghnet/net.go:208:9: undefined: closePortChecker
internal/aghnet/net.go:218:9: undefined: isAddrInUse
*** Error code 2

Stop.
make[1]: stopped in /usr/pkgsrc/wip/adguardhome-0.107.2
*** Error code 1
`

### Screenshots
`# github.com/AdguardTeam/AdGuardHome/internal/aghnet
internal/aghnet/net.go:26:9: undefined: ifaceHasStaticIP
internal/aghnet/net.go:31:9: undefined: ifaceSetStaticIP
internal/aghnet/net.go:74:9: undefined: canBindPrivilegedPorts
internal/aghnet/net.go:208:9: undefined: closePortChecker
internal/aghnet/net.go:218:9: undefined: isAddrInUse
*** Error code 2

Stop.
make[1]: stopped in /usr/pkgsrc/wip/adguardhome-0.107.2
*** Error code 1
`
### Additional Information
Similar problem has been reported for 0.106.2 in the past: #3147

`netbsd# go117 version
go version go1.17.6 netbsd/amd64`

`netbsd# cat go-modules.mk
# $NetBSD$

GO_MODULE_FILES+=       cloud.google.com/go/@v/v0.26.0.mod
GO_MODULE_FILES+=       cloud.google.com/go/@v/v0.31.0.mod
GO_MODULE_FILES+=       cloud.google.com/go/@v/v0.34.0.mod
GO_MODULE_FILES+=       cloud.google.com/go/@v/v0.37.0.mod
[...]

PR #8142, opened 2025-12-01
Build on NetBSD

moubctez:
Add support for NetBSD.
This issue https://github.com/AdguardTeam/AdGuardHome/issues/4127 is still open for 3 years, so let's get it done.

Note:  `internal/aghnet/net_netbsd.go` is identical `internal/aghnet/net_freebsd.go`.


Sil3ntVip3r (changes requested):
The directly modified NetBSD packages cross-compile, but the full application
still does not:

```text
CGO_ENABLED=0 GOOS=netbsd GOARCH=amd64 go build ./...
internal/ossvc/config.go:17:2: undefined: configureOSOptions
```

NetBSD needs an `internal/ossvc` implementation or supported fallback for
`configureOSOptions`. Please add that and make the full NetBSD cross-build
part of verification; package-only compilation currently misses this blocker.

Sil3ntVip3r:
I completed the current-`master` repair and full NetBSD cross-build requested
in my review. The PR's existing NetBSD changes are viable after two small
additions:

1. Add `internal/ossvc/config_netbsd.go`:

```go
//go:build netbsd

package ossvc

import "github.com/kardianos/service"

// configureOSOptions defines additional settings of the service configuration
// on NetBSD.  conf must not be nil.
func configureOSOptions(_ *service.Config) {}
```

2. Add `-o -name '*_netbsd.go'` to the allowed OS suffixes in the
   `underscores` check in `scripts/make/go-lint.sh`.

The second repair is required because the PR correctly introduces NetBSD build
suffixes, but the repository's filename-policy script otherwise rejects them.

Validation on the current-master prototype:

```text
NetBSD amd64 full build                         PASS
NetBSD arm64 full build                         PASS
NetBSD amd64 go vet ./...                       PASS
native affected-package race suites             PASS
make go-check                                   PASS
  including full race suite and govulncheck     PASS (0 callable)
git diff --check + formatting                   PASS
```

The exact old PR head's full build reproduced the missing
`configureOSOptions` error. After adding the fallback, its full quality gate
progressed to unrelated stale toolchain/dependency vulnerabilities, so rebasing
onto master remains necessary.

Runtime DHCP and service installation were not claimed: no real NetBSD host was
available. The fallback intentionally provides compilation only and does not
advertise native service-manager support. With those two repairs and a rebase,
this PR is ready for maintainer runtime validation.

Sil3ntVip3r (changes requested):
Re-reviewed exact head 38d620c16617e7d26920b8b930b1a1213fa27d3c. The previous full-build blocker is fixed: NetBSD amd64 and arm64 full builds, NetBSD amd64 vet, compilation of the affected NetBSD test packages, and make go-check all pass locally.

One NetBSD correctness blocker remains. NetBSD officially supports a static interface through either ifconfig_nnX in /etc/rc.conf or /etc/ifconfig.nnX (https://man.netbsd.org/ifconfig.if.5). ifaceHasStaticIP currently walks only etc/rc.conf. On a standard static /etc/ifconfig.em0 configuration with no ifconfig_em0 assignment, it returns false; enableDHCP then calls the NetBSD unsupported static-IP setter and fails instead of starting DHCP.

Please recognize the interface-file form too, honoring NetBSD’s configuration precedence, and add a NetBSD test fixture for the file-only static configuration (plus variable/file precedence and a dynamic dhcp configuration). That keeps the new platform support correct beyond successful cross-compilation.

Sil3ntVip3r:
I prepared a small, ready-to-cherry-pick follow-up on top of the current head `38d620c16617e7d26920b8b930b1a1213fa27d3c`:

https://github.com/Sil3ntVip3r/AdGuardHome/commit/f4e8a078753b2a8282b9cbae1a6fe2dc80b58fbf

```text
git cherry-pick f4e8a078753b2a8282b9cbae1a6fe2dc80b58fbf
```

It recognizes the documented `/etc/ifconfig.<interface>` form when the
`ifconfig_<interface>` value in `/etc/rc.conf` is missing or empty, while
preserving NetBSD startup precedence: a nonempty `rc.conf` value, including
`dhcp`, wins over the file.  The new NetBSD fixture covers file-only static
configuration, empty-variable fallback, dynamic-variable precedence, static
variable precedence, and semicolon-delimited values.  I also restored the
FreeBSD-only test tag, since the NetBSD behavior now has its own fixture.

Validation completed locally:

- NetBSD amd64 and arm64 full cross-builds
- NetBSD amd64 vet and cross-compilation of the NetBSD fixture
- native `./internal/aghnet` and `./internal/dhcpd` race tests
- `make go-check`

I do not have a native NetBSD host here, so the new fixture still needs one
native NetBSD test run after cherry-picking.  No competing PR was opened.

Issue #961, opened 2019-08-25
RFC - Using MAC identifier in Client settings REQUIRES using AGH DHCP Server. Why?


AnthonyBe:
### Problem Description
Currently, Clients can only be identified by MAC address if AdGuard Home is also a DHCP server.
This is a significant limitation given the newness of limited capability of AGH DHCP server currently.

### Proposed Solution
As the client MAC will be in the network packets, why can't MAC address identifiers be used even if AGH is not the DHCP server? Requiring that AGH be the DHCP server significantly limits the abilty to use client filters and this would likely impact the vast majority of AGH users.

### Alternatives Considered
Do not wish to use AGH DHCP server at this time due to features missing that are available in more mature DHCP servers

ameshkov:
Any other way of using a MAC address would be unreliable. We can't simply synchronize the IP<->Mac table with the DHCP server, there will always be a time when it's out of sync.

> Do not wish to use AGH DHCP server at this time due to features missing that are available in more mature DHCP servers.

Could you please explain what important features are missing?

> AGH DHCP Server is still considered "Experimental" (it is labeled this way in the UI)

Well, yeah, but it's close to removing this label:)

davidbdeath:
One problem with the DHCP server I ran into. My server had a static IP address assigned by my router. When I switched over to using AdGuards DHCP server I was unable to set that IP as it's static when I tried it said IP already in use. Why can't assign the IP a device already has as it's static?

szolin:
> when I tried it said IP already in use
It was already fixed in #833, it will be published in the next version.

AnthonyBe:
> 
> Could you please explain what important features are missing?
> 

Apologies if anything below is incorrect, but it is based on my observation so I may have overlooked or just not worked out how to do something, or a feature might just not be obviously documented.
I'm basing my comparison on Windows Server DHCP Server which I currently use, and within that, the features I currently use.

1. Doesn't work on Windows (I just tried to configure a scope and got error "Can't detect static IP: not supported on Windows")
2. Doesn't support setting DHCP Reservations
3. Not sure, but doesn't appear to support multiple IP ranges, or conversely, range exclusions?
4. Not option to define DNS Server IP (is this 'hard coded' assuming that AGH's IP is offered as DNS server address?).
5. No option to offer multiple IP addresses for DNS setting (i.e. primary and secondary DNS) so clients have fallback
6. DHCP Logging (audit logs)
[...]

PR #8248, opened 2026-02-15
client: identify persistent clients via IPv6 NDP neighbor table

AndyHazz:
## Summary

- Persistent clients configured with a MAC address are only matched when they query over IPv4, because `MACByIP()` only knows about DHCPv4 leases.  Over IPv6 the same device shows up as an unidentified client, so per-client filtering doesn't apply to it.
- This adds the kernel's IPv6 neighbor table (NDP) as a fallback for the IP-to-MAC lookup in `findByIP()`, `FindLoose()`, and `ApplyClientFiltering()`.
- It also covers part of #961 for IPv6: no DHCP server is needed, only that the device is a neighbor on the local link.

## Motivation

On home networks where AdGuard Home is the DNS server, per-client rules (e.g. blocking a service on a child's device) are usually keyed on MAC address, since that's the only identifier that survives address changes.  With IPv6 enabled, devices send queries from SLAAC addresses that have no DHCPv4 lease, so those queries arrive as unidentified clients and the rules are skipped.

The kernel's neighbor table already holds the IPv6-to-MAC mapping, populated when it processes the incoming packets, but nothing consults it.  `internal/arpdb` can parse `ip neigh` output, but on Linux its first source, `/proc/net/arp`, succeeds and is IPv4-only, so the `ip neigh` fallback is never reached.

## Implementation

`internal/client/ndp.go` keeps a `map[netip.Addr]net.HardwareAddr` read from `ip -6 neigh`, run through `executil` with a deadline.

- The table is read at startup and every 30s after that.  A lookup that finds its address missing, or the data expired, additionally schedules a single background read.  Nothing on the DNS request path runs or waits for a command, and no command runs while the client index lock is held.
- Reads are limited to one per 30s, counted on failure as well as success, and only one read runs at a time.  A failed read keeps the previously known neighbors, which stay usable until they expire.
- Data that no successful read has confirmed within 60s isn't used for identifying clients at all: an address may have been reassigned to another device, and a wrong MAC address means the settings of a wrong client.  With the periodic reread, this only comes up when the reads themselves are failing.
- Link-local entries are keyed with the interface from the `dev` field, e.g. `fe80::1%eth0`, since the same link-local address can be in use on several links and request addresses carry their zone.
- Nothing is read unless at least one persistent client is identified by a MAC address.
- Linux only, as `ip -6 neigh` is Linux-specific.  `ndp -an` on BSD/Darwin and `netsh interface ipv6 show neighbors` on Windows are left as a TODO.

## Test plan

- [x] `go test -race -count=1 ./...` and `make go-lint`, also against a merge with current `master`.
- [x] Unit tests for the parser, including entries with no `lladdr`, malformed addresses, and the same link-local address on two interfaces.
- [x] Behavior tests for the reader: rate limiting across failed reads, one read in flight, data kept but no longer used once expired, expired data replaced after a read, IPv4 and IPv4-in-IPv6 addresses ignored.
- [x] Integration tests for `Find`, `FindLoose`, and `ApplyClientFiltering` identifying an IPv6 client through the neighbor table, plus unknown addresses, a failing command, expired data not applying a client's settings, and the settings applying again after a successful read.
- [x] Manual: Raspberry Pi (armv7l) running AGH as DNS + DHCP server; a laptop querying over IPv6 is matched to its persistent client, IPv4 queries unchanged.

## Question for maintainers

Would you rather have this in `internal/arpdb`?  Adding the IPv6 neighbors to the Linux implementation there and building the IP-to-MAC index from `Neighbors()` would drop the separate command and cache from `internal/client` entirely, and would extend MAC matching to IPv4 without DHCP as well.  The cost is that the data would only be as fresh as the 10-minute ARP cycle, which is too slow for filtering a device that has just joined, so it would need its own refresh policy anyway.  Happy to rework it that way if you prefer.

### DIFF
--- a/internal/client/storage.go
+++ b/internal/client/storage.go
@@ -1,12 +1,16 @@
 package client
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"log/slog"
 	"net"
 	"net/netip"
+	"os/exec"
 	"slices"
+	"strings"
 	"sync"
 	"time"
 
@@ -114,6 +118,10 @@ type StorageConfig struct {
 	// configuration file.  Each client must not be nil.
 	InitialClients []*Persistent
 
+	// NDPData returns the raw output of the IPv6 neighbor table.  If nil,
+	// defaults to running "ip -6 neigh".
+	NDPData func() ([]byte, error)
+
 	// ARPClientsUpdatePeriod defines how often [SourceARP] runtime client
 	// information is updated.
 	ARPClientsUpdatePeriod time.Duration
@@ -150,6 +158,15 @@ type Storage struct {
 	// arpDB is used to update [SourceARP] runtime client information.
 	arpDB arpdb.Interface
 
+	// ndpData returns the raw output of the IPv6 neighbor table.
+	ndpData func() ([]byte, error)
+
+	// ndpCache maps IPv6 addresses to MAC addresses from the kernel NDP
+	// neighbor table.  Used to identify persistent clients querying over IPv6.
+	ndpCache   map[netip.Addr]net.HardwareAddr
+	ndpCacheMu sync.RWMutex
+	ndpCacheAt time.Time
+
 	// done is the shutdown signaling channel.
 	done chan struct{}
 
@@ -173,6 +190,11 @@ func NewStorage(ctx context.Context, conf *StorageConfig) (s *Storage, err error
 	tags := slices.Clone(allowedTags)
 	slices.Sort(tags)
 
+	ndpData := conf.NDPData
+	if ndpData == nil {
+		ndpData = defaultNDPData
+	}
+
 	s = &Storage{
 		logger:                 conf.Logger,
 		mu:                     &sync.Mutex{},
@@ -182,6 +204,8 @@ func NewStorage(ctx context.Context, conf *StorageConfig) (s *Storage, err error
 		dhcp:                   conf.DHCP,
 		etcHosts:               conf.EtcHosts,
 		arpDB:                  conf.ARPDB,
+		ndpData:                ndpData,
+		ndpCache:               make(map[netip.Addr]net.HardwareAddr),
 		done:                   make(chan struct{}),
 		allowedTags:            tags,
 		arpClientsUpdatePeriod: conf.ARPClientsUpdatePeriod,
@@ -241,6 +265,91 @@ func (s *Storage) ReloadARP(ctx context.Context) {
 	if s.arpDB != nil {
 		s.addFromSystemARP(ctx)
 	}
+
+	s.refreshNDP()
+}
+
+// defaultNDPData returns the output of "ip -6 neigh".
+func defaultNDPData() ([]byte, error) {
+	return exec.Command("ip", "-6", "neigh").Output()
+}
+
+// refreshNDP reads the IPv6 neighbor table and caches the IPv6 address to MAC
+// address mappings.
+func (s *Storage) refreshNDP() {
+	out, err := s.ndpData()
+	if err != nil {
+		return
+	}
+
+	cache := parseNDPNeigh(out)
+
+	s.ndpCacheMu.Lock()
+	s.ndpCache = cache
+	s.ndpCacheAt = time.Now()
+	s.ndpCacheMu.Unlock()
+}
+
+// parseNDPNeigh parses the output of "ip -6 neigh" and returns a map of IPv6
+// addresses to MAC addresses.  The expected input format:
+//
+//	fe80::1 dev eth0 lladdr aa:bb:cc:dd:ee:ff REACHABLE
+func parseNDPNeigh(data []byte) (cache map[netip.Addr]net.HardwareAddr) {
+	cache = make(map[netip.Addr]net.HardwareAddr)
+	sc := bufio.NewScanner(bytes.NewReader(data))
+
+	for sc.Scan() {
+		fields := strings.Fields(sc.Text())
+		if len(fields) < 5 {
+			continue
+		}
+
+		ip, parseErr := netip.ParseAddr(fields[0])
+		if parseErr != nil {
+			continue
+		}
+
+		for i, f := range fields {
+			if f == "lladdr" && i+1 < len(fields) {
+				mac, macErr := net.ParseMAC(fields[i+1])
+				if macErr == nil {
+					cache[ip] = mac
+				}
+
+				break
+			}
+		}
+	}
+
+	return cache
+}
+
+// macFromNDP returns the MAC address for the given IPv6 address from the NDP
+// neighbor cache.  If the cache is stale (>30s) and the address is not found,
+// it refreshes the cache and retries once.
+func (s *Storage) macFromNDP(ip netip.Addr) net.HardwareAddr {
+	if !ip.Is6() {
+		return nil
+	}
+
+	s.ndpCacheMu.RLock()
+	mac := s.ndpCache[ip]
+	stale := time.Since(s.ndpCacheAt) > 30*time.Second
+	s.ndpCacheMu.RUnlock()
+
+	if mac != nil {
+		return mac
+	}
+
+	if stale {
+		s.refreshNDP()
+
+		s.ndpCacheMu.RLock()
+		mac = s.ndpCache[ip]
+		s.ndpCacheMu.RUnlock()
+	}
+
+	return mac
 }
 
 // addFromSystemARP adds the IP-hostname pairings from the output of the arp -a
@@ -568,6 +677,12 @@ func (s *Storage) findByIP(addr netip.Addr) (p *Persistent, ok bool) {
 	}
 
 	foundMAC := s.dhcp.MACByIP(addr)
+	if foundMAC == nil {
+		// Fall back to the NDP neighbor table for IPv6 addresses that
+		// don't have a corresponding DHCPv4 lease.
+		foundMAC = s.macFromNDP(addr)
+	}
+
 	if foundMAC != nil {
 		return s.index.findByMAC(foundMAC)
 	}
@@ -594,6 +709,10 @@ func (s *Storage) FindLoose(ip netip.Addr, id string) (p *Persistent, ok bool) {
 	}
 
 	foundMAC := s.dhcp.MACByIP(ip)
+	if foundMAC == nil {
+		foundMAC = s.macFromNDP(ip)
+	}
+
 	if foundMAC != nil {
 		return s.index.findByMAC(foundMAC)
 	}
@@ -775,6 +894,10 @@ func (s *Storage) ApplyClientFiltering(id string, addr netip.Addr, setts *filter
 
 	if !ok {
 		foundMAC := s.dhcp.MACByIP(addr)
+		if foundMAC == nil {
+			foundMAC = s.macFromNDP(addr)
+		}
+
 		if foundMAC != nil {
 			c, ok = s.index.findByMAC(foundMAC)
 		}
--- a/internal/client/storage.go
+++ b/internal/client/storage.go
@@ -1,16 +1,12 @@
 package client
 
 import (
-	"bufio"
-	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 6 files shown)


Sil3ntVip3r (commented):
The failed-refresh path can turn this into a per-query process storm. `macFromNDP` decides whether to retry from `ndpCacheAt`, but `refreshNDP` updates that timestamp only after `ip -6 neigh` succeeds. If the command is unavailable or returns an error, the startup refresh leaves the timestamp zero (and a later failure leaves it expired), so every unknown IPv6 DNS request synchronously launches the command again. `ApplyClientFiltering` is on the normal request path, and concurrent requests can all observe the stale value before any refresh completes, multiplying processes and request latency. Please track the last attempt on both success and failure and serialize/singleflight refreshes, while retaining the last successful cache on errors. A regression test can inject failing and blocking `NDPData` functions, issue repeated and concurrent unknown-IPv6 lookups, and assert at most one attempt per cooldown and one refresh in flight.

internal/client/storage.go
```diff
@@ -241,6 +265,91 @@ func (s *Storage) ReloadARP(ctx context.Context) {
...
+
+// defaultNDPData returns the output of "ip -6 neigh".
+func defaultNDPData() ([]byte, error) {
+	return exec.Command("ip", "-6", "neigh").Output()
+}
+
+// refreshNDP reads the IPv6 neighbor table and caches the IPv6 address to MAC
+// address mappings.
+func (s *Storage) refreshNDP() {
+	out, err := s.ndpData()
```
Sil3ntVip3r: **Medium severity, high confidence:** Please make this refresh bounded and singleflight, and record the attempt even when `ndpData` fails. `macFromNDP` treats the cache as stale until a successful refresh, so an unavailable/failing `ip` command is retried by every unknown IPv6 request; concurrent misses all enter this function at once. It is also called synchronously from `Find`/`FindLoose` while `Storage.mu` is held and from the DNS filtering path, with no command context or timeout. Review-only tests observed 3 attempts for 3 consecutive misses, overlapping refreshes in 20/20 runs, and an unrelated lookup blocked behind a slow refresh. Keep the last successful cache, add an attempt cooldown plus singleflight/serialization, and run the external command with a deadline outside the main client-index lock. Cover failed, slow, and concurrent refreshes in regression tests.

internal/client/storage.go
```diff
@@ -241,6 +265,91 @@ func (s *Storage) ReloadARP(ctx context.Context) {
...
+	cache = make(map[netip.Addr]net.HardwareAddr)
+	sc := bufio.NewScanner(bytes.NewReader(data))
+
+	for sc.Scan() {
+		fields := strings.Fields(sc.Text())
+		if len(fields) < 5 {
+			continue
+		}
+
+		ip, parseErr := netip.ParseAddr(fields[0])
```
Sil3ntVip3r: **Medium severity, high confidence:** Preserve the interface zone for link-local neighbors. Linux reports `fe80::1 dev eth0 ...`, so this stores the unzoned key `fe80::1`, while DNS socket addresses retain their scope as `fe80::1%eth0`; `macFromNDP` then misses and the feature does not identify that client. Simply stripping the incoming zone is unsafe because the same link-local address may exist on multiple interfaces. Parse the `dev` field and key link-local entries with `ip.WithZone(interfaceName)` (global addresses should remain unzoned). A regression test with identical link-local addresses on two interfaces should verify that each zoned source resolves to its own MAC.

Sil3ntVip3r (changes requested):
I revalidated exact head `26f2f8ca7ee483464f007790e63ea13a32d4da3b`. The earlier refresh-storm concern remains, and deterministic review-only tests now confirm both repeated retries and concurrent refreshes. I also found a separate correctness issue for scoped link-local IPv6 clients; details and smallest fixes are inline.

Validation performed:

- `go test -race -count=1 ./internal/client` passed at the exact head.
- `go test -race -count=1 ./...` passed at the exact head.
- The PR merged cleanly onto current upstream `master` (`091d9287fc81613f08072a8d5381214f5422df87`), and `go test -race -count=1 ./internal/client` passed in that synthetic merge.
- Review-only failure-cooldown test: three consecutive misses caused three refresh attempts instead of one.
- Review-only singleflight test: two refreshes overlapped in all 20 runs.
- Review-only lock test: a blocked refresh held the client-storage mutex long enough to block an unrelated lookup.
- Review-only scoped-address test: `fe80::1%eth0` failed to match the parsed `fe80::1 dev eth0` entry in all 20 runs.
- `make go-check` stopped in `go-lint`: `internal/client/ndp_internal_test.go` is not `gofmt`-clean (one map-literal alignment change). Please run the repository formatter as part of the update.
- GitHub reports no checks for this branch.

The normal-path tests are useful, but the request-path refresh behavior and link-local keying need correction before this is safe to merge.

internal/client/storage.go
```diff
@@ -241,6 +265,91 @@ func (s *Storage) ReloadARP(ctx context.Context) {
...
+	if !ip.Is6() {
+		return nil
+	}
+
+	s.ndpCacheMu.RLock()
+	mac := s.ndpCache[ip]
+	stale := time.Since(s.ndpCacheAt) > 30*time.Second
+	s.ndpCacheMu.RUnlock()
+
+	if mac != nil {
```
Sil3ntVip3r: **Medium severity, high confidence:** Do not return an expired positive entry before refreshing it. If an IPv6 address moves from device/MAC A to device/MAC B, this branch keeps returning MAC A without consulting the now-updated kernel table; the production periodic refresh interval is ten minutes, so A's client filtering or upstream policy can be applied to B throughout that window. A review-only test seeded an expired A mapping and made `ndpData` return B: `macFromNDP` returned A and called the provider zero times in all 20 runs. When `stale` is true, refresh before accepting either a hit or miss; if that refresh fails, retaining the prior hit is a reasonable availability fallback. Add a regression test covering old-MAC to new-MAC replacement.

Sil3ntVip3r (commented):
One additional identity-correctness case was confirmed after the main review: an expired cache hit is returned without any refresh, so a reassigned IPv6 address can retain the previous device's MAC mapping until the ten-minute periodic refresh. The deterministic review-only test failed in all 20 runs; details are inline.

Sil3ntVip3r (changes requested):
Thanks for the substantial update. I re-reviewed exact head
`908e24d9f066c50c73afd5a6f448b2fefff9e62c`. The asynchronous read,
attempt cooldown, single-flight guard, command timeout, Linux gating, and
link-local zone handling resolve the earlier findings. One identity-correctness
blocker remains in the new stale-data contract.

`ndpNeighbors.macFor` reads the cached MAC, notices that the data is at least
30 seconds old, schedules a refresh, and still returns the expired MAC
(`internal/client/ndp.go:109-118`). If the IPv6 address has been reassigned,
the triggering request is attributed to the previous device. If the refresh
fails, the same expired mapping continues to be returned for every request,
indefinitely, with another attempt only after each cooldown.

This is not only stale display data: `Storage.findByIP`, `FindLoose`, and
`ApplyClientFiltering` use that MAC to select a persistent client. The new
device can therefore inherit the previous client's filtering, blocked-service,
and custom-upstream policy until a refresh succeeds.

The submitted test currently makes this behavior explicit:
`TestNDPNeighbors_macFor/outdated_data_is_replaced` expects the old MAC on the
request that detects expiry. I added a disposable exact-head regression that
performed a successful read, advanced the fake clock by `ndpDataTTL`, then made
the refresh fail. It expected an expired identity not to be returned. The test
failed both before and after the failed refresh:

```text
Expected nil, but got aa:bb:cc:dd:ee:ff
--- FAIL: TestReviewNDPNeighbors_staleIdentityIsNotUsed
```

The smallest safe fix is to schedule the background read but return no MAC once
`updated` is expired. The last successful map can remain stored across read
errors, but it should not be used for identity or policy until a successful
read refreshes `updated`. Please adjust the outdated-data test accordingly and
add a storage-level regression proving that an expired NDP mapping does not
apply the previous persistent client's settings; after a successful refresh,
the new mapping should apply normally.

Fresh validation on the clean exact head:

```text
PASS: go test -race -count=20 -run '^TestNDPNeighbors_macFor$' ./internal/client
PASS: go test -race -count=1 ./internal/client
PASS: go vet ./internal/client/...
PASS: gofmt inspection and git diff --check
EXPECTED FAIL: disposable stale-identity regression above
```

The disposable test was removed afterward and the checkout is clean. GitHub's
current build/lint `action_required` runs contain zero jobs, so they are fork
approval gates rather than test failures.

AndyHazz:
Conceded - an expired mapping selecting a persistent client is worse than not identifying one, and the unbounded case when reads keep failing was a real gap.

`macFor` now returns no MAC once no successful read has confirmed the data within the TTL, and only schedules the read. `outdated_data_is_replaced` is replaced by `expired_data_is_replaced`, and `failed_read_keeps_data` covers your scenario directly: successful read, clock advanced by the TTL, refresh failing, no MAC returned. Storage-level regressions added as asked: `expired_data_not_applied` (expired mapping applies no client settings, and still doesn't after the failing read completes) and `refreshed_data_applied`.

One addition beyond the fix. Returning nil on expiry on its own costs identification in normal operation: `updated` only advances on a successful read, so one request per TTL finds the data expired and gets no client, which for per-client filtering means that query silently falls back to the global settings. So the table is now also reread on a 30s ticker (`periodicNDPUpdate`, alongside the existing `periodicARPUpdate`, with the same Linux and MAC-configured gating), and `ndpDataTTL` is two intervals, so one transient failure is tolerated. With that, a request only finds the data expired when the reads themselves are failing - which is exactly when it shouldn't be trusted.

`make go-lint` clean, `go test -race -count=20` on the NDP tests and `-count=1 ./...` pass, also against a merge with current `master` (`369d4761`).

Sil3ntVip3r (changes requested):
Thanks for addressing the stale-identity case. I rechecked exact head
`83d9b0bee75e7bb9bdefafcc6ec34b72fe3cfc42`. The implementation now returns no
MAC once the NDP data expires, preserves the previous map only while it remains
within its TTL, and covers both failed-refresh policy isolation and successful
re-identification. The earlier request-storm, single-flight, timeout, and
link-local-zone findings also remain resolved.

One merge-integration issue remains. This branch is 130 commits behind current
`master` (`369d47616342c8be82262bf798f2fed2f64b697f`). Although the branch view
shows the NDP changelog entry under its old `Unreleased` section, GitHub's
generated merge commit places the entry and its `[#961]` reference inside the
already released `v0.107.72` section:

```text
## [v0.107.72] - 2026-02-19
...
### Added
...
- On Linux, persistent clients ... ([#961]).
```

Please update the branch from current `master` and relocate both the NDP entry
and the `[#961]` link into the current `Unreleased` section. No production-code
change is requested.

Exact-head validation completed locally:

```text
focused NDP race tests, count 20
PASS

complete internal/client race tests, count 3
PASS

go vet ./internal/client/...
PASS

make go-check
PASS
```

A clean synthetic merge onto current `master` also passed the focused and
package race tests, vet, `make go-check`, `make go-os-check`, `make md-lint`,
and `git diff --check`. The exact-head build and lint runs currently show
`action_required` with zero jobs; those are fork-approval gates, not test
failures.

Sil3ntVip3r (approved):
Re-reviewed exact head `c3f50ec1fa684c264aef1a6a1d6ef80445cb6b6d`. The requested integration update is complete: the branch is fully based on current `master` (`369d47616342c8be82262bf798f2fed2f64b697f`), the range-diff preserves the previously reviewed NDP implementation and regressions, and both the NDP entry and `[#961]` reference now live in the current `Unreleased` section rather than `v0.107.72`.

Fresh exact-head validation passed:

- focused NDP race tests, count 20;
- complete `internal/client` race tests, count 3;
- `go vet ./internal/client/...`;
- `make go-check`;
- `make go-os-check`;
- `make md-lint`;
- `git diff --check upstream/master...HEAD`.

I found no remaining blocker at this exact head. The current build and lint workflow runs are `action_required` with zero jobs, so they are fork-approval gates rather than test failures.

PR #8274, opened 2026-03-02
all: add per-client filter list assignment

fegauthier-paragon:
## Summary

Adds the ability to assign specific blocklists and allowlists to individual clients. When a client has `use_own_filter_lists` set, only the lists named in `filter_list_ids` and `allow_filter_list_ids` apply to it. Rules from the user's own list always apply, either way.

## What changed since the first review

**Backward compatibility.** The flag is now `use_own_filter_lists`, in the positive form, so its zero value keeps the global lists. The previous `use_global_filter_lists` decoded to `false` for every config file and every API request written before this feature existed, which flipped all existing clients onto an empty own list set and silently dropped every downloaded list. Thanks @Sil3ntVip3r for spotting it. The positive form fixes YAML and JSON at once, with no schema migration. Regression tests cover a legacy YAML client and an add/update payload without the field.

**Resource usage.** A globally disabled filter list is now downloaded and loaded into the engine only while at least one client references it, through a new `ClientFilterListIDs` hook on `filtering.Config`. A configuration without per client lists keeps exactly the memory and traffic it has today. In that case no match time filtering happens at all, so requests from clients on the global lists stay allocation free. Disabling a list also no longer triggers a download.

**Rule selection.** The match result is now narrowed with `rules.GetDNSBasicRule`, the same function `urlfilter` uses, instead of a hand written scan. `$badfilter`, `$replace`, and rule priority behave identically for clients with and without their own lists. The previous version resolved conflicts by position, so a rule negated by `$badfilter` could still block.

**Concurrency.** The enabled list ID sets are swapped inside the same `engineLock` critical section as the engines they describe, since `matchHost` reads both together. They were previously written under `filtersMu` only.

**Housekeeping.** OpenAPI schema and both changelogs updated. Unrelated `.gitignore` and `package-lock.json` churn removed. Rebased onto master.

## Test plan

* [x] `make go-lint` clean, same baseline as master
* [x] `make go-test` clean, race detector enabled
* [x] Legacy YAML config with no `use_own_filter_lists`: both global blocklists still apply
* [x] Client with `filter_list_ids: [1]`: list 1 blocks, list 2 does not, user rules still apply
* [x] Globally disabled list assigned to a client: the list works, and it is absent from the engine when no client uses it
* [x] Allowlists: global allowlist exempts, an unassigned allowlist does not, an assigned one does
* [x] `POST /control/clients/add` and `/update` without the field keep the global lists
* [x] YAML round trip saves and reloads the new fields
* [x] Assigning a disabled list through the API takes effect without a restart

Verified against a running instance with `dig`, not only in unit tests. Building the previous revision of this branch and pointing it at a legacy config reproduced the reported regression, which confirms the fix.

### DIFF
--- a/internal/filtering/filter.go
+++ b/internal/filtering/filter.go
@@ -21,6 +21,7 @@ import (
 	"github.com/AdguardTeam/golibs/errors"
 	"github.com/AdguardTeam/golibs/ioutil"
 	"github.com/AdguardTeam/golibs/logutil/slogutil"
+	"github.com/AdguardTeam/urlfilter/rules"
 )
 
 // filterDir is the subdirectory of a data directory to store downloaded
@@ -119,19 +120,19 @@ func (d *DNSFilter) filterSetProperties(
 		"filter_url", flt.URL,
 	)
 
-	defer func(oldURL, oldName string, oldEnabled bool, oldUpdated time.Time, oldRulesCount int) {
+	// Restore the whole entry on failure, including the checksum, so that a
+	// failed update doesn't leave metadata that disagrees with the contents on
+	// disk.
+	defer func(old FilterYAML) {
 		if err != nil {
-			flt.URL = oldURL
-			flt.Name = oldName
-			flt.Enabled = oldEnabled
-			flt.LastUpdated = oldUpdated
-			flt.RulesCount = oldRulesCount
+			*flt = old
 		}
-	}(flt.URL, flt.Name, flt.Enabled, flt.LastUpdated, flt.RulesCount)
+	}(*flt)
 
 	flt.Name = newList.Name
 
-	if flt.URL != newList.URL {
+	urlChanged := flt.URL != newList.URL
+	if urlChanged {
 		if d.filterExistsLocked(newList.URL) {
 			return false, errFilterExists
 		}
@@ -148,12 +149,10 @@ func (d *DNSFilter) filterSetProperties(
 		shouldRestart = true
 	}
 
-	if !flt.Enabled {
-		// TODO(e.burkov):  The validation of the contents of the new URL is
-		// currently skipped if the rule list is disabled.  This makes it
-		// possible to set a bad rules source, but the validation should still
-		// kick in when the filter is enabled.  Consider changing this behavior
-		// to be stricter.
+	if d.shouldUnload(flt, isAllowlist, urlChanged) {
+		// TODO(e.burkov):  The contents of a rule list that stays out of the
+		// engine are not validated, which makes it possible to keep a bad rules
+		// source.  Consider changing this behavior to be stricter.
 		flt.unload()
 
 		return shouldRestart, err
@@ -163,7 +162,7 @@ func (d *DNSFilter) filterSetProperties(
 		return false, nil
 	}
 
-	return d.update(flt)
+	return d.update(flt, urlChanged)
 }
 
 // filterExists returns true if a filter with the same url exists in d.  It's
@@ -218,9 +217,14 @@ func (d *DNSFilter) filterAdd(flt FilterYAML) (err error) {
 	return nil
 }
 
-// Load filters from the disk
-// And if any filter has zero ID, assign a new one
-func (d *DNSFilter) loadFilters(ctx context.Context, array []FilterYAML) {
+// loadFilters loads filters from the disk and assigns a new ID to any filter
+// that has a zero one.  A disabled filter is only loaded if clientIDs contains
+// it, since a client may use a filter that is disabled globally.
+func (d *DNSFilter) loadFilters(
+	ctx context.Context,
+	array []FilterYAML,
+	clientIDs map[rules.ListID]bool,
+) {
 	for i := range array {
 		filter := &array[i] // otherwise we're operating on a copy
 		if filter.ID == 0 {
@@ -230,8 +234,7 @@ func (d *DNSFilter) loadFilters(ctx context.Context, array []FilterYAML) {
 			filter.ID = newID
 		}
 
-		if !filter.Enabled {
-			// No need to load a filter that is not enabled
+		if !filter.Enabled && !clientIDs[filter.ID] {
 			continue
 		}
 
@@ -273,8 +276,13 @@ func (d *DNSFilter) tryRefreshFilters(block, allow, force bool) (updated int, is
 	return updated, isNetworkErr, ok
 }
 
-// listsToUpdate returns the slice of filter lists that could be updated.
-func (d *DNSFilter) listsToUpdate(filters *[]FilterYAML, force bool) (toUpd []FilterYAML) {
+// listsToUpdate returns the slice of filter lists that could be updated.  A
+// disabled filter list is only updated if clientIDs contains it.
+func (d *DNSFilter) listsToUpdate(
+	filters *[]FilterYAML,
+	clientIDs map[rules.ListID]bool,
+	force bool,
+) (toUpd []FilterYAML) {
 	now := time.Now()
 
 	d.conf.filtersMu.RLock()
@@ -283,7 +291,7 @@ func (d *DNSFilter) listsToUpdate(filters *[]FilterYAML, force bool) (toUpd []Fi
 	for i := range *filters {
 		flt := &(*filters)[i] // otherwise we will be operating on a copy
 
-		if !flt.Enabled {
+		if !flt.Enabled && !clientIDs[flt.ID] {
 			continue
 		}
 
@@ -313,9 +321,10 @@ func (d *DNSFilter) listsToUpdate(filters *[]FilterYAML, force bool) (toUpd []Fi
 func (d *DNSFilter) refreshFiltersArray(
 	ctx context.Context,
 	filters *[]FilterYAML,
+	clientIDs map[rules.ListID]bool,
 	force bool,
 ) (updateCount int, updateFilters []FilterYAML, updateFlags []bool, isNetErr bool) {
-	updateFilters = d.listsToUpdate(filters, force)
+	updateFilters = d.listsToUpdate(filters, clientIDs, force)
 	if len(updateFilters) == 0 {
 		return 0, nil, nil, false
 	}
@@ -342,7 +351,7 @@ func (d *DNSFilter) updateFilterList(
 ) (failNum int, updateFlags []bool) {
 	for i := range updateFilters {
 		uf := &updateFilters[i]
-		updated, err := d.update(uf)
+		updated, err := d.update(uf, false)
 		updateFlags = append(updateFlags, updated)
 		if err != nil {
 			failNum++
@@ -426,13 +435,15 @@ func (d *DNSFilter) refreshFiltersIntl(block, allow, force bool) (int, bool) {
 	var toUpd []bool
 	isNetErr := false
 
+	blockIDs, allowIDs := d.clientFilterListIDs()
 	if block {
-		updNum, lists, toUpd, isNetErr = d.refreshFiltersArray(ctx, &d.conf.Filters, force)
+		updNum, lists, toUpd, isNetErr = d.refreshFiltersArray(ctx, &d.conf.Filters, blockIDs, force)
 	}
 	if allow {
 		updNumAl, listsAl, toUpdAl, isNetErrAl := d.refreshFiltersArray(
 			ctx,
 			&d.conf.WhitelistFilters,
+			allowIDs,
 			force,
 		)
 
@@ -477,10 +488,10 @@ func removeOldFilterFile(ctx context.Context, l *slog.Logger, fltPath string) {
 }
 
 // update refreshes filter's content and a/mtimes of it's file.
-func (d *DNSFilter) update(filter *FilterYAML) (b bool, err error) {
+func (d *DNSFilter) update(filter *FilterYAML, replace bool) (b bool, err error) {
 	ctx := context.TODO()
 
-	b, err = d.updateIntl(ctx, filter)
+	b, err = d.updateIntl(ctx, filter, replace)
 	filter.LastUpdated = time.Now()
 	if !b {
 		chErr := os.Chtimes(
@@ -497,8 +508,13 @@ func (d *DNSFilter) update(filter *FilterYAML) (b bool, err error) {
 }
 
 // updateIntl updates the flt rewriting it's actual file.  It returns true if
-// the actual update has been performed.  flt must not be nil.
-func (d *DNSFilter) updateIntl(ctx context.Context, flt *FilterYAML) (ok bool, err error) {
+// the actual update has been performed.  When replace is true, the contents are
+// saved even if they parse to the same checksum.  flt must not be nil.
+func (d *DNSFilter) updateIntl(
+	ctx context.Context,
+	flt *FilterYAML,
+	replace bool,
+) (ok bool, err error) {
 	d.logger.DebugContext(ctx, "downloading update for filter", "id", flt.ID, "url", flt.URL)
 
 	var res *rulelist.ParseResult
@@ -524,7 +540,7 @@ func (d *DNSFilter) updateIntl(ctx context.Context, flt *FilterYAML) (ok bool, e
 		return false, err
 	}
 
-	return res.Checksum != flt.checksum, nil
+	return replace || res.Checksum != flt.checksum, nil
 }
 
 // readFromHTTP reads filter data from urlStr via HTTP and parses it into the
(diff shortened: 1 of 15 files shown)


Sil3ntVip3r (commented):
Compatibility concern: `internal/home/clients.go:193,222-224` adds `use_global_filter_lists` as a plain bool and immediately negates it, and `internal/home/clientshttp.go:61,219-221` does the same for JSON. Legacy YAML and older API clients omit this new field, so it decodes as false and becomes `UseOwnFilterLists=true` with no list IDs. `ApplyClientFiltering` then creates non-nil empty ID maps, causing the filtering path to discard all downloaded global-list matches for that persistent client. This can silently weaken filtering after an upgrade or an update from an older client. Could absence default safely to global lists, for example through an optional/absence-aware field plus migration, or a `use_own_filter_lists` field whose zero value preserves existing behavior? A regression test should load legacy YAML and an older add/update JSON payload without the field and verify that a downloaded global blocklist still applies.

fegauthier-paragon:
Good catch, thanks. You were right, and it was worse than a config issue: I built the previous revision and pointed it at a legacy config, and every downloaded blocklist stopped applying. Only the user rules survived.

Fixed by switching the field to `use_own_filter_lists` in the positive form, for both YAML and JSON, so the zero value keeps the global lists. That covers the API path too, which a schema migration alone would have missed. The flag is now explicit in `filtering.Settings` instead of relying on the ID map being non nil.

Added the regression tests you asked for: a legacy YAML client and an add/update payload without the field, both asserting the global lists still apply.

While in there I also made a globally disabled list get downloaded and loaded only while a client actually references it, so a setup without per client lists keeps the same memory and traffic as before.

Sil3ntVip3r (changes requested):
Reviewed exact head `9b2ac9505a760fd541ae6047f0e12375289a438d`.  The positive-form compatibility change fixes the earlier legacy-zero-value problem, but I found two remaining policy regressions that should be addressed before merge.

1. **Dashboard edits erase per-client filter-list settings (Medium, high confidence).**

   `client_v2/src/api/model/client.ts:7-62` has none of the three new fields, and `client_v2/src/stores/clientForm.ts:96-147` omits them from both `buildFormPayload` and `buildClientConfig`.  The save path therefore sends no per-client list policy.  `internal/home/clientshttp.go:223-225` unconditionally assigns the decoded zero values, and the update handler calls it at line 476.

   Trigger: configure `use_own_filter_lists` and list IDs through YAML or the API, then edit any unrelated property for that client in the current dashboard.  The update silently changes `UseOwnFilterLists` to false and clears both ID lists, reverting the client to global-list policy.  The new JSON regression only covers conversion with `prev=nil`; it does not exercise this UI/update round trip.

   The smallest fix is to regenerate the `Client` model and preserve all three fields through `ClientFormState`, `buildFormPayload`, and `buildClientConfig`, even if dedicated controls arrive later.  A presence-aware backend update would add compatibility protection for older API clients.  Please add a UI form/config round-trip test and an update regression that preserves an existing own-list client while another field changes.

2. **Assigning a changed disabled list can activate stale rules (Medium, high confidence).**

   When a downloaded list is disabled and unreferenced, changing its URL takes the early return in `internal/filtering/filter.go:135-160`.  Metadata is unloaded, but the existing ID file is left on disk (`filter.go:48-51`).  Assigning that list to a client then follows `internal/home/clientshttp.go:385-400` and the new `appendEnabledFilters` path at `filter.go:676-703`, which reads the unchanged file.  If automatic refresh is disabled (`internal/filtering/filtering.go:1235-1240`), nothing repairs it.

   I reproduced this at the exact head with local sources A and B: after downloading A, disabling the list, changing its URL to B, and assigning it, A's domain was still filtered and B's was not.  For allowlists, stale rules could unexpectedly exempt domains.

   The smallest fix is to track `urlChanged` and avoid the disabled/unreferenced early return when the URL changes, allowing the existing atomic update path to replace the file.  The candidate fix passed the regression 20 times under `-race` and the full filtering package.  Please cover both blocking and allowing lists across download, disable, URL change, assignment, and final filtering.

Independent validation on the exact head:

- focused client/filtering/home regressions, 20 times under `-race`;
- complete affected packages under `-race`;
- `make go-check`, including `govulncheck` with zero called vulnerabilities;
- `git diff --check`.

I did not find another confirmed locking, panic, migration, or serialization blocker.

fegauthier-paragon:
Both confirmed and fixed at `8665cb0f`. Thanks for the detail, the reproduction steps saved me a lot of time.

**1. Update erasing the policy.** You were right that the save path had no protection. Rather than only fixing the dashboard, I made the backend absence aware, which also covers older API clients and scripts: `use_own_filter_lists` is now `aghalg.NullBool` and the two ID fields are `*[]int64`, so an omitted field keeps the stored value while an empty array clears it. This follows the same pattern as `ignore_querylog`. `handleUpdateClient` now passes the stored client as `prev`, which is what `initPrev` was built for.

Verified at runtime: create a client with `use_own_filter_lists: true` and IDs, then update only its identifier with a payload that omits all three fields. The IP changes and the policy survives. Table driven test added for omitted, explicit false, empty array, and replacement.

The new UI in #8275 does send all three fields, and the generated `Client` model now has them.

One side effect worth flagging: passing `prev` also makes the documented preservation of `ignore_querylog`, `ignore_statistics`, and `upstreams_cache_enabled` actually work on update, since that branch was unreachable before. It matches what `openapi.yaml` already promises for those fields, but say the word if you would rather see it split out.

**2. Stale rules after a URL change.** Reproduced exactly as you described. The list file is named after the ID, so `unload` left the previous source's rules on disk and my new load path picked them up. Fixed by threading `urlChanged` into `shouldUnload`, so a changed URL always goes through the normal atomic update instead of the early return. Regression test covers download, disable, URL change, assignment, and final filtering, for both blocking and allowing lists. It fails on the previous revision, for both.

While in there I also guarded the `prev.BlockedServices` dereference in `copyBlockedServices`, since passing `prev` makes that branch reachable for the first time.

On the exact head: `make go-lint` clean at the same baseline as master, `go test -race ./...` clean, the focused regressions 20 times under `-race`, and the Playwright suite for the UI. The OpenAPI schema and changelog now document the absence semantics.

Sil3ntVip3r (changes requested):
Re-reviewed exact head `8665cb0f4703dfb98c4de1b3e5a2dfc4bf0969c2`.

The omitted-field/dashboard-update finding is fully addressed: the API now distinguishes omitted fields from explicit false/empty values, preserves the stored policy on update, and documents/tests those semantics.

One blocking defect remains:

**Medium, high confidence — changing a disabled list to a valid empty or comments-only source retains the previous rules.**

On a URL change, `filterSetProperties` calls `flt.unload()`, which resets the in-memory checksum to zero (`internal/filtering/filter.go:135-166`). A valid empty/comments-only source also parses to checksum zero (`internal/filtering/rulelist/parser.go:83-89,115-120`). `updateIntl` therefore reports “unchanged” at `filter.go:539`; `finalizeUpdate` removes the pending empty file at `filter.go:606-611`; and `update` only touches the old ID file's timestamps at `filter.go:497-505`.

Trigger: download a non-empty list, disable it while unreferenced, change its URL to an empty/comments-only list, then assign it to a client. The old ID file is loaded again. A blocklist continues blocking old domains; an allowlist can reactivate stale exceptions.

I reproduced this independently on the exact head: after the URL change succeeded, the ID file still contained `||old.example^\n`. The new regression uses a non-empty replacement, so it does not cover this zero-checksum case.

The smallest robust fix is to make a successful URL-change parse force atomic replacement regardless of checksum, while retaining the old file on download/parse/replace failure. Please also restore the old checksum during rollback. Add blocking and allowing cases for empty/comments-only replacements followed by client assignment, plus failed-download rollback coverage.

Independent validation otherwise passed:

- focused new regressions 20 times under `-race`;
- full `internal/client`, `internal/filtering`, and `internal/home` tests under `-race`;
- `make go-check`, including zero called vulnerabilities;
- `git diff --check`.

I found no other confirmed blocker.

fegauthier-paragon:
Confirmed and fixed at `3e563aa7`. Your zero checksum analysis was spot on.

`unload` zeroes the checksum and a comments only source parses to zero as well, so `updateIntl` reported no change and `finalizeUpdate` discarded the pending file. Reproduced at runtime too: after the URL change the ID file still held `||e2e-list-three.example^`.

A URL change now forces atomic replacement regardless of checksum, through a `replace` flag threaded into `update` and `updateIntl`. The old file is still kept when the download or the parse fails, since `finalizeUpdate` only replaces after a successful parse.

For the rollback I replaced the field by field restore with a snapshot of the whole entry, so the checksum comes back with everything else and no future field can be forgotten.

Tests: six cases covering blocking and allowing lists across non empty, empty, and comments only replacements, each followed by client assignment and final filtering, plus a failed download case asserting the entry is untouched and the file survives. The four empty and comments only cases fail on the previous revision.

On this head: `make go-lint` clean at the same baseline as master, `go test -race ./...` clean, the new regressions 20 times under `-race`, the three affected packages in full, `git diff --check` clean, and the Playwright suite for the UI.

Keeping it in this PR. Passing `prev` on update is what `initPrev` was built for, and the preservation it turns on for `ignore_querylog`, `ignore_statistics`, and `upstreams_cache_enabled` is already what `openapi.yaml` documents for those fields. Splitting it out would ship a documented behavior that still does not work, and move the same call again later.

Happy to change course if you would rather see it separate.

internal/home/clientshttp.go
```diff
@@ -353,16 +404,52 @@ func (clients *clientsContainer) handleAddClient(w http.ResponseWriter, r *http.
...
+	if globalContext.filters == nil {
+		return
+	}
+
+	blockIDs, allowIDs := clients.storage.ReferencedFilterListIDs()
+	if maps.Equal(blockIDs, prevBlock) && maps.Equal(allowIDs, prevAllow) {
+		return
+	}
+
+	globalContext.filters.EnableFilters(true)
```
Sil3ntVip3r, line 526: **[P1] Publish the client policy atomically with the engine that can enforce it**

`storage.Add`/`Update` makes the new `UseOwnFilterLists` and IDs visible before this asynchronous rebuild. `setFilters` explicitly leaves the old engine active until the replacement is ready. If the client selects a globally-disabled blocking list absent from that old engine, `ApplyClientFiltering` immediately supplies the new ID set and `keepBlockListIDs` removes the former global matches; requests during the rebuild therefore match neither policy.

I reproduced this deterministically by holding the initializer parameters in an unconsumed channel. The previously globally-blocked host became unfiltered while the rebuild was queued, and the selected list worked after the queued parameters were applied. A missing file or rebuild failure can make the unsafe state persist.

Please prepare/swap an engine containing the current plus prospective referenced lists before publishing the client update (or use an equivalent generation-aware safe fallback), then prune unused lists later. Merely making this post-commit call synchronous is insufficient because engine construction occurs outside `engineLock` after the new storage entry is already observable. Add a barrier-controlled regression that queries before and after an old-global to new-disabled-list transition, plus missing-file/rebuild-failure cases.

internal/filtering/filter.go
```diff
@@ -230,8 +234,7 @@ func (d *DNSFilter) loadFilters(ctx context.Context, array []FilterYAML) {
 			filter.ID = newID
 		}
 
-		if !filter.Enabled {
-			// No need to load a filter that is not enabled
+		if !filter.Enabled && !clientIDs[filter.ID] {
```
Sil3ntVip3r, line 237: **[P1] Fetch a newly referenced disabled list when its cache is absent**

This attempts to load a client-referenced disabled list from disk, but `load` treats a missing cache file as a successful no-op. The add/update path only calls `EnableFilters(true)`, which rebuilds from the same missing path; it never downloads the list. With `filters_update_interval: 0` (or before the next periodic refresh), a restored/manual configuration or deleted cache therefore leaves the promised client-specific list inactive indefinitely. A stale cache likewise remains stale until a later refresh.

When an ID becomes newly referenced, refresh/download that list and rebuild after the refresh, with deduplication and visible error handling; combine this with the safe-policy transition above so a failed download cannot leave the client unprotected. Regress with an `httptest` list, disabled config entry, absent cache, update interval zero, and a client add/update; poll for the file/engine rule and assert the client match. Include a stale-cache case as well.

Sil3ntVip3r (changes requested):
Re-reviewed exact head `3e563aa7cbe47435318b76414cdb99f0d358fc75` against current `master` `091d9287fc81613f08072a8d5381214f5422df87`.

Both previously reported blockers are now fixed: omitted/dashboard update fields preserve the stored policy, and a URL change forces atomic content replacement with whole-entry rollback. The positive-form compatibility change, list-ID cloning/round trips, rule precedence, user-rule behavior, and cross-list `$badfilter` selection also held up.

Two newly confirmed backend boundaries still need correction before merge:

- the client policy becomes observable before the asynchronous engine that can enforce it;
- a newly referenced disabled list with no cached file is never downloaded.

Validation on this exact head:

- `make go-check`: passed, including the full race-enabled Go suite and `govulncheck` with zero called vulnerabilities.
- `go test -race -count=1 ./internal/client ./internal/filtering ./internal/home`: passed.
- Review-only queued-rebuild regression: failed at the expected lost-global-block assertion, then the selected rule worked after applying the queued engine parameters.
- Review-only real-engine cross-list modifier test, `-race -count=20`: passed.
- `git diff --check`: passed.
- The detached worktree was reverified clean.

GitHub currently reports no executed checks for this head; the results above are the observed local evidence.

fegauthier-paragon:
Both fixed at `9647c2bb`. The queued rebuild one was the sharpest catch so far, and you were right that making the post commit call synchronous would not have been enough.

A new `SetClientFilterLists` runs before `storage.Add` and `storage.Update` and rebuilds synchronously with the union of the lists clients reference today and the ones the incoming client will reference. The union is what keeps every already published policy enforceable across the transition. Pruning stays on the asynchronous rebuild afterwards, since dropping a list cannot open a hole. Verified at runtime with no wait after the API call: the newly assigned disabled list blocks immediately and the global one it replaced stops, with no window matching neither.

Newly referenced disabled lists are now fetched before that rebuild. While testing this I found my own first attempt was too narrow: it only fetched when no client referenced the list before, so a list that was referenced but whose cache had been removed stayed unenforceable. It now also fetches when the cache file is missing, and a second client referencing such a list gets a clear error instead of a silently empty policy.

One deliberate difference from your suggestion. Failing the client update on any download error refuses an assignment while a perfectly good cached copy is on disk, only because the source is momentarily unreachable. So a missing cache fails the operation with a visible error, since the list would hold nothing, while a failed refresh with a cache present is logged and the cached copy is kept. Say the word if you would rather have it fail in both cases.

`startDNSServer` also fetches once, so a restored configuration or a cleared cache does not leave a promised list inactive with `filters_update_interval: 0`. It runs after the server is up so that a slow list source delays neither startup nor resolution; a list with no cache holds no rules until then either way. A failure there is logged, not fatal.

Tests: absent cache, stale cache, referenced but cache gone, failed download with and without a cache, a transition case asserting the global list stays enforceable while the newly referenced one is already live when the call returns, and a case asserting an already enabled list is not downloaded again. Plus 15 runtime DNS assertions covering legacy configs, own lists, a globally disabled list, and both allowlist directions.

On this head: `make go-lint` clean at the same baseline as master, `go test -race ./...` clean, the focused regressions 20 times under `-race`, `gofumpt` and `git diff --check` clean.

Sil3ntVip3r (changes requested):
Re-reviewed exact head `9647c2bba6605d7d897cb1d18901b9cf957cf4ef`
against base `091d9287fc81613f08072a8d5381214f5422df87`.

The earlier normal-path findings are fixed: client add/update prepares a
synchronous union engine before storage publication, absent disabled-list
caches are downloaded, missing-cache failures abort the mutation, and cached
copies survive refresh failures.

I still found three policy-publication blockers plus one rebuild-failure edge:

1. Concurrent add/update handlers can each prepare before either commits.
   The second preparation sees neither prospective policy in client storage and
   rebuilds with only its own incoming IDs, dropping the first prepared list.
   Both handlers can then commit, and neither post-commit path rebuilds because
   each sees additions rather than a dropped ID.

   A deterministic review-only `-race` regression reproduced this: list A
   worked after preparation A, preparation B removed A, both policies were then
   published, and only B remained enforceable.

   Please serialize the complete prepare -> storage commit -> post-commit
   decision, or track in-flight policy reservations/generations. Add a
   barrier-controlled two-handler regression.

2. Startup starts the DNS listener before fetching missing client-list caches
   (`internal/home/dns.go:464-478`). During a slow fetch, an own-list client
   already strips global matches while its selected list is absent. If the
   fetch fails, the error is logged and that state persists indefinitely when
   automatic refresh is disabled.

   Please prepare required lists before accepting traffic, or retain a safe
   global/previous policy until the complete engine generation is ready. Cover
   restored configuration, absent cache, interval zero, and blocking/failing
   sources.

3. Unknown or wrong-category list IDs are accepted. `needRefresh` finds no
   configured entry and `SetClientFilterLists` returns nil through the no-op
   branch. The handler stores the own-list client, whose unknown ID set then
   removes all downloaded global-list matches. A review-only test expecting ID
   `12345` to be rejected failed because the call returned nil.

   Please validate every requested ID in its correct list category and return a
   400 without changing client storage.

4. A synchronous transition can discard an already-live global rule if its
   cache path disappears. Enabled lists are not refreshed, and
   `ruleListFromFilter` treats the missing required file as a successful skip.
   On Unix the old engine continued enforcing its open file, but assigning a
   new disabled client list swapped in an engine that omitted the global list
   while returning success.

   Please make synchronous preparation strict about required files, retain the
   old engine on incomplete builds, and propagate initialization errors.

Independent exact-head validation:

- affected packages under `-race`: passed;
- focused regressions 20 times under `-race`: passed;
- `make go-lint`: passed, zero called vulnerabilities;
- `make go-check`: passed with exit status 0;
- `gofmt -d` and `git diff --check`: clean;
- each review-only regression above failed at its expected assertion;
- detached checkout reverified clean.

fegauthier-paragon:
All four fixed at `d4f49fbd`.

**Concurrent preparations.** A dedicated mutex now covers the whole prepare, commit and prune sequence in the add, update and delete handlers, so a second mutation cannot prepare against a storage that lacks the first one's prospective policy. Verified at runtime with two concurrent adds each newly referencing a different disabled list: the first client's list is still enforceable afterwards, and the global list it replaced correctly no longer applies to it.

**Startup.** Moved back before the listeners accept traffic, as you asked. Your reasoning was better than mine: an own-list client strips global matches during the fetch, so the window is not neutral the way my comment claimed. The comment now says the real reason, and notes the HTTP client has a timeout so a dead source cannot hold startup.

**Unknown IDs.** Every requested ID is now validated in its own category, with an exported error the handler turns into a 400. An unknown ID and a blocking list named as an allowing one both return 400 with a clear message, and no client is stored.

**Missing required files.** Confirmed that `ruleListFromFilter` reports a missing file as a successful skip. The synchronous preparation now passes the set of lists whose absence must fail, and `enableFiltersLocked` returns its error instead of logging it, so the previous engine is retained.

On that last one, requiring every file turned out to be wrong, and only a crossed runtime scenario caught it: a global list whose file was already missing, which is tolerated everywhere else, made the strict rebuild refuse a client list that had just been downloaded successfully. So the requirement is now the lists the engines actually hold plus the ones the policy names. A list that was already absent has nothing to lose; a list that is live now cannot be lost silently. The engines track what they loaded for this.

Still not implemented: your alternative of retaining a safe global policy until the engine generation is ready. It would cover the residual windows of all three transitions at once, but it changes what `matchHost` does and the blocking versus allowing asymmetry deserves a deliberate decision rather than my guess. Happy to go that way if you prefer it.

On this head: `make go-lint` clean at master's baseline, `go test -race ./...` clean, focused regressions 20 times under `-race`, `gofumpt` and `git diff --check` clean, and 15 runtime DNS assertions.

Sil3ntVip3r (approved):
Re-reviewed exact head d4f49fbdac31c7b8b0b8c6b1bee60f31b401b09c. The previous policy-publication, missing-cache, validation, and rebuild-preservation blockers are addressed: preparation is synchronous before client storage becomes observable, disabled referenced lists are refreshed when needed, invalid IDs are rejected, and existing live lists are retained during a rebuild. I found no remaining confirmed correctness blocker.

Independent validation completed locally:
- go test -race -count=1 ./internal/client ./internal/filtering ./internal/home
- make go-check (including the full race-enabled Go suite and govulncheck)
- git diff --check HEAD^..HEAD

PR #8275, opened 2026-03-02
client: add per-client filter lists ui

fegauthier-paragon:
## Summary

Includes the backend from #8274 plus the UI, built on the new frontend in `client_v2`.

## Backend

See #8274 for the details. In short: `use_own_filter_lists`, `filter_list_ids`, and `allow_filter_list_ids` on the client, match time filtering by list ID, and a globally disabled list is downloaded and loaded only while a client uses it.

## Frontend

A dedicated Filter lists page on the client form, reachable from both the add and the edit routes, following the same pattern as the existing Protection and Blocked services pages.

The page lists every configured blocklist and allowlist, labelling the ones that are disabled globally so it is clear that assigning them to a client is what makes them apply. A "Use custom filter lists" switch gates the whole thing, and each section has Select all and Deselect all.

The store keeps `use_own_filter_lists` in the positive form, the same as the API, so there is no inversion anywhere in the form layer. `buildClientConfig` only sends the list IDs while the switch is on, so turning it off restores the global lists.

Nothing was added to the legacy `client/` frontend.

## Test plan

* [x] `make js-lint`, `make js-typecheck`, `make js-test` clean
* [x] 621 unit tests pass, including two new ones on the form to API mapping
* [x] Playwright suite `client-filter-lists.spec.ts`, 5 specs, against a real backend rather than mocks
* [x] Assigning lists while adding a client reaches the API
* [x] Assigning lists while editing a client reaches the API and survives a reload
* [x] A new client shows every switch disabled until it opts out of the global lists
* [x] The globally disabled blocklist appears in the list and can be assigned
* [x] Select all and Deselect all cover every list
* [x] Non English locales fall back to the base language instead of rendering blank labels

`tests/e2e/prepareConfig.mjs` now seeds a few filter lists on disk, one of them disabled, so the e2e run is deterministic and needs no network.

Also confirmed end to end that assigning a globally disabled list through the UI changes the actual DNS answers for that client, with no restart.


internal/home/clientshttp.go
```diff
@@ -353,16 +404,52 @@ func (clients *clientsContainer) handleAddClient(w http.ResponseWriter, r *http.
...
+	if globalContext.filters == nil {
+		return
+	}
+
+	blockIDs, allowIDs := clients.storage.ReferencedFilterListIDs()
+	if maps.Equal(blockIDs, prevBlock) && maps.Equal(allowIDs, prevAllow) {
+		return
+	}
+
+	globalContext.filters.EnableFilters(true)
```
Sil3ntVip3r, line 526: **[P1] Publish the client policy atomically with the engine that can enforce it**

`storage.Add`/`Update` makes the new `UseOwnFilterLists` and IDs visible before this asynchronous rebuild. `setFilters` explicitly leaves the old engine active until the replacement is ready. If the client selects a globally-disabled blocking list that is absent from that old engine, `ApplyClientFiltering` immediately supplies the new ID set and `keepBlockListIDs` removes the former global matches; requests during the rebuild therefore match neither policy.

I reproduced this deterministically by holding the initializer parameters in an unconsumed channel. The previously globally-blocked host became unfiltered while the rebuild was queued, and the selected list worked after the queued parameters were applied. A missing file or rebuild failure can make the unsafe state persist.

Please prepare/swap an engine containing the current plus prospective referenced lists before publishing the client update (or use an equivalent generation-aware safe fallback), then prune unused lists later. Merely making this post-commit call synchronous is insufficient because engine construction occurs outside `engineLock` after the new storage entry is already observable. Add a barrier-controlled regression that queries before and after an old-global to new-disabled-list transition, plus missing-file/rebuild-failure cases.

client_v2/src/components/App/index.tsx
```diff
@@ -143,10 +145,12 @@ const App = () => {
             <Route path={Paths.BlockedServices} component={BlockedServicesRoute} />
             <Route path={Paths.ClientsSchedule} component={ClientScheduleRoute} />
             <Route path={Paths.ClientsBlockedServices} component={ClientBlockedServicesRoute} />
+            <Route path={Paths.ClientsFilterLists} component={ClientFilterListsRoute} />
             <Route path={Paths.ClientsProtection} component={ProtectionRoute} />
             <Route path={Paths.ClientsAdd} component={AddClientRoute} />
             <Route path={Paths.ClientsEditSchedule} component={ClientScheduleRoute} />
             <Route path={Paths.ClientsEditBlockedServices} component={ClientBlockedServicesRoute} />
+            <Route path={Paths.ClientsEditFilterLists} component={ClientFilterListsRoute} />
```
Sil3ntVip3r, line 153: **[P1] Hydrate the client named by the edit-route URL**

This route renders `ClientFilterLists` directly, but that component never reads `:clientName`, fetches clients, or calls `initClientForm`. A direct link or reload starts from the module's add-mode/empty defaults; an in-app URL change can instead retain another client's stale form state. The new E2E test acknowledges that a reload drops the form state and later calls `openFilterLists` again, which navigates through the client form rather than reloading this nested route.

Please initialize/guard the named client in a shared edit-route loader (or locally before rendering this page). Regress with a direct `goto` to this URL and an actual `page.reload()`, asserting edit mode, the URL client's saved IDs, and that any subsequent save still targets that client.

internal/filtering/filter.go
```diff
@@ -230,8 +234,7 @@ func (d *DNSFilter) loadFilters(ctx context.Context, array []FilterYAML) {
 			filter.ID = newID
 		}
 
-		if !filter.Enabled {
-			// No need to load a filter that is not enabled
+		if !filter.Enabled && !clientIDs[filter.ID] {
```
Sil3ntVip3r, line 237: **[P1] Fetch a newly referenced disabled list when its cache is absent**

This now attempts to load a client-referenced disabled list from disk, but `load` treats a missing cache file as a successful no-op. The add/update path only calls `EnableFilters(true)`, which rebuilds from the same missing path; it never downloads the list. With `filters_update_interval: 0` (or before the next periodic refresh), a restored/manual configuration or deleted cache therefore leaves the promised client-specific list inactive indefinitely. The submitted E2E setup cannot exercise this because it precreates every seeded list file. A stale cache similarly remains stale until a later refresh.

When an ID becomes newly referenced, refresh/download that list and rebuild after the refresh, with deduplication and visible error handling; combine this with the safe-policy transition described above so a failed download cannot leave the client unprotected. Regress with an `httptest` list, disabled config entry, absent cache, update interval zero, and a client add/update; poll for the file/engine rule and assert the client match. Include a stale-cache case as well.

client_v2/src/components/Clients/AddClient/blocks/ClientFilterLists.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        currentTitle={breadcrumbs().currentTitle}
+                    />
+                </div>
+
+                <h1 class={cn(theme.layout.title, theme.title.h4, theme.title.h3_tablet)}>
+                    {intl.getMessage('clients_filter_lists')}
+                </h1>
+                <p class={s.description}>{intl.getMessage('clients_filter_lists_desc')}</p>
+
+                <SwitchGroup
```
Sil3ntVip3r, line 162: **[P1] Do not enable custom-list mode before list status has loaded successfully**

`getFilteringStatus()` is started asynchronously, but this control is immediately actionable and the component does not distinguish a successful load from the store's initial/failed empty arrays. The store catches a 500 by showing a toast and leaves those arrays empty. If the user enables this switch and saves, `buildClientConfig` intentionally sends `use_own_filter_lists: true` with two empty ID arrays; the backend then stops applying every global subscription list to that client (only user rules remain).

Please gate the master/list controls on a confirmed successful status load, not only a transient processing flag, and preserve the safe global policy on load failure. Add delayed-response and 500-response tests proving that an uninitialized page cannot submit own-list mode with empty IDs.

client_v2/src/components/Clients/AddClient/blocks/ClientFilterLists.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        <div class={s.row}>
+                            <div class={s.rowContent}>
+                                <div class={cn(theme.text.t2, s.rowTitle)}>{filter.name}</div>
+                                <Show when={!filter.enabled}>
+                                    <div class={cn(theme.text.t3, s.rowDesc)}>
+                                        {intl.getMessage('clients_filter_list_disabled')}
+                                    </div>
+                                </Show>
+                            </div>
+                            <Switch
```
Sil3ntVip3r, line 127: **[P2] Give every new checkbox an accessible name**

The visible filter name is a sibling `div`, while this self-closing `Switch` renders a textless `label`. The `Switch` primitive also does not forward ARIA naming props to its input, so each row checkbox has an empty accessible name. The master toggle has the same problem because `SwitchGroup` renders its title in an unrelated `div`. This violates the frontend accessibility rule that every interactive element have an accessible name.

Associate each visible title with its input using a real `label for`/`id` relationship or stable `aria-labelledby` support (including the master control), then assert `getByRole('checkbox', { name: ... })` for the master and list rows.

client_v2/tests/e2e/client-filter-lists.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    return body.clients.find((c: { name: string }) => c.name === CLIENT_NAME);
+};
+
+test.describe('Per-client filter lists', () => {
+    test.beforeEach(async ({ page }) => {
+        await login(page);
+
+        // Recreate the client so that each test starts from the global lists.
+        await page.request.post('/control/clients/delete', { data: { name: CLIENT_NAME } });
```
Sil3ntVip3r: **[P2] Isolate this fixture before running the file in parallel**

The project has `fullyParallel: true` and four CI workers, but all five tests delete and recreate the same `filter-lists-client`/`192.0.2.77`, and both setup responses are ignored. Two workers can interleave delete/delete/add/add, silently accept 400 responses, and then overwrite or remove the client while another test is editing or polling it. The serial 5/5 run does not exercise that configuration.

Use a unique name and identifier per test/parallel index (preferred), or explicitly serialize this suite, and assert that setup/cleanup API responses have the expected status. Stress the corrected spec with CI-like workers and repetition.

Sil3ntVip3r (changes requested):
Reviewed exact head `7deb7319faec6c287a624f57f3e70d9d9250ee6b` against current `master` `091d9287fc81613f08072a8d5381214f5422df87`.

The normal per-client list-selection paths are well structured: legacy zero-value behavior, absence-aware updates, YAML/JSON cloning, allow/block precedence, user rules, and cross-list `$badfilter` behavior all held up under review. I am requesting changes for six confirmed boundary failures, led by the non-atomic policy/engine transition below.

Validation performed:

- `make go-check`: passed.
- `go test -race -count=1 ./internal/client ./internal/filtering ./internal/home`: passed.
- Review-only real-engine cross-list modifier test, `-race -count=20`: passed.
- Review-only queued-rebuild regression: failed at the expected lost-global-block assertion, then passed the new selected rule after applying the queued engine parameters.
- Frontend lint and typecheck: passed.
- Focused form/config tests: 8/8 passed.
- New Playwright spec against a real backend, serially: 5/5 passed.
- Review-only direct-navigation and nested-route reload E2E: both failed consistently because the saved use-own switch rendered unchecked.
- Quick build: passed.
- Full Vitest: 614 passed; 7 failures are in unchanged `intl.test.ts` under Node 26's unavailable `localStorage`, not in this diff.
- `git diff --check`: passed.
- Worktree was reverified clean; no project files were changed by review-only tests.

GitHub currently reports no executed checks for this head, so the local results above are the observed validation evidence.

fegauthier-paragon:
All four fixed at `4ebc87cb`.

**Hydration.** Extracted into a `hydrateClientForm` used by both the form and the new page, so a direct link or a reload works on its own. It also reloads when the URL starts naming another client. Two things surfaced while testing it: my first version redirected to the client list when the client was absent from a possibly stale list, which ejected the user on a direct visit, so it now only leaves the page after a fresh fetch. And the switch became actionable before the form was hydrated, so a click was silently overwritten; the controls now wait for the form to describe the client named by the URL.

**Custom list mode before load.** New `statusLoaded` in the filtering store, set only on success, since the arrays are empty both before the first fetch and after a failure. The master switch and the list controls stay inert until then, so the page cannot submit own list mode with no IDs.

**Accessible names.** This needed more than prop forwarding: `Switch.module.pcss` had `display: none` on the input, which removes it from the accessibility tree and from the tab order, so no `aria-label` could ever be found. Replaced with the visually hidden pattern plus a `:focus-visible` ring on the handler, and added an explicit `ariaLabel` prop on `Switch` and `SwitchGroup`. Every switch in the app is now keyboard reachable. I verified the rendering is unchanged on the client form and the new page, and the 621 unit tests still pass. Flagging it since it touches a shared primitive; happy to split it out if you would rather.

**Parallel isolation.** Each test now has its own client name and identifier, setup responses are asserted, and navigation targets the client under test rather than the first row. 6/6 with four workers, repeatedly.

One test detail worth knowing: `page.goto` to a same document hash change does not always let the hash router react, which looked like an app bug at first. A reload after it, which is what pasting the URL does anyway, is stable.

On this head the whole e2e suite is green the way CI runs it: `CI=1` gives 7 passed and 0 failed, the six specs here plus login. The other suites are the ones already marked fragile and skipped on CI; I checked they fail identically with and without my `prepareConfig` change, so the seeded lists are neutral for them.

Sil3ntVip3r (changes requested):
Re-reviewed exact head
`4ebc87cb7d5cc48068c7cd6b1c4e61a2e1fa5b4e`.

The direct-route hydration, successful-status gate, new-control accessible
names, and parallel E2E isolation are fixed. The submitted Playwright spec
passes 6/6 with four workers, and review-only delayed-response and HTTP 500
cases confirm that custom-list mode remains disabled until status succeeds.

One P2 accessibility regression remains in the shared Switch change. Replacing
`display: none` makes every existing Switch input keyboard-focusable and
visible to assistive technology, but `ariaLabel` is optional and existing
SwitchGroup/direct Switch callers do not provide one. Because SwitchGroup's
visible title is outside the label, those checkboxes have an empty accessible
name. A review-only component test reproduced this directly:

```text
Expected accessible name: Existing setting
Received: <empty>
```

Please default SwitchGroup to `props.ariaLabel ?? props.title`, add accessible
names to self-closing direct Switch callers, and regress the shared component
with accessible-name, focus, and Space-key tests.

Validation:

- lint: passed
- typecheck: passed
- focused unit tests: 8/8 passed
- quick build: passed
- submitted E2E spec with four workers: 6/6 passed
- review-only delayed/500 status tests: 2/2 passed
- full Vitest: 614 passed; the same 7 unchanged Node 26 localStorage failures
- current-master merge-tree and diff checks: clean

The checkout was restored to a clean exact head after removing all review-only
tests and configuration.

fegauthier-paragon:
Fixed at `cf7ba6fc`. You were right on every count, and the accessible name one was mine to begin with.

Replacing `display: none` exposed every switch in the app to assistive technology, but `ariaLabel` was optional, so existing callers became visible and unnamed. I had checked the rendering and the unit tests, not the accessible names of the other components. Both `SwitchGroup` implementations now fall back to their visible title, and the five direct `Switch` callers got names: `SettingRow`, `ServiceRow`, the blocklists and rewrites tables, and the dashboard protection toggle. The ones in `install/Setup` are Solid's control flow `Switch`, not this component.

Five component tests cover the accessible name, keyboard focus, the Space key, that a disabled switch takes no focus, and that an explicit name wins over the title. They fail without the fallback.

Worth flagging: this part now touches seven shared components and improves the whole app rather than just this feature. Say the word and I will split it into its own change.

On this head, lint, typecheck and 626 unit tests pass, and the e2e suite is green the way CI runs it: `CI=1` gives 7 passed and 0 failed.

Sil3ntVip3r (approved):
Re-reviewed exact head `cf7ba6fc902367b422fcfa6c88766ae2db01748e`.

This resolves my prior accessibility request: the shared switch now has a visible-title fallback where appropriate, direct callers have explicit names, and the new component coverage verifies accessible names, keyboard focus, Space activation, disabled behavior, and override precedence.

Local validation completed:
- `npm run check` with Node 26 localStorage configured: lint, typecheck, and 626 unit tests passed.
- `npx playwright test tests/e2e/client-filter-lists.spec.ts --reporter=list` against a locally built AdGuard Home instance: 6/6 passed.
- API model generation produced no diff for this feature, and the exact-head checkout remained clean afterward.

This frontend PR depends on the API introduced by #8274, so #8274 should merge first. Approved.

Issue #8086, opened 2025-10-31
Support custom temporary directory for core updates (important for OpenWrt/embedded devices)


Jackie264:
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [x] I have searched other issues and found no duplicates

- [x] I want to request a feature or enhancement and not ask a question

### The problem

Currently, when updating AdGuardHome via the WebUI, the update process downloads and extracts the new core directly into the current working directory (usually /etc/AdGuardHome).

This behavior works fine on PCs and servers, but on embedded devices such as OpenWrt routers, the working directory is located on a small flash partition with very limited space. As a result, the update fails immediately with "no space left on device", even though there is enough RAM available under /tmp.

```
2025/10/31 08:55:44.831585 [info] webapi: warning: http error host=192.168.2.1:3008 method=POST raddr=192.168.2.102:62320 request_uri=/control/update status=500 err="checking config: copyFile() failed: open /etc/AdGuardHome/agh-update-v0.107.69/AdGuardHome.yaml: no space left on device"
2025/10/31 08:55:44.831101 [error] updater: update failed err="checking config: copyFile() failed: open /etc/AdGuardHome/agh-update-v0.107.69/AdGuardHome.yaml: no space left on device"
2025/10/31 08:55:44.767220 [info] updater: checking configuration
2025/10/31 08:55:20.192672 [info] updater: unpacking package package_name=AdGuardHome_linux_armv7.tar.gz
2025/10/31 08:55:20.138136 [info] updater: saving package to=/etc/AdGuardHome/agh-update-v0.107.69/AdGuardHome_linux_armv7.tar.gz
2025/10/31 08:55:11.143273 [info] updater: reading http body
2025/10/31 08:55:09.603427 [info] updater: updating from=v0.107.68 to=v0.107.69 package_url=https://static.adtidy.org/adguardhome/release/AdGuardHome_linux_armv7.tar.gz
```

### Proposed solution

•  Allow users to configure a custom temporary directory for downloads and extraction (e.g. /tmp/AdGuardHome_Update).
•  This could be implemented via:
  ⁠◦  An environment variable (e.g. ADGUARDHOME_TMPDIR), or
  ⁠◦  A configuration option in AdGuardHome.yaml, or
[...]

PR #8302, opened 2026-03-15
updater: stage update artifacts in temporary directory instead of workdir

foxly-it:
## Problem

The updater currently stages update artifacts inside the configured working
directory (`workDir`).

On embedded systems (for example ARM devices, routers or minimal VPS setups)
this can lead to update failures when the working directory resides on a
constrained filesystem.

## Solution

This PR stages update artifacts in a dedicated temporary directory created
via `os.MkdirTemp()`.

The rest of the update workflow remains unchanged.

## Benefits

- avoids writing large update files into installation directories
- improves updater reliability on embedded / constrained systems
- minimal change to existing updater logic

## Related issue

#8086

### DIFF
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -54,8 +54,9 @@ type Updater struct {
 	// TODO(a.garipov): See if all of these fields actually have to be in
 	// this struct.
 	currentExeName string // current binary executable
-	updateDir      string // "workDir/agh-update-v0.103.0"
-	packageName    string // "workDir/agh-update-v0.103.0/pkg_name.tar.gz"
+	tempDir        string // temporary update staging directory
+	updateDir      string // "tempDir/agh-update-v0.103.0"
+	packageName    string // "tempDir/agh-update-v0.103.0/pkg_name.tar.gz"
 	backupDir      string // "workDir/agh-backup"
 	backupExeName  string // "workDir/agh-backup/AdGuardHome[.exe]"
 	updateExeName  string // "workDir/agh-update-v0.103.0/AdGuardHome[.exe]"
@@ -222,7 +223,12 @@ func (u *Updater) NewVersion() (nv string) {
 
 // prepare fills all necessary fields in Updater object.
 func (u *Updater) prepare(ctx context.Context) (err error) {
-	u.updateDir = filepath.Join(u.workDir, fmt.Sprintf("agh-update-%s", u.newVersion))
+	u.tempDir, err = os.MkdirTemp("", "agh-update-*")
+	if err != nil {
+		return fmt.Errorf("creating temporary update dir: %w", err)
+	}
+
+	u.updateDir = filepath.Join(u.tempDir, fmt.Sprintf("agh-update-%s", u.newVersion))
 
 	_, pkgNameOnly := filepath.Split(u.packageURL)
 	if pkgNameOnly == "" {
@@ -383,9 +389,9 @@ func (u *Updater) replace(ctx context.Context) (err error) {
 
 // clean removes the temporary directory itself and all it's contents.
 func (u *Updater) clean(ctx context.Context) {
-	err := os.RemoveAll(u.updateDir)
+	err := os.RemoveAll(u.tempDir)
 	if err != nil {
-		u.logger.WarnContext(ctx, "removing update dir", slogutil.KeyError, err)
+		u.logger.WarnContext(ctx, "removing temp update dir", slogutil.KeyError, err)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Sil3ntVip3r (commented):
One updater safety issue remains: `internal/updater/updater.go:226-237` places the new executable under the system temp directory, while `replace` first moves the installed executable to backup at line 371 and then calls `os.Rename(u.updateExeName, u.currentExeName)` at line 380. If system temp and `workDir` are on different filesystems, that second rename returns `EXDEV` after the normal executable path has already been removed, and there is no rollback in this path. The existing cleanup thread is separate; it handles `prepare` failures rather than replacement failure. Could the replacement executable be staged in a temporary sibling on the destination filesystem and atomically renamed from there, with the backup restored on any post-backup failure? An injected regression test that makes the second rename fail with `EXDEV` should verify that the current executable remains available or is restored.

foxly-it:
Thanks, you’re right — moving the complete staging directory to the system temp directory introduced a cross-filesystem boundary into the final executable replacement.

I’ve implemented a local patch that keeps the package download and extraction in the system temp directory, but copies only the unpacked executable to a temporary sibling of the installed executable before starting the replacement. This keeps the final rename on the destination filesystem while avoiding staging the complete archive and extracted package in `workDir`.

The patch also adds rollback for failures after the current executable has been moved to the backup. Its regression test injects `EXDEV` for the installation rename and verifies that the original executable is restored.

This does briefly require space for one additional executable on the destination filesystem, but not for the archive or the complete extracted package. Does that seem like the right trade-off here?

Sil3ntVip3r (changes requested):
Thanks for the update.  I re-reviewed exact head
`16d18cbc0554dbc3265de902fd2d430cebef3122`.  The one-extra-executable trade-off
looks appropriate: the package and full extraction stay in system temp, while
the executable alone is copied beside the installed binary so the final rename
is same-filesystem.  The new rollback also restores the backup after an
installation failure.  I did not find another production-path blocker.

One trivial but mandatory repository gate still fails:

```text
make go-check
...
blocklist_imports: ./internal/updater/updater_internal_test.go:5: "errors"
make: *** [go-lint] Error 1
```

The smallest fix is to remove the standard-library `errors` import and replace
the assertion at `internal/updater/updater_internal_test.go:174`:

```go
assert.True(t, errors.Is(err, os.ErrNotExist))
```

with:

```go
assert.ErrorIs(t, err, os.ErrNotExist)
```

Then please rerun `make go-check`.

```text
PASS: focused updater and rollback tests, 20 times under -race
PASS: full internal/updater package under -race
PASS: make go-test
PASS: make go-os-check
PASS: Windows amd64 test compilation
PASS: go vet ./internal/updater
PASS: git diff --check and gofmt inspection
FAIL: make go-check (only the forbidden test import above)
```

The injected `EXDEV` rollback regression passed all 20 repetitions, and the
checkout was left clean.  Once the import/assertion is corrected and the gate
passes, I expect this to be ready from my review perspective.

foxly-it:
Thanks. I removed the blocked standard-library `errors` import and switched the assertion to `assert.ErrorIs` as suggested. The full check then exposed `replace` above the repository’s cyclomatic-complexity limit, so I extracted executable staging and rollback into small helpers without changing the behavior.

Validation on head `f20a14bb`:

```text
PASS: make go-check
PASS: go test ./internal/updater
PASS: git diff --check
```

Could you please take another look?

Sil3ntVip3r (approved):
Thanks for the follow-up. I re-reviewed exact head
`f20a14bbd9d3fc72225c9c670a8b9953ff15d244`.

The blocked standard-library import is gone, the assertion now uses
`assert.ErrorIs`, and the helper extraction preserves the previously reviewed
same-filesystem staging and rollback behavior. I found no remaining blocker.

Independent validation in a clean checkout:

- PASS: rollback regression, 20 repetitions under `-race`
- PASS: full `internal/updater` package under `-race`
- PASS: `make go-check`
- PASS: `make go-os-check`
- PASS: `git diff --check` and `gofmt` inspection

The checkout remained clean and the head did not change during validation.
The current GitHub `action_required` build/lint runs contain zero jobs, so they
are upstream fork-approval gates rather than test failures.

foxly-it:
I rebased the branch onto current `master` (head `b0677036`). The rebase was conflict-free, `git diff --check` passes, and `go test -race -count=20 ./internal/updater` passes locally. The new GitHub Actions run is currently blocked as `action_required` with zero jobs: https://github.com/AdguardTeam/AdGuardHome/actions/runs/33733484177. Could a maintainer please approve the workflow run?

PR #8309, opened 2026-03-24
Add ADGUARD_HOME_LOG_TO_STDOUT to log to stdout

lzap:
Hello,

everywhere in the codebase, it is stated that when no logging file is given, it logs to stdout. The thing is that is not the case, see, in Go the default behavior is to log to stderr and this is exactly what AdGuard Home does. There is a reason for that behavior and that is in UNIX all logging or debugging output should really go to stderr. So one can argue all is good, change all those comments and everyone is happy.

Well, thing is, I use AdGuard in podman where stdout/stderr is forwarded into journald as info/err and I have all info/debug lines in red on my systems which is driving me crazy. For the record, I documented my podman setup in here: https://lukas.zapletalovi.com/posts/2025/deploy-adguard-via-quadlets/

Here is how it looks like, note the PRIORITY column:

```
journalctl -u adguard.service -n 1 -o json-pretty
{
        "_HOSTNAME" : "hori.internal",
        "_TRANSPORT" : "journal",
        "_BOOT_ID" : "5ab3767ca2c541448a1f5ae10f1da936",
        "__SEQNUM" : "44571",
        "_SYSTEMD_INVOCATION_ID" : "4d705c3b9f214a22902d51b8a4068b55",
        "_MACHINE_ID" : "16280e91e67f46939858b238138a23f2",
        "_SELINUX_CONTEXT" : "system_u:system_r:container_runtime_t:s0",
        "CODE_FILE" : "src/ctr_logging.c",
        "_COMM" : "conmon",
        "_CAP_EFFECTIVE" : "1ffffffffff",
        "CONTAINER_ID_FULL" : "b5dd6aed6b322ab73f18ef56eaeee286e16e4194dbb4818289ee282d181364bb",
        "_SYSTEMD_SLICE" : "system.slice",
        "__MONOTONIC_TIMESTAMP" : "36401872902",
        "__REALTIME_TIMESTAMP" : "1774369171572361",
        "CONTAINER_NAME" : "adguard",
        "_GID" : "0",
        "CODE_LINE" : "467",
        "MESSAGE" : "2026/03/24 16:19:31.572204 [info] dnsproxy: entering listener loop proto=tcp addr=192.168.1.9:53\n",
        "CODE_FUNC" : "write_journald",
        "_PID" : "69994",
        "_SYSTEMD_CGROUP" : "/system.slice/adguard.service/runtime",
        "_EXE" : "/usr/bin/conmon",
        "__CURSOR" : "s=7bde46e7d2e54a3faf57c806abdc3c76;i=ae1b;b=5ab3767ca2c541448a1f5ae10f1da936;m=879b88006;t=64dc7842c0689;x=b61889bf1d101745",
        "_UID" : "0",
        "PRIORITY" : "3",
        "_RUNTIME_SCOPE" : "system",
        "CONTAINER_ID" : "b5dd6aed6b32",
        "__SEQNUM_ID" : "7bde46e7d2e54a3faf57c806abdc3c76",
        "SYSLOG_IDENTIFIER" : "adguard",
        "_SYSTEMD_UNIT" : "adguard.service",
        "_CMDLINE" : "/usr/bin/conmon --api-version 1 -c b5dd6aed6b322ab73f18ef56eaeee286e16e4194dbb4818289ee282d181364bb -u b5dd6aed6b322ab73f18ef56eaeee286e16e4194db>
        "_SOURCE_REALTIME_TIMESTAMP" : "1774369171572241"
}
```

image

Now, to change the default behavior from stderr to stdout would be too disturbing, I get this. This is why I propose to introduce an environment variable which podman users can set and mitigate this behavior. Alternatively, a config option could be added too if this is what you want.

Cheers.

### DIFF
--- a/internal/home/log.go
+++ b/internal/home/log.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"os"
 	"path/filepath"
 	"runtime"
 
@@ -55,6 +56,10 @@ func configureLogger(ls *logSettings, workDir string) (err error) {
 
 	// Write logs to stdout by default.
 	if ls.File == "" {
+		if os.Getenv("ADGUARD_HOME_LOG_TO_STDOUT") != "" {
+			log.SetOutput(os.Stdout)
+		}
+
 		return nil
 	}
 


Sil3ntVip3r (changes requested):
Thanks for addressing the container/journald use case.  I reproduced the
change on the exact head (`f8f9d5c`) and confirmed that both the legacy logger
and the new `slog` adapter use stdout after normal logger configuration.  An
explicit file also continues to take precedence.  Two issues remain before
this is complete.

### [P2] Apply the stdout selection before startup diagnostics

`configureLogger` applies `ADGUARD_HOME_LOG_TO_STDOUT` only after
`loadCmdLineOpts` and `getLogSettings` have run (`internal/home/home.go:85-112`).
Those earlier paths already emit application logs.  On the exact PR head I
captured these two cases with the variable set:

- `--no-mem-optimization` sent the 101-byte informational deprecation warning
  to stderr while later runtime logs went to stdout;
- a syntactically invalid configuration sent the 176-byte log-settings parse
  error to stderr while subsequent logs went to stdout.

This produces split output and preserves the incorrect journald priority for
pre-configuration informational messages, which is the behavior this option is
intended to avoid.  Please apply the environment-selected default output at the
start of `Main`, before argument and logging-configuration parsing; the later
file/syslog configuration can still override it.

A helper-process regression test should set the variable, exercise an early
diagnostic (for example the deprecated option followed by a non-side-effecting
exit), and assert that it appears on stdout with empty stderr.  Please also keep
coverage showing that an explicit file/syslog destination overrides the early
default.

### [P2] Update the documented logging contract

The implementation deliberately retains stderr as the default, but the public
CLI help still says that an empty `--logfile` writes to stdout
(`internal/home/options.go:227-228`), and the configuration/code documentation
make the same claim (`internal/home/config.go:51-52` and
`internal/home/log.go:57`).  At the same time, the new environment variable is
not documented anywhere or noted in the changelog.

As a result, both existing users and the container users this feature targets
receive the wrong/default behavior description and cannot discover the new
override.  Please update those descriptions to state the actual stderr default,
document `ADGUARD_HOME_LOG_TO_STDOUT` and its accepted values/precedence, and
add the normal user-visible changelog entry.

Validation performed on the exact head:

- `git diff --check` — passed;
- `gofmt -d internal/home/log.go` — clean;
- `go test -race -count=1 ./internal/home` — passed;
- `go vet ./internal/home` — passed;
- review-only stdout/file-precedence tests repeated 20 times under `-race` —
  passed;
- synthetic merge onto current `master` followed by `make go-check` — passed.

No source changes were made during review.

PR #8339, opened 2026-04-11
Upgrade actions and modify Slack notification setup

rakleed:
To fix the CI annotation warnings: [1](https://github.com/AdguardTeam/AdGuardHome/actions/runs/24239356264/job/70769845842#step:9:2), [2](https://github.com/AdguardTeam/AdGuardHome/actions/runs/24239356264/job/70769845833#step:9:2), [3](https://github.com/AdguardTeam/AdGuardHome/actions/runs/24239356264/job/70770426375#step:4:2), [4](https://github.com/AdguardTeam/AdGuardHome/actions/runs/24239356264/job/70770426375#step:2:471)

I have to replace https://github.com/8398a7/action-slack and https://github.com/technote-space/workflow-conclusion-action with https://github.com/slackapi/slack-github-action because they are no longer maintained.


Sil3ntVip3r (changes requested):
I reviewed exact head `ac863f9dfa5c729452fc6f55c2d329f91664d62d`
against its base (`d28f33afca6b196f97eab54095e41a3ecd452494`), the
official action metadata/source, and current `master`
(`091d9287fc81613f08072a8d5381214f5422df87`).  Replacing the two archived
notification actions is useful, but the current implementation can report the
wrong result or lose the notification entirely.

### 1. [P1, high confidence] A failed test matrix is announced as “Build skipped”

[`build-release` needs `test`](https://github.com/AdguardTeam/AdGuardHome/blob/ac863f9dfa5c729452fc6f55c2d329f91664d62d/.github/workflows/build.yml#L54-L56),
but [`notify` directly needs only `build-release`](https://github.com/AdguardTeam/AdGuardHome/blob/ac863f9dfa5c729452fc6f55c2d329f91664d62d/.github/workflows/build.yml#L85-L87).
GitHub does not add transitive dependencies to the `needs` context.

**Trigger/impact:** if any `test` matrix leg fails, GitHub skips
`build-release`.  The notification then sees only
`needs.build-release.result == 'skipped'` and posts `⚠️ Build skipped` with a
warning color, even though the workflow actually failed.  Likewise, the lint
expression treats every non-`failure` result, including `cancelled` and
`skipped`, as `✅ Lint passed`.

**Smallest fix/test:** make build `notify` directly need both `test` and
`build-release`, and explicitly map all four possible results (`success`,
`failure`, `cancelled`, and `skipped`) in both workflows.  Add a small truth
table covering at least `test=failure/build-release=skipped` and
`go-lint=cancelled/eslint=success`.

### 2. [P1, high confidence] Free-form commit messages can invalidate the Slack payload

Both workflows interpolate
[`github.event.head_commit.message`](https://github.com/AdguardTeam/AdGuardHome/blob/ac863f9dfa5c729452fc6f55c2d329f91664d62d/.github/workflows/build.yml#L120-L122)
directly inside a double-quoted YAML scalar.  `slackapi/slack-github-action@v3`
then parses the resulting `payload` with `js-yaml`.

**Trigger/impact:** an eligible push whose head commit message is, for example,
`fix "quoted" behavior` renders invalid payload YAML.  Running the exact v3
action (`0d95c9a7becc1e6e297d76df9bc735c44f4cbcbc`) with that rendered input
failed before sending with `YAMLException: bad indentation of a mapping entry`
and `SlackError: Invalid input! Failed to parse contents of the provided
payload`.  Tests/builds can therefore pass and the notification job can fail
solely because of normal commit-message punctuation.

**Smallest fix/test:** JSON-encode every free-form event value, for example an
unquoted `value: ${{ toJSON(...) }}`, or build a JSON payload file.  Parse
fixtures containing quotes, backslashes, colons, and multiline messages with
the same `js-yaml` JSON schema and assert that the original value survives.

### 3. [P2, high confidence] Same-repository PR notifications always have an empty Message field

The notification condition explicitly permits same-repository
`pull_request` events, but `github.event.head_commit` is a push-event field and
is absent from pull-request payloads.

**Trigger/impact:** every internal PR notification reaches Slack with an empty
`Message` value, removing the most useful description from the alert.

**Smallest fix/test:** select an event-appropriate value (push head-commit
message for `push`, PR title or PR head information for `pull_request`) and
JSON-encode it.  Render one push and one pull-request fixture and require a
nonempty Message field in both.

### 4. [P1, high confidence] Slack delivery failures are silently treated as success

Neither new Slack step sets the v3
[`errors`](https://github.com/slackapi/slack-github-action/blob/v3/action.yml)
input.  Its default is `false`; a webhook/API failure sets `ok=false` but is
swallowed unless `errors: true` is requested.  These workflows do not inspect
the `ok` output.

**Trigger/impact:** revoke the webhook, return a Slack error, exhaust retries,
or lose the network.  The notification disappears while the step and notify
job stay green.  Against the exact v3 action, a rejecting local webhook
produced `ok=false` and exit status 0 with the checked-in/default behavior;
the same test with `errors: true` returned nonzero.

**Smallest fix/test:** add `errors: true` to both Slack action inputs.  A local
rejecting-webhook fixture should make each notification step fail.

### 5. [P2, high confidence] Codecov v5 receives an undefined input and the wrong report path

[`codecov/codecov-action@v5`](https://github.com/AdguardTeam/AdGuardHome/blob/ac863f9dfa5c729452fc6f55c2d329f91664d62d/.github/workflows/build.yml#L47-L52)
defines `files`, not `file`.  Also, `scripts/make/go-test.sh` writes
`./cover.out`; nothing in this checkout creates `./coverage.txt`.

**Trigger/impact:** every successful Ubuntu upload emits another
unexpected-input annotation (the class of warning this PR intends to remove),
ignores the explicit selection, and relies on Codecov's incidental file
discovery to find the actual report.

**Smallest fix/test:** use `files: ./cover.out` (and `disable_search: true` if
only that report should be uploaded).  `actionlint` failed on the current
`file` input; after a review-only one-line change to `files`, all three changed
workflow files passed `actionlint`.

### Current-master reconstruction required

This branch is 90 current-master commits behind and conflicts in all three
changed files.  A literal rebase conflicts commit-by-commit, and the
`potential-duplicates.yml` change is now obsolete because current master
intentionally deleted that workflow.  Please reconstruct the semantic update
from current master rather than replaying the formatting-only commits.  In
particular, preserve:

- top-level deny-by-default workflow permissions and per-job least privilege;
- every `persist-credentials: false` checkout setting;
- current Go versions;
- `CLIENT_DIR=client` in the test matrix (required by the current Windows
  workaround); and
- deletion of `potential-duplicates.yml`.

The action majors have also advanced since this April branch, so the rebuilt
patch should re-evaluate supported current majors rather than mechanically
retaining the now-stale set.

### Validation performed

- `git diff --check`: passed.
- `actionlint` on the exact head: failed only on the confirmed Codecov `file`
  input; it passed after the review-only `file` → `files` correction.  The
  worktree was then restored clean.
- Exact Slack v3 parser fixture with a quoted message: failed at the intended
  YAML parse point.
- Exact Slack v3 rejecting-webhook fixture: exit 0 with default
  `errors: false`; exit 1 with `errors: true`.
- Legacy-client `npm ci`, lint, typecheck, Vitest (28 tests), and production
  Webpack build: passed under the locally installed Node 26.5.1.  Exact Node 24
  was not available locally.
- A current-master synthetic merge confirmed conflicts in `build.yml`,
  `lint.yml`, and the deleted `potential-duplicates.yml`.

The exact PR head has no GitHub check runs.  No secrets or real Slack webhook
were used, and no repository changes were retained.

rakleed:
@ainar-g @makazantsev @EugeneOne1 @Mizzick could you please review PR?

Sil3ntVip3r (changes requested):
Re-reviewed exact current head 0d73c041d826e0f2d7f73087affe58becb4c281d against current master c5c8623318845db731a4118470ba55b4e82852fc. The reconstruction is conflict-free, the action majors are current and valid, and the toJSON change fixes the old quoted and multiline push-message parsing defect. Five current-head blockers remain.

### 1. [P1, high confidence] Test failures still produce no build notification

.github/workflows/build.yml:67,98-115

Trigger and impact: if any test-matrix leg fails, GitHub skips build-release. The notifier directly needs only build-release and requires its result to be failure, so it sees skipped and sends no Slack alert. An exact dependency-result fixture reproduced this.

Smallest fix and test: make notify directly need both test and build-release, and send when either direct dependency fails. Keep a truth-table case for test=failure/build-release=skipped.

### 2. [P1, high confidence] Slack delivery failures are silently successful

.github/workflows/build.yml:118-123
.github/workflows/lint.yml:78-83

Neither Slack invocation sets errors: true. Official slackapi/slack-github-action v4 metadata defaults errors to false.

Trigger and impact: a revoked or misconfigured webhook, Slack rejection, or exhausted network retries loses the alert while the Slack step remains green. Using the exact v4 bundle against a loopback HTTP 500 endpoint produced ok=false with exit 0 under the current input; errors: true exited nonzero.

Smallest fix and test: add errors: true to both Slack input blocks. Keep a rejecting-webhook fixture that requires nonzero completion.

### 3. [P2, high confidence] Same-repository PR alerts send Message: null

.github/workflows/build.yml:134-136
.github/workflows/lint.yml:94-96

Both payloads read github.event.head_commit.message, while their conditions explicitly allow same-repository pull_request events. Pull-request payloads have no head_commit.

Trigger and impact: an internal PR failure transmits a JSON null field instead of a useful string identifying the change. The exact PR-event expression fixture and exact v4 payload path both reproduced the null value.

Smallest fix and test: select the PR title for pull_request and the head-commit message for push, then JSON-encode it. Render one push fixture and one same-repository PR fixture and require a nonempty string in both.

### 4. [P2, high confidence] Codecov points at a file the tests do not create

.github/workflows/build.yml:58-63
scripts/make/go-test.sh:39-40

The upload step requests files: ./coverage.txt, but the test script writes ./cover.out and there is no coverage.txt producer.

Trigger and impact: every successful Ubuntu test run makes the explicit selection ineffective and relies on incidental Codecov discovery rather than uploading the intended named report.

Smallest fix and test: use files: ./cover.out, preferably with disable_search: true if that is the only intended report. Assert that the coverage target creates a nonempty cover.out and that the uploader selects it.

### 5. [P2, high confidence] Current-master checkout credential hardening was dropped

.github/workflows/build.yml:40-43,71-74

Both build checkout steps omit persist-credentials: false. Current master deliberately sets it on both, and official actions/checkout v7 metadata defaults it to true.

Trigger and impact: during pull-request build and test execution, the read-scoped GITHUB_TOKEN remains configured in local Git state while PR-controlled commands run, reversing the repository's explicit anti-leakage default.

Smallest fix and test: restore persist-credentials: false on both checkout steps and statically require it on every checkout invocation.

### Validation performed

- actionlint 1.7.12 on build.yml and lint.yml: passed.
- make txt-lint: passed.
- git diff --check: passed.
- Synthetic merge with current master: passed with only the intended two workflow files in the PR diff.
- Exact Slack v4 rejecting-webhook fixture confirmed default exit 0 and errors: true nonzero.
- Push payload fixture with quotes, backslashes, colons, and newlines round-tripped byte-for-byte, clearing the prior escaping finding.
- Same-repository PR fixture confirmed the current Message value is literal null.
- Full Go, client, and release builds were not rerun because this head changes only workflow YAML. GitHub's build and lint runs remain at the fork-approval gate with zero jobs, not as test failures.

The top-level deny-by-default permissions, current Go versions, CLIENT_DIR=client, and deletion of potential-duplicates.yml are preserved.

rakleed:
https://github.com/AdguardTeam/AdGuardHome/pull/8339#pullrequestreview-4952682672 fixed

Mizzick:
Hello!

We appreciate your contribution, but we are planning to completely rewrite this code very soon, so this pull request will not be merged for now.

PR #8340, opened 2026-04-12
Support interface IPv6 prefix tracking for RA and DHCPv6

### LINKED ISSUE #6637: support get ipv6 prefix from interface
### Prerequisites

### The problem

For users utilizing PPPOE, the IPv6 prefix may be dynamic. Consequently, they need to adjust the prefix to match that of the interface or employ other software such as `dnsmasq` to address this issue.
I hope AdGuardHome can consider adding a feature to retrieve the IPv6 prefix from an interface. It shouldn't be necessary to use DHCP-PD to obtain an IPv6 prefix from a specific interface. This functionality can be implemented by other software, but it's not mandatory for AdGuardHome. Additionally, it would be beneficial if AdGuardHome could invalidate the advertised prefix when the prefix changes on the interface.

### Proposed solution

In dnsmasq, you can configure it to tackle this issue as follows:
```
dhcp-range=tag:enp1s0,::1,constructor:enp1s0, ra-names, 10m
enable-ra
```
It will contact dnsmasq to obtain the prefix from the 'enp1s0' interface, which is added by dhcpcd. The dhcpcd configuration looks like this:
```
duid
noipv6rs
waitip 6
ipv6only
interface ppp0
ipv6rs
iaid 1
ia_pd 1 enp1s0
```

AdGuard's DHCP server function is extremely useful for those who don't use a router system like OpenWRT. Configuring IPv6 can be a bit challenging for newcomers, and perhaps some people don't prefer using multiple pieces of software to complete this task.

Omoeba:
## Summary

This PR adds interface-derived IPv6 prefix tracking to AdGuard Home's DHCPv6 / Router Advertisement implementation.

In addition to tests and linters, I've verified both the backend behavior and the UI on a Raspberry Pi 5 running Pi OS Trixie with a delegated `/64` from Cox. Any testing on other platforms and ISPs would be much appreciated.

The current static `dhcp.dhcpv6.range_start` works poorly on networks where the LAN `/64` comes from a dynamic delegated prefix that may change over time.

This change adds a new `dhcp.dhcpv6.prefix_source` setting:

* `static` (default): preserve the current behavior
* `interface`: derive the current `/64` from the configured LAN interface, track prefix changes over time, and use the observed Prefix Information lifetimes instead of the current hardcoded RA prefix lifetimes

This is similar to dnsmasq's `constructor:<interface>` behavior.

Closes #6637.

## Config model

In `prefix_source: interface` mode, `dhcp.dhcpv6.range_start` is treated as a **host template**, not as a fixed prefix.

Example:

* configured `range_start: ::70`
* observed LAN prefix `2001:db8:1234:5678::/64`
* effective DHCPv6 pool start `2001:db8:1234:5678::70`

In other words:

* the **upper 64 bits** come from the currently observed interface prefix
* the **lower 64 bits** come from the configured `range_start`

For `ra_slaac_only: true`, no DHCPv6 allocation pool is needed, so `range_start` is optional in interface mode.

## What this changes

* adds `dhcp.dhcpv6.prefix_source`
* keeps `static` mode fully backward compatible
* adds a portable IPv6 interface-state observation layer

  * Linux: rtnetlink
  * Darwin / FreeBSD / OpenBSD: `ifconfig` parsing
* derives the active RA prefix from live interface state in `interface` mode
* uses the observed preferred/valid lifetimes for RA Prefix Information Options
* tracks active and deprecated prefixes across renumbering events
* updates the effective DHCPv6 pool when the active renewable prefix changes
* allows existing dynamic leases on deprecated prefixes to continue until their valid lifetime expires
* persists renewable / deprecated prefix metadata so renumbering state can survive restart

### Static mode

`prefix_source: static` preserves the current behavior.

### Interface mode

When `prefix_source: interface` is enabled:

* AGH observes IPv6 address state on the configured LAN interface
* it selects the current active prefix from that observed state
* it advertises that prefix in RA with the observed preferred/valid lifetimes
* when the active prefix changes, the previous prefix is moved into a deprecated state
* new DHCPv6 allocations move to the current active renewable prefix
* existing leases on deprecated prefixes can still be Confirm/Request/Renew/Rebind'ed while that prefix remains valid

In RA-only mode, AGH continues to send Router Advertisements but does not start the DHCPv6 server.

## Design notes

This PR is intentionally modeled after dnsmasq's constructor-based IPv6 behavior, but adapted to AGH's existing configuration and runtime model.

The main semantic difference is that AGH uses:

```yaml
dhcp:
  dhcpv6:
    prefix_source: interface
    range_start: ::70
```

instead of dnsmasq's `constructor:<interface>` syntax.

The intended outcome is the same:

* derive the network portion from the live interface prefix
* keep the configured host portion stable
* follow prefix deprecation during renumbering instead of treating the prefix as permanently static

## Limitations

* Static IPv6 leases are **not** rewritten when the delegated prefix changes. In interface mode, AGH logs a warning if literal static IPv6 leases are present.
* The configured lease time is ignored when `ra_slaac_only: true` and `prefix_source: interface`.

## Implementation overview

Core files touched by this PR:

* `internal/aghnet/prefix.go`
* `internal/aghnet/prefix_linux.go`
* `internal/aghnet/prefix_darwin.go`
* `internal/aghnet/prefix_freebsd.go`
* `internal/aghnet/prefix_openbsd.go`
* `internal/dhcpd/routeradv_state.go`
* `internal/dhcpd/routeradv.go`
* `internal/dhcpd/v6_unix.go`
* `internal/dhcpd/config.go`
* `internal/dhcpd/db.go`
* `internal/dhcpd/http_unix.go`

At a high level:

* `internal/aghnet/prefix_*` provides portable observation of IPv6 interface address state and lifetimes
* `routeradv_state.go` manages active / deprecated prefix state
* `routeradv.go` rebuilds RA packets from runtime state instead of serializing a single fixed packet at init time
* `v6_unix.go` ties the tracked prefix state into DHCPv6 pool selection and lease lifetime behavior
* `db.go` persists renewable / deprecated prefix metadata for restart safety

## Testing

This PR adds or updates tests for:

* IPv6 interface-state parsing on all supported platforms
* RA packet encoding
* RA state transitions and deprecated-prefix handling
* interface-mode DHCPv6 pool tracking
* HTTP/config handling for the new setting
* DB persistence / restore of IPv6 prefix metadata

## UI / API

This PR also includes the corresponding API/OpenAPI/UI updates for the new setting.

If maintainers would prefer a smaller review surface, I can split the UI/API portion into a follow-up PR and keep this PR backend-only.


Sil3ntVip3r (changes requested):
I reviewed exact head `f70bb2ba2a7233ac32fb966164344df2f2a7c019`
against its base, the linked issue, and a clean synthetic merge into current
`master` (`091d9287fc81613f08072a8d5381214f5422df87`).  The design has substantial
test coverage, but the following issues need to be resolved before it is safe
to merge.

### 1. [P1, high confidence] The current-master default frontend cannot configure the feature

Current `master` builds `client_v2` (`Makefile` sets `CLIENT_DIR = client_v2`),
but this PR changes only the legacy `client/`.  In the clean current-master
merge, [`DhcpConfigV6`](https://github.com/AdguardTeam/AdGuardHome/blob/091d9287fc81613f08072a8d5381214f5422df87/client_v2/src/api/model/dhcpConfigV6.ts)
still contains only `range_start` and `lease_duration`, and
[`DhcpV6Modal`](https://github.com/AdguardTeam/AdGuardHome/blob/091d9287fc81613f08072a8d5381214f5422df87/client_v2/src/components/Dhcp/blocks/DhcpV6Modal.tsx#L13-L16)
neither displays nor submits a prefix source.

**Trigger/impact:** merge this branch into current master and open DHCPv6
settings.  There is no way to select `prefix_source: interface`, so the feature
requested by #6637 is inaccessible from the default current-master build.

**Smallest fix/test:** rebase, regenerate the `client_v2` API, implement the
control and state hydration/submission in `client_v2`, and cover static,
interface-with-host-template, and interface RA-only configurations in store and
component tests.

### 2. [P1, high confidence] An infinite prefix lifetime clears finite lease expiries

[`deadlineFromRemaining`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/routeradvstate.go#L786-L796)
represents `math.MaxUint32` (infinity) as `time.Time{}`.  However,
[`updateTrackedLeaseExpiry`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L1683-L1697)
treats that zero sentinel as an earlier deadline and assigns it to a finite
dynamic lease.

**Trigger/impact:** reconcile an existing finite lease against an interface
prefix reported as `forever`/`infinity` (or a Linux address without cache-info).
The lease expiry becomes year 1.  `leaseNotExpired` then treats it as
uncommitted/usable, `FindMACbyIP` treats it as expired, and the zero timestamp is
persisted.

A review-only test reproduced this exactly: expected the original finite
expiry, got `0001-01-01 00:00:00 +0000 UTC`.

**Smallest fix/test:** if `until.IsZero()`, do not clamp a finite lease expiry.
Add a regression that calls `setTrackedRangeStart` with `ValidSec:
math.MaxUint32` and asserts that the lease expiry remains finite and unchanged.

### 3. [P1, high confidence] Rapid Commit is not committed and ignores prefix validity

[`newPacketResponse`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L1117-L1135)
turns a Solicit with Rapid Commit into a Reply, but
[`commitLeaseLifetime`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L797-L831)
handles all Solicits through the default branch: it returns the full configured
lease time without capping it to the observed prefix and without setting
`lease.Expiry` or notifying persistence.

**Trigger/impact:** with a one-hour configured lease and a prefix valid for five
minutes, a Rapid-Commit client receives a one-hour Reply for an address that
the server leaves uncommitted.  The client can retain it after the delegated
prefix expires.  RFC 8415 section 18.3.1 requires a Rapid Commit exchange to be
processed as Request/Reply and committed.

A review-only test observed all three failures: `1h0m0s` instead of at most
`5m0s`, zero lease expiry, and no `LeaseChangedDBStore` notification.

**Smallest fix/test:** cap normal Solicit advertisements to the prefix
deadlines, and apply Request commit semantics when the Solicit contains Rapid
Commit.  Test both ordinary and Rapid Solicits; only Rapid Commit should update
expiry/persistence.

### 4. [P2, high confidence] DAD-failed/duplicated addresses remain eligible

Linux parsing records only `IFA_F_TEMPORARY` and `IFA_F_TENTATIVE` in
[`prefix_linux.go`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/aghnet/prefix_linux.go#L137-L144),
discarding `IFA_F_DADFAILED`.  The shared BSD parser recognizes only
`temporary` and `tentative` in
[`prefix.go`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/aghnet/prefix.go#L75-L99),
so real `ifconfig` states such as `duplicated` and `detached` are also lost.
All downstream consumers reject only `Tentative`.

**Trigger/impact:** IPv6 DAD finds a collision.  The unusable address can still
become the RA source/RDNSS address, a DHCPv6 DNS address, or the source of the
active /64, causing listener failure or advertisement of unusable network
state.

Review-only tests failed on both paths: BSD `duplicated` remained usable, and a
Linux-native `IFA_F_DADFAILED` test reported `DAD-failed address must be
excluded as unusable`.

**Smallest fix/test:** carry an explicit unusable/DAD-failed bit from Linux and
BSD observation and reject it in DNS, RA-source, and prefix selection.  Cover
`IFA_F_DADFAILED`, `duplicated`, `detached`, and mixed good/bad observations.

### 5. [P2, high confidence] Confirm answers binding ownership instead of on-link status

[`process`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L1027-L1080)
requires a lease owned by the extracted MAC.  If none exists,
[`replyStatusForProcessFailure`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L1228-L1242)
returns `NotOnLink`.

**Trigger/impact:** after lost lease storage, or when another DHCP server issued
the address, a client Confirms an address under the current or still-advertised
prefix.  AGH incorrectly tells it the address is off-link.  RFC 8415 section
18.3.3 requires Confirm to evaluate whether each address's prefix is appropriate
for the link, independently of an existing binding.

**Smallest fix/test:** give Confirm a prefix-based path over all requested IA
addresses.  Test unknown bindings under active and deprecated prefixes
(`Success`), a foreign prefix (`NotOnLink`), and a request with no evaluable
addresses (no Reply).

### 6. [P2, high confidence] Lease selection and commit are not atomic across renumbering

[`process`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L1048-L1079)
returns a lease pointer while holding `leasesLock`, releases the lock, then
reacquires it in `commitLease`.  During that gap,
[`setTrackedRangeStart`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/v6_unix.go#L1365-L1400)
can remove the prefix and lease.

**Trigger/impact:** an observation tick removes a prefix between lookup/reserve
and commit.  Renew/Rebind/Confirm can return Success using a removed binding;
a Solicit can advertise or Rapid-Commit a positive-lifetime address from the
old prefix.  This is a semantic TOCTOU, so the race detector does not catch it.

**Smallest fix/test:** select/revalidate and commit under one critical section,
then notify after unlocking.  Add a barrier-controlled test that removes the
prefix between selection and the old commit point and asserts no successful IA
is returned.

### 7. [P2, high confidence] OpenAPI omits `ra_slaac_only` from the new response contract

The PR changes `RASLAACOnly` to serialize as `ra_slaac_only` in
[`config.go`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/internal/dhcpd/config.go#L278-L282),
and the added legacy UI relies on that response value to decide whether a host
template is optional.  But
[`DhcpConfigV6`](https://github.com/AdguardTeam/AdGuardHome/blob/f70bb2ba2a7233ac32fb966164344df2f2a7c019/openapi/openapi.yaml#L2051-L2070)
does not declare it.

**Trigger/impact:** validate or generate a client for an interface RA-only
status response.  The response and schema disagree, and generated clients
cannot hydrate the mode correctly.

**Smallest fix/test:** add a read-only response property or split request and
response schemas, regenerate clients, and validate an interface RA-only status
response against the schema.

### Rebase housekeeping

The branch is 90 current-master commits behind.  Although Git reports a clean
merge, its changelog entry lands under the already released `v0.107.74`
section.  Please relocate it to current `Unreleased` while rebasing.

### Validation performed

Passed on the exact PR head and/or the clean current-master merge:

- `git diff --check` and `gofmt -d` on all changed Go files.
- `go test -race -count=1 ./internal/aghnet ./internal/dhcpd`.
- `go test -race -count=1 ./...`.
- `go vet ./...`.
- `make go-check` on the current-master merge (including `govulncheck`; no
  called vulnerabilities found).
- Linux, FreeBSD, and OpenBSD amd64 cross-compilation of `internal/aghnet` and
  `internal/dhcpd`; Darwin was built/tested natively.
- Legacy-client lint, typecheck, Vitest, and production Webpack build in an
  independent pass.

The three targeted review-only regression tests described above failed at the
intended assertions and were removed.  The original PR worktree is clean.  I
did not run privileged live DHCPv6/RA tests or frontend E2E tests against a real
backend.  This PR currently has no GitHub check runs.

Issue #7773, opened 2025-04-17
Upgrade to v0.108.0-b.67 failed


rphillips:
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) and found no answer

- [x] I have searched other issues and found no duplicates

- [x] I want to report a bug and not [ask a question or ask for help](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a)

- [x] I have set up AdGuard Home correctly and [configured clients to use it](https://github.com/AdguardTeam/AdGuardHome/wiki/Clients). (Use the [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) for help with installing and configuring clients.)

### Platform (OS and CPU architecture)

Linux, AMD64 (aka x86_64)

### Installation

GitHub releases or script from README

### Setup

On one machine

### AdGuard Home version

v0.108.0-b.67

### Action

I upgraded to v0.108.0-b.67 using the automated button within the website. Usually this works fine, but with this release the AdguardHome binary is failing to start.

```
root@adguard:/opt/AdGuardHome# ./AdGuardHome
2025/04/17 19:19:48.692167 [info] AdGuard Home, version v0.108.0-b.67
2025/04/17 19:19:48.695718 [info] tls_manager: using default ciphers
2025/04/17 19:19:48.700403 [info] Initializing auth module: /opt/AdGuardHome/data/sessions.db
2025/04/17 19:19:48.700538 [info] auth: initialized.  users:1  sessions:3
2025/04/17 19:19:48.700749 [info] webapi: initializing
2025/04/17 19:19:48.722541 [error] closing query log: flushing log buffer: nothing to write to a file
2025/04/17 19:19:48.722613 [fatal] newServerConfig: constructing tls config: parsing tls key pair: tls: failed to find any PEM data in certificate input
```

```
root@adguard:/opt/AdGuardHome# ./AdGuardHome --version
AdGuard Home, version v0.108.0-b.67
```

Partial Config:

```
tls:
  enabled: true
  server_name: <redacted>
  force_https: false
  port_https: 55355
  port_dns_over_tls: 785
  port_dns_over_quic: 784
  port_dnscrypt: 0
  dnscrypt_config_file: ""
  allow_unencrypted_doh: true
  certificate_chain: ""
[...]

Issue #741, opened 2019-05-01
[Request]web interface can sets different port from DoH port.


280blocker:
If DoH is enabled, admin interface and DoH can be accessed by same port.
I want to publish DoH open and restrict access to admin interface by the port number.

Is it possible to use different ports for DoH and admin interface?

ameshkov:
@280blocker thank you!

I've marked this issue as a feature request. This is doable of course, but not as easy as we'd like to.

Meanwhile, the temporary solution would be to use a simple nginx proxy on a different port.

ibksturm:
hi @280blocker 

lock at [https://github.com/AdguardTeam/AdGuardHome/issues/662](https://github.com/AdguardTeam/AdGuardHome/issues/662)

and on my gitrepo [https://github.com/ibksturm/dnscrypt-switzerland](https://github.com/ibksturm/dnscrypt-switzerland)

Onepamopa:
This should be high priority ...

ameshkov:
Merging https://github.com/AdguardTeam/AdGuardHome/issues/2548 to this task as they seem to be alike.

So what should be done:

1. Separate configuration for the web admin interface and for DOH/DOT
2. Allow them to run on different network interfaces/ports
3. Allow them to run on the same interface/port, but to have different domain names
4. Allow configuring different HTTPS configuration for both of them

timkgh:
At least for the admin UI, it should not error on self signed certificates. E.g. I use the excellent [mkcert](https://github.com/FiloSottile/mkcert) to have a "home" CA that I import in my browsers, but AGH does not like it:
```
Certificate chain is invalid
Your certificate does not verify: x509: certificate signed by unknown authority
```

ameshkov:
@timkgh I don't think this is a suitable issue for discussing this. However, there's an easy way to solve that issue, make sure your self-signed cert is added to trusted certs on the server.

timkgh:
+1 to being able to limit the admin UI to a specific VLAN/interface.

Habetdin:
Quoting #3692 here just to keep additional information in place:

> ### Expected Behavior
> 
> DNS over HTTPS to listen the same IPs as the regular DNS, DNS over TLS and DNS over QUIC, etc
> ### Actual Behavior
> 
> DNS over HTTPS listens only on the Web interface IP address
> ### Additional Information
> 
> `AdGuardHome.yaml` contains next configuration (I've replaced the real IPs with `<stubs>`:
[...]

PR #8356, opened 2026-04-24
home: add tls.admin_listen_addr for separate https admin listener

### LINKED ISSUE #7424: Why use same https port for DNS and Web Portal
### Prerequisites

### The problem

Whys use same port for DNS service and Administration portal.

How can we filter public VPS access to the administration portal? There is no sense of having this like this.

### Proposed solution

Different port for different services

_No response_

### LINKED ISSUE #7598: https admin interface setting independent of enabling DoH
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [x] I want to request a feature or enhancement and not ask a question

### The problem

Please provide a separate setting to enable https for the admin interface that is independent of enabling DoH too. While I want to access the web admin interface in a secure way, I do not want to give clients DoH access too.

### Proposed solution

Simple config option that disables these endpoints:

https://github.com/AdguardTeam/AdGuardHome/blob/6633ad63049468ad35283ee914651587cf39ef2d/internal/dnsforward/http.go#L742

_No response_

Raviu56:
## Summary

Addresses #7424 and #7598 by adding a single optional `tls.admin_listen_addr` YAML setting (a `netip.AddrPort`, e.g. `127.0.0.1:4443`) that decouples the HTTPS admin UI/API from DoH while sharing the same TLS certificate.

- **Unset / zero (default)** — no change. One HTTPS server on `tls.port_https` serves both admin UI and DoH, exactly as today. Existing installs are unaffected and DNS clients using DoH see no change.
- **Set** — two HTTPS servers sharing the same TLS certificate:
  - `tls.port_https` serves **DoH only** (`/dns-query`, etc.). Admin UI + API are not reachable here.
  - `tls.admin_listen_addr` serves **admin UI + API only**. DoH routes are not registered on this listener.

This covers both open feature requests without risking DoH clients:

- **#7424** — set `admin_listen_addr: 127.0.0.1:4443` → DoH stays public on 443, admin HTTPS is LAN/VPN-only.
- **#7598** — set `admin_listen_addr: 0.0.0.0:443` and `port_https: 0` → admin HTTPS only, no DoH over HTTPS.

`HTTPSListenAddrs` (DDR SVCB advertisement) and mobileconfig / DNS-encryption display intentionally stay tied to `port_https`, so DNS clients see no change. HTTP/3 on `port_https` mirrors the HTTPS mux split, so DoH-over-HTTP/3 is not blocked by the admin auth middleware and admin routes are not exposed there.

Port conflicts involving `admin_listen_addr` are validated at config parse time (including `--check-config`), not just as a runtime bind failure. `checkPortAvailability` validates the admin port on the admin listener's own IP (which may differ from the web API bind host). `tlsConfigChanged` uses a fresh shutdown context per server so the admin server is not handed an already-cancelled context.

**Scope:** backend + YAML only. No frontend / UI changes. `AdminListenAddr` is intentionally not exposed in the JSON API (`json:"-"`); the field is preserved across frontend TLS-settings saves via `setPrivateFieldsAndCompare`. The `internal/next` experimental code path is not touched. Default plain-HTTP admin behavior (`http.address`, default `:3000`) is unchanged; only the `force_https` redirect target is updated to point at the admin port when the feature is enabled.

Example config:

```yaml
tls:
  enabled: true
  port_https: 443
  admin_listen_addr: 127.0.0.1:4443
  certificate_path: /etc/adguardhome/cert.pem
  private_key_path: /etc/adguardhome/key.pem
```

### Changes

- New field `tlsConfigSettings.AdminListenAddr netip.AddrPort` (YAML: `admin_listen_addr`, `json:"-"`).
- Split the shared mux into `adminMux` (web UI + API + install + control handlers) and `dohMux` (DoH routes) when the feature is enabled; single shared mux when not.
- Second `http.Server` (`adminHTTPSServer`) with its own lifecycle, started when `admin_listen_addr` is valid and non-zero; shares the cert with the DoH HTTPS server.
- HTTP/3 (`mustStartHTTP3`) mirrors `serveTLS` mux selection + auth decision.
- `validateConfig`, `checkPorts`, `validatePorts`, and `checkPortAvailability` all cover the admin port (and the admin port uses its own IP in binding availability checks).
- `validateTLSSettings` reads `AdminListenAddr` from the server-side `tlsConf` snapshot (not the frontend payload) so UI `port_https` changes still trigger the admin-port conflict check.
- `force_https` plain-HTTP redirect target switches to the admin port when configured.
- CHANGELOG entry under **Added**.
- New unit tests in `internal/home/adminlistenaddr_internal_test.go` covering the helper, port-conflict detection, config preservation across frontend updates, and mux-split routing.
- Audited against [CodeGuidelines/Go](https://github.com/AdguardTeam/CodeGuidelines/blob/master/Go/Go.md) (line length ≤ 100, no naked returns, no `init`, lowercase log/error messages, godoc `[bracket]` links, explicit zero values on error returns).

### Test results

Three scenarios, 14 assertions, **all pass** against a binary built from this branch with a shared self-signed cert.

**Scenario A — feature off (`admin_listen_addr: ""`)**

| # | Check | Expected | Got |
|---|---|---|---|
| A1 | `GET https://:8443/` (admin UI) | 302 | **302** |
| A2 | `GET https://:8443/dns-query?…` (DoH) | 200 | **200** |
| A3 | `GET https://:4444/` (admin port not bound) | refused | **refused** |

**Scenario B — feature on (`admin_listen_addr: 127.0.0.1:4444`, `port_https: 8443`, `force_https: true`)**

| # | Check | Expected | Got |
|---|---|---|---|
| B1 | `GET https://127.0.0.1:4444/` (admin UI) | 302 | **302** |
| B2 | `GET https://127.0.0.1:4444/dns-query?…` (DoH not served here) | 4xx | **401** |
| B3 | `GET https://:8443/control/status` (admin API not served here) | 404 | **404** |
| B4 | `GET https://:8443/dns-query?…` (DoH still works) | 200 | **200** |
| B5 | TLS cert fingerprint on :8443 vs :4444 | byte-identical | **match** |
| B6 | `GET http://:3000/login.html` `Location:` header | `https://…:4444/login.html` | **`https://127.0.0.1:4444/login.html`** |

**Scenario C — port conflict (`admin_listen_addr: 127.0.0.1:8443`, `port_https: 8443`)**

| # | Check | Expected | Got |
|---|---|---|---|
| C1 | `AdGuardHome --check-config` | exit non-zero, mentions duplicated port | **exit 1**, `validating tcp ports: duplicated values: [8443]` |

`go test ./...`, `go vet ./...`, and `bash scripts/make/go-lint.sh` (gofumpt, govulncheck, gocyclo, gocognit, ineffassign, unparam, misspell, gosec, errcheck, staticcheck) are green.

Closes #7424
Closes #7598

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ See also the [v0.107.75 GitHub milestone][ms-v0.107.75].
 NOTE: Add new changes BELOW THIS COMMENT.
 -->
 
+### Added
+
+- New optional `tls.admin_listen_addr` configuration setting (an `IP:port` listen address, e.g. `127.0.0.1:4443`).  When set, AdGuard Home starts a second HTTPS server on this address dedicated to the admin UI and API, while the HTTPS server on `tls.port_https` serves DNS-over-HTTPS only.  Both servers share the same TLS certificate.  When unset (the default), behavior is unchanged: one HTTPS server on `tls.port_https` serves both admin UI and DoH ([#7424], [#7598]).
+
 ### Changed
 
 - `enable_dnssec` in `dns` configuration now defines whether the proxy should set the DO flag in the upstream requests, the default is `true` ([#7046]).
@@ -27,6 +31,8 @@ NOTE: Add new changes BELOW THIS COMMENT.
 - Safe Browsing and Parental Control labels on the General Settings page not updating after changing the UI language.
 
 [#7046]: https://github.com/AdguardTeam/AdGuardHome/issues/7046
+[#7424]: https://github.com/AdguardTeam/AdGuardHome/issues/7424
+[#7598]: https://github.com/AdguardTeam/AdGuardHome/issues/7598
 
 <!--
 NOTE: Add new changes ABOVE THIS COMMENT.
--- a/internal/home/adminlistenaddr_internal_test.go
+++ b/internal/home/adminlistenaddr_internal_test.go
@@ -1,0 +1,237 @@
+package home
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/netip"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestAdminListenAddr verifies that [adminListenAddr] returns the configured
+// address only when it is valid and has a non-zero port.
+func TestAdminListenAddr(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name string
+		in   netip.AddrPort
+		want netip.AddrPort
+	}{{
+		name: "zero",
+		in:   netip.AddrPort{},
+		want: netip.AddrPort{},
+	}, {
+		name: "zero_port",
+		in:   netip.AddrPortFrom(netip.MustParseAddr("127.0.0.1"), 0),
+		want: netip.AddrPort{},
+	}, {
+		name: "loopback_port",
+		in:   netip.MustParseAddrPort("127.0.0.1:4443"),
+		want: netip.MustParseAddrPort("127.0.0.1:4443"),
+	}, {
+		name: "unspecified_port",
+		in:   netip.MustParseAddrPort("0.0.0.0:8443"),
+		want: netip.MustParseAddrPort("0.0.0.0:8443"),
+	}}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			conf := &tlsConfigSettings{AdminListenAddr: tc.in}
+			assert.Equal(t, tc.want, adminListenAddr(conf))
+		})
+	}
+
+	t.Run("nil", func(t *testing.T) {
+		t.Parallel()
+
+		assert.Equal(t, netip.AddrPort{}, adminListenAddr(nil))
+	})
+}
+
+// TestValidatePorts_adminHTTPS verifies that [validatePorts] includes the
+// optional dedicated admin HTTPS port in its uniqueness check.
+func TestValidatePorts_adminHTTPS(t *testing.T) {
+	t.Parallel()
+
+	// Baseline ports used below.  These are non-overlapping so that the test
+	// cases below fail or pass solely based on the admin HTTPS port.
+	const (
+		bindPort        tcpPort = 3000
+		dohPort         tcpPort = 443
+		dotPort         tcpPort = 853
+		dnscryptTCPPort tcpPort = 5443
+		dnsPort         udpPort = 53
+		doqPort         udpPort = 784
+	)
+
+	testCases := []struct {
+		name           string
+		adminHTTPSPort tcpPort
+		wantErr        bool
+	}{{
+		name:           "unused",
+		adminHTTPSPort: 0,
+		wantErr:        false,
+	}, {
+		name:           "unique",
+		adminHTTPSPort: 4443,
+		wantErr:        false,
+	}, {
+		name:           "conflicts_with_doh",
+		adminHTTPSPort: dohPort,
+		wantErr:        true,
+	}, {
+		name:           "conflicts_with_webapi",
+		adminHTTPSPort: bindPort,
+		wantErr:        true,
+	}, {
+		name:           "conflicts_with_dot",
+		adminHTTPSPort: dotPort,
+		wantErr:        true,
+	}, {
+		name:           "conflicts_with_plain_dns",
+		adminHTTPSPort: tcpPort(dnsPort),
+		wantErr:        true,
+	}}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := validatePorts(
+				bindPort,
+				dohPort,
+				dotPort,
+				dnscryptTCPPort,
+				tc.adminHTTPSPort,
+				dnsPort,
+				doqPort,
+			)
+			if tc.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+// TestSetPrivateFieldsAndCompare_adminListenAddr verifies that
+// [tlsConfigSettings.setPrivateFieldsAndCompare] preserves the server-side
+// [tlsConfigSettings.AdminListenAddr] value across updates from the frontend
+// and that [cmp.Equal] is able to compare the field.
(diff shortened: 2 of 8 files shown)


rphillips:
This PR might fix https://github.com/AdguardTeam/AdGuardHome/issues/7773 but 7773 is still a regression.

Sil3ntVip3r (changes requested):
I reviewed exact head `ea6ad9a31b4e67a35ff3de992991a51c329e6246`, including the listener lifecycle, TLS configure path, HTTP/1.1+h2c routing, HTTP/3 routing, redirects, port validation, linked issues, and current `master`.  The split is a useful direction, and the default combined-listener path remains intact, but the new modes still have production blockers.

## Confirmed blockers

1. **P1: saving TLS settings through the UI disables the dedicated admin listener.**

   `AdminListenAddr` is `json:"-"` (`config.go:323-326`), so the request decoded by `handleTLSConfigure` has a zero value.  `setConfig` restores the private field only into its by-value `newConf` and `m.conf` (`tls.go:522-535`), but the restart at `tls.go:651-652` passes the original raw `req` to `tlsConfigChanged`.  `web.go:281-296` consequently shuts down the admin server and leaves it disabled.  The persisted config still retains the split, so the main HTTPS listener comes back DoH-only and the UI can be locked out until a process restart or certificate reload.

   The smallest fix is to validate and restart from the normalized effective config (for example, a clone of `m.conf` after private fields are restored), never from the raw JSON request.  Please add a configure-handler regression: begin with a YAML admin address, submit a frontend-style TLS update omitting that field, and assert that the effective restart config retains the address and both listeners rebind with their routes isolated.

2. **P1: the advertised admin-HTTPS-only configuration cannot start.**

   `tlsConfigChanged` makes `enabled` require `PortHTTPS != 0` at `web.go:254-257`, then derives `adminEnabled` from that same value at `web.go:281-284`.  Therefore the PR's documented #7598 configuration (`admin_listen_addr` set, `port_https: 0`, valid certificate) enables neither listener.  `handleHTTPSRedirect` also returns immediately when the DoH HTTPS server is nil (`control.go:357-360`), so this mode cannot force HTTPS either.

   A deterministic review-only regression failed exactly at `adminHTTPSServer.enabled`: expected `true`, got `false`.  Separate certificate readiness from listener readiness: `dohEnabled` should require a nonzero `PortHTTPS`, while `adminEnabled` should require a valid admin address.  Parse/share the certificate when either listener is enabled and gate redirects on the actual admin listener.  Cover this with a real loopback lifecycle test proving that UI/API work while no DoH HTTPS listener exists.

3. **P1: split mode removes supported unencrypted DoH.**

   With an admin address set, `registerDoHHandlers` moves all DoH routes exclusively to `dohMux` (`dns.go:594-615`).  The plain HTTP/h2c server always serves only `web.conf.mux` (`web.go:317-331`).  As a result, `http.doh.insecure_enabled: true` no longer makes `/dns-query` reachable for reverse proxies.

   This needs listener-specific routing (for example, separate plain, admin-TLS, and encrypted-DoH handlers); registering DoH on both current muxes would expose it again on admin TLS.  Add HTTP/1.1 and h2c tests with split mode + insecure DoH, alongside negative checks on admin TLS and positive HTTP/2/3 DoH checks on the public listener.

4. **P2: admin responses advertise the DoH-only HTTP/3 endpoint.**

   `handleHTTPSRedirect` emits `Alt-Svc: h3=":<port_https>"` whenever HTTP/3 is enabled (`control.go:371-384`).  In split mode, however, HTTP/3 on that port intentionally uses only `dohMux` (`web.go:590-610`).  A browser may cache that alternative for the admin origin and send later UI/API requests to a listener that returns 404 for them.

   Suppress the DoH `Alt-Svc` on admin responses unless there is an admin HTTP/3 listener, and add a response-level regression.

5. **P2: `force_https` can redirect to an address where admin HTTPS is not listening.**

   `readHTTPSRedirectConfig` substitutes only the admin port; `httpsURL` keeps the incoming request host (`control.go:420-454`).  For example, with admin bound to `127.0.0.1:4443`, a request received as `http://192.0.2.10:3000/login.html` is redirected to `https://192.0.2.10:4443/login.html`, not the configured admin endpoint.  A review-only regression reproduced that exact output.

   Please define an externally reachable admin-host policy.  A concrete bind address can be used when appropriate; an unspecified bind address should preserve the request host; deployments needing a certificate hostname may require a separate advertised-host setting.  Test concrete/unspecified IPv4 and IPv6 plus hostname cases.

6. **P2: conflict validation discards the address part of `AddrPort`.**

   Startup and API validation reduce all TCP listeners to bare port numbers (`config.go:810-835`, `tls.go:693-733`).  This rejects bind-compatible endpoints such as public DoH on `192.0.2.10:443` and admin on `127.0.0.1:443`.  Canonical issue #741 explicitly requires different interfaces/ports, and #7424 has since been closed as its duplicate.

   Validate endpoint overlap using address + port, with wildcard-address semantics.  Tests should accept disjoint concrete IPs on the same port and reject identical or wildcard-overlapping endpoints.

## Branch and verification

The PR is currently 78 commits behind `master` and GitHub reports `CONFLICTING/DIRTY`.  A disposable merge conflicted in `CHANGELOG.md`, `internal/home/home.go`, `internal/home/tls.go`, and `internal/home/web.go`; current HTTPS lifecycle and TLS validation have materially changed, so this needs a careful rebase/reimplementation rather than a mechanical conflict resolution.  Please also update the description: #7424 is now closed as a duplicate of #741, while #7773 is a separate certificate-less encryption regression.

Observed on the exact head:

- focused new tests, `-race -count=20`: pass;
- `go test -race -count=1 ./internal/home`: pass;
- `go test -race -count=1 ./...`: pass;
- `go vet ./...`: pass;
- `gofmt -d` and `git diff --check`: clean;
- two review-only regressions for admin-only readiness and redirect targeting: fail at the intended assertions; the temporary test file was removed and the worktree is clean;
- `make go-check`: did not complete in this environment because the lint script stopped while bootstrapping its pinned Go 1.26.2 toolchain, so I am not claiming that full gate here.

No authentication bypass was found: the dedicated admin HTTPS handler remains behind the existing auth middleware, and the public HTTP/3 handler is DoH-only.

Issue #8348, opened 2026-04-18
Warning at start when no DHCP dæmon configured: dhcpd: warning: creating dhcpv4 server err="dhcpv4: invalid IP is not an IPv4 address"


mirabilos:
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) and found no answer

- [x] I have searched other issues and found no duplicates

- [x] I want to report a bug and not [ask a question or ask for help](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a)

- [x] I have set up AdGuard Home correctly and [configured clients to use it](https://github.com/AdguardTeam/AdGuardHome/wiki/Clients). (Use the [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions/categories/q-a) for help with installing and configuring clients.)

### Platform (OS and CPU architecture)

Linux, ARMv6

### Installation

GitHub releases or script from README

### Setup

On one machine

### AdGuard Home version

just updated to latest

### Action

I start AGH.

### Expected result

No superfluous warnings.

### Actual result

    dhcpd: warning: creating dhcpv4 server err="dhcpv4: invalid IP is not an IPv4 address"

This occurs when no DHCP dæmon is configured, so naturally there’s no address for it to use.

### Additional information and/or screenshots

The setup is that this is a normal host in the LAN that only acts as resolver, not as DHCP server; it’s got a static IP itself, and DHCP is handled by the gateway.

tjharman:
I don't see this on my AdguardHome instance - I also have DHCP disabled.

Can you post the output from the dhcp: section of your config file please?
Mine looks like so:

```
dhcp:
  enabled: false
  interface_name: ""
  local_domain_name: lan
  dhcpv4:
    gateway_ip: 192.168.22.1
    subnet_mask: 255.255.255.0
    range_start: 192.168.22.10
    range_end: 192.168.22.20
    lease_duration: 86400
    icmp_timeout_msec: 0
    options: []
  dhcpv6:
    range_start: ""
    lease_duration: 86400
    ra_slaac_only: false
    ra_allow_slaac: false
```

mirabilos:
On Sun, 19 Apr 2026, Tim Harman wrote:

>Can you post the output from the dhcp: section of your config file please?

Sure:

dhcp:
  enabled: false
[...]

PR #8375, opened 2026-05-11
dhcpd: skip empty disabled dhcpv4 config

0xnuminous:
Updates #8348.

When DHCPv4 is disabled and its address fields are empty, `v4Create` now returns an unconfigured disabled server without validating the empty IPv4 addresses.

Disabled but configured DHCPv4 settings still go through validation, so existing status output for saved disabled DHCPv4 ranges is preserved.

Tests:

- `go test ./internal/dhcpd`
- `make go-check`

Note: `make go-check` prints existing `govulncheck` findings for the pinned Go 1.26.2 standard library, but the target exits successfully.

### DIFF
--- a/internal/dhcpd/dhcpd_unix_internal_test.go
+++ b/internal/dhcpd/dhcpd_unix_internal_test.go
@@ -129,6 +129,14 @@ func TestV4Server_badRange(t *testing.T) {
 	}
 }
 
+func TestV4Server_Create_disabled(t *testing.T) {
+	s, err := v4Create(&V4ServerConf{})
+
+	require.NoError(t, err)
+	assert.NotNil(t, s)
+	assert.False(t, s.enabled())
+}
+
 // cloneUDPAddr returns a deep copy of a.
 func cloneUDPAddr(a *net.UDPAddr) (clone *net.UDPAddr) {
 	return &net.UDPAddr{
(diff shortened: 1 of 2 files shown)


0xnuminous:
One clarification after re-reading #8348 and the recent edge note: this PR is intentionally narrower than changing log level.

The edge build now logs the empty disabled DHCPv4 validation failure at debug level, but the reset/empty disabled config still causes `v4Create` to construct a validation error for address fields that are not meaningful while DHCPv4 is disabled. This patch skips that validation only when DHCPv4 is disabled and all IPv4 address fields are empty.

Disabled DHCPv4 configs that still contain saved address settings continue to go through the existing validation path, which preserves the current status behavior for configured-but-disabled DHCPv4 ranges.

Rechecked locally before marking ready:

- `go test ./internal/dhcpd`
- `make go-check`

`make go-check` exits successfully; it still prints the existing upstream `govulncheck` findings for the pinned Go 1.26.2 toolchain.

Sil3ntVip3r (approved):
Reviewed exact head `f636f1ad57b81b63bbfd800ae8603faf44c79d39` and transplanted the two-file change onto current `upstream/master` (`091d9287fc81613f08072a8d5381214f5422df87`).

The test has valid red-before-green evidence on current master: with only `TestV4Server_Create_disabled`, `v4Create(&V4ServerConf{})` returns `dhcpv4: invalid IP is not an IPv4 address`; with the production guard, the focused test passes 20 times under `-race`. The complete `internal/dhcpd` race suite and `make go-check` also pass on the current-base transplant, and `govulncheck` reports zero reachable vulnerabilities.

The early return is limited to a disabled configuration with all four IPv4 address fields unset. Disabled configurations that retain address data still follow the existing validation and status-preservation path. I found no correctness blocker. The branch is 71 commits behind current master, so it should be rebased before merge.

Issue #8036, opened 2025-09-28
Blocking services to a specific client on a schedule


sfizuli:
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [x] I have searched other issues and found no duplicates

- [x] I want to request a feature or enhancement and not ask a question

### The problem

It is not possible to block services to a client by schedule.

### Proposed solution

Version 107.66. OpenWrt
How to set a schedule to block services to a specific client? For example, you need to block a child's device from accessing Tik-Tok between 15 and 20 hours on weekdays.

Dynamic5912:
Yes it is.

Use the per-client blocking.

sfizuli:
Well, when using the client's settings, there is no option to enable or disable services on a schedule. There is only a way to completely disable or enable the service for the client, and there is no schedule!

tjharman:
@sfizuli In the meantime, you can use the API to block/unblock services, and then feed this into cron.  It's a little complex to get working, but it works.

Issue #7146, opened 2024-07-21
Change the blocked services so that the pause schedule can be for specific services


virtualvoyager367:
### Prerequisites

- [X] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [X] I have searched other issues and found no duplicates

- [X] I want to request a feature or enhancement and not ask a question

### The problem

When you add a schedule to it , it unblocks ALL services within the schedule

### Proposed solution

Make it so you select what services it unblocks for the schedule so certain things stay blocked while others can be used within certain times

### Alternatives considered and additional information

Only other ways are to modify the software with your own code that is NOT online because it is not made yet

reijop:
I am doing this currently via the API with ( curls | jq | sed ) from a home assistant rest_switch and a schedule.  It turns just youtube on and off for a specific client, but the overhead smells like Yak. +1 for wanting this in the service directly.

Use case - kids network with a list of services I never want enabled as well as a few services I wish to enable on a schedule.

I suppose a workaround would also be to blacklist via custom filtering rules and then use the client services schedule for the time-allowed services. This would be 100% easier if there were a way to specify a shortcut in the rewrite language for an existing 'service' as defined by adguard, so you could do something like: 
```
# blacklist by service, but only on one client, always.
||[service:roblox]*^$important,$client=kinder
```

PR #8380, opened 2026-05-16
feat(filtering): add time-windowed filtering support to custom filtering rules

hossihub:
This change enables users to apply time-based constraints to custom

filtering rules via new syntax modifiers: $start, $end
   - Added support for `$start='HH:mm'` and `$end='HH:mm'` in rule strings.
   - Implemented logic to handle standard windows and cross-midnight intervals.
   - Integrated time-check into the DNS filtering pipeline (matchHost).

   - Leverages the existing global timezone settings for consistency.

   - Added a 'now' function hook in DNSFilter for deterministic unit testing. Syntax
   - ||example.com^$start='09:00'$end='18:00' (Active during work hours)

   - @@||example.com^$start='22:00'$end='04:00' (Active during night hours)

### DIFF
--- a/client/src/components/Filters/Services/Form.tsx
+++ b/client/src/components/Filters/Services/Form.tsx
@@ -1,10 +1,11 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 
 import { Trans, useTranslation } from 'react-i18next';
 
 import { Controller, useForm } from 'react-hook-form';
 
 import { ServiceField } from './ServiceField';
+import { ScheduleForm } from './ScheduleForm';
 
 export type BlockedService = {
     id: string;
@@ -17,6 +18,14 @@ export type ServiceGroups = {
     id: string;
 }
 
+export type ServiceSchedule = {
+    id: string;
+    schedule?: {
+        time_zone: string;
+        [key: string]: any;
+    };
+};
+
 type FormValues = {
     blocked_services: Record<string, boolean>;
 };
@@ -25,7 +34,9 @@ interface FormProps {
     initialValues: Record<string, boolean>;
     blockedServices: BlockedService[];
     serviceGroups: ServiceGroups[];
-    onSubmit: (values: FormValues) => void;
+    serviceSchedules?: ServiceSchedule[];
+    onSubmit: (values: FormValues & { services?: ServiceSchedule[] }) => void;
+    onScheduleSubmit?: (serviceId: string, schedule: any) => void;
     processing: boolean;
     processingSet: boolean;
 }
@@ -34,11 +45,15 @@ export const Form = ({
     initialValues,
     blockedServices,
     serviceGroups,
+    serviceSchedules = [],
+    onSubmit,
+    onScheduleSubmit,
     processing,
     processingSet,
-    onSubmit,
 }: FormProps) => {
     const { t } = useTranslation();
+    const [scheduleModalOpen, setScheduleModalOpen] = useState(false);
+    const [currentServiceId, setCurrentServiceId] = useState<string | null>(null);
 
     const {
         handleSubmit,
@@ -63,6 +78,14 @@ export const Form = ({
         }, {} as Record<string, BlockedService[]>);
     }, [blockedServices]);
 
+    const serviceScheduleMap = useMemo(() => {
+        const map: Record<string, any> = {};
+        serviceSchedules.forEach(svc => {
+            map[svc.id] = svc.schedule;
+        });
+        return map;
+    }, [serviceSchedules]);
+
     const handleToggleAllServices = (isSelected: boolean) => {
         blockedServices.forEach((service) => {
             if (!isServicesControlsDisabled) {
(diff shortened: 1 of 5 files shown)


Sil3ntVip3r (commented):
There is a backward-compatibility issue in the legacy schedule branch at `internal/filtering/blocked.go:181-185`. Existing AdGuard Home behavior applies blocked-service rules only when `!bsvc.Schedule.Contains(now)`, and the UI describes this range as an inactivity/pause schedule. With this change, an existing configuration with no per-service entries instead applies the rules when `Contains(now)` is true. That reverses the schedule after upgrade: services become blocked during the configured pause period and unblocked outside it. Could the legacy branch preserve `!Contains`, with the new per-service semantics handled separately and consistently with the pause-schedule contract? A fixed-clock regression should cover times inside and outside the configured range for a legacy `BlockedServices` value whose `Services` slice is empty.

Sil3ntVip3r (changes requested):
Requesting changes on exact head
`82407f95eaf37245986749badd2ae3ce2f18b13d`.  The legacy-schedule reversal
reported in my earlier review is still present, and the full integration review
found additional data-loss and policy-inversion blockers.

1. **Existing configurations are reversed after upgrade.**
   `internal/filtering/blocked.go:181-185` applies legacy blocked-service rules
   when `Schedule.Contains(now)` is true.  Existing behavior applies them when
   `!Contains(now)`, because this schedule is an inactivity/pause schedule.
   Every existing scheduled configuration would therefore block during its
   configured pause and unblock outside it.

2. **The new per-service behavior is also the inverse of its UI.**
   `blocked.go:166-178` adds or retains a service while its schedule contains
   `now` and removes it outside that range.  The reused `ScheduleForm` describes
   the selected range as “Pause service blocking” and an “inactivity period.”
   For example, selecting Monday 09:00–17:00 as a pause blocks the service from
   09:00–17:00 and allows it for the rest of the day.  It can even activate an
   unselected service during the purported pause.

   There is a second interaction at `blocked.go:181-191`: as soon as *any*
   per-service entry exists, the global inactivity schedule is ignored for
   every other service.  With YouTube and Facebook selected, a YouTube override
   makes Facebook continuously blocked even during the still-configured global
   pause.

   The effective policy should be computed per selected ID: use that service's
   pause schedule when an override exists, otherwise fall back to the global
   pause schedule, and omit the service while the effective schedule contains
   `now`.  A schedule should not activate an ID that the user did not select.

3. **Current clients cannot preserve the new field and will silently delete
   it.**  Current master builds `client_v2`, but this PR changes only the legacy
   `client/`.  `openapi/openapi.yaml` and the generated
   `BlockedServicesSchedule` model still expose only `ids` and `schedule`.
   Current UI update paths therefore PUT only those fields.  The changed
   handler at `blocked.go:287-310` decodes a complete replacement and assigns
   it, so toggling a service, editing the global inactivity schedule, or using
   the query-log “allow service” action erases every saved per-service schedule.

   Please add the public schema and compatibility behavior, regenerate the
   client, implement the feature in the shipped `client_v2`, and make every
   read-modify-write path preserve `services`.  API round-trip tests should
   prove that an IDs-only UI action cannot discard existing schedules.

4. **The shared configuration is incomplete and can break startup.**
   `BlockedServices` is also used for persistent clients, but
   `internal/filtering/filter.go:702-708` ignores `Services`, and the client
   HTTP conversion does not preserve it.  In addition, `FilterUnknownIDs`
   filters stale `IDs` but not stale `Services`; validation then rejects the
   latter, so a removed or renamed service with a saved schedule can make the
   next global or persistent-client initialization fail.  Either implement the
   same effective-policy and round-trip behavior for clients, or keep this data
   out of the shared type; sanitize both collections consistently.

5. **The proposed and implemented scopes do not match.**  The title/body claim
   custom-rule `$start`/`$end` modifiers, but the diff never changes the rule
   parser and instead adds per-blocked-service configuration.  There is no
   linked design issue and no backend, API, or UI regression test.  The commit
   also says a deterministic clock hook was added, but the implementation calls
   `time.Now()` directly.  This should first be reconciled with the existing
   per-service scheduling discussions (notably #7146 and #8036), then rebuilt
   around an injected/fixed clock and an explicit compatibility contract.

Minimum regression coverage should include fixed times inside/outside: legacy
global pause, one per-service override, mixed override-plus-global fallback,
selected and unselected IDs, global and per-client policy; stale IDs; GET/PUT
round trips; and each shipped `client_v2` update path.

Validation recorded:

- Exact head: `git diff --check`, `gofmt -d`, `go vet` and race-enabled
  filtering tests passed.
- Clean merge into current master
  `091d9287fc81613f08072a8d5381214f5422df87` succeeded; full
  `go test -race -count=1 ./internal/...` passed.
- On that merge, frontend dependency install, type-check, and 35 tests passed.
- Frontend lint failed at
  `client/src/components/Filters/Services/ServiceField.tsx:16` (`max-len`), and
  an independent Prettier check rejected all three changed TSX files.
- GitHub currently reports no CI checks, maintainer review, linked issue, or
  author response.

Given the behavior reversal and silent deletion path, this needs a scoped
redesign rather than incremental merge fixes on the current head.

PR #8394, opened 2026-05-22
fix: ignore downgrade update checks

### LINKED ISSUE #8393: Updating to version v0.107.76 prompts to “upgrade” to v0.107.75
### Prerequisites

Linux, AMD64 (aka x86_64)

### Installation

### Setup

On one machine

### AdGuard Home version

v0.107.76

### Action

Upgrade to v0.107.76

### Expected result

Not be promoted to “upgrade” to a lower version number 

### Actual result

Promoted to “upgrade” to v0.107.75

Image

---

Image

immanuwell:
Fixes #8393.

Repro: run v0.107.76, point /control/version.json at v0.107.75, then hit Check for updates. It offers a downgrade as an upgrade

Now the updater ignores true downgrades, but beta checks still work. Small fix

### DIFF
--- a/internal/updater/check.go
+++ b/internal/updater/check.go
@@ -8,11 +8,13 @@ import (
 	"maps"
 	"net/http"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/AdguardTeam/AdGuardHome/internal/aghalg"
 	"github.com/AdguardTeam/golibs/errors"
 	"github.com/AdguardTeam/golibs/ioutil"
+	"github.com/AdguardTeam/AdGuardHome/internal/version"
 	"github.com/c2h5oh/datasize"
 	"golang.org/x/mod/semver"
 )
@@ -99,7 +101,7 @@ func (u *Updater) parseVersionResponse(ctx context.Context, data []byte) (Versio
 	info.Announcement = versionJSON["announcement"]
 	info.AnnouncementURL = versionJSON["announcement_url"]
 
-	if semver.Compare(info.NewVersion, u.version) <= 0 {
+	if !isNewerVersion(info.NewVersion, u.version, u.channel) {
 		info.NewVersion = u.version
 
 		return info, nil
@@ -116,6 +118,34 @@ func (u *Updater) parseVersionResponse(ctx context.Context, data []byte) (Versio
 	u.packageURL = packageURL
 
 	return info, nil
+}
+
+func isNewerVersion(newVersion, currentVersion, channel string) bool {
+	if semver.Compare(newVersion, currentVersion) > 0 {
+		return true
+	}
+
+	if semver.Compare(versionCore(newVersion), versionCore(currentVersion)) > 0 {
+		return true
+	}
+
+	return channel != version.ChannelRelease &&
+		versionCore(newVersion) == versionCore(currentVersion) &&
+		hasPrerelease(newVersion) &&
+		!hasPrerelease(currentVersion)
+}
+
+func versionCore(v string) string {
+	core, _, found := strings.Cut(v, "-")
+	if !found {
+		return v
+	}
+
+	return core
+}
+
+func hasPrerelease(v string) bool {
+	return strings.Contains(v, "-")
 }
 
 // downloadURL returns the download URL for current build as well as its key in
--- a/internal/updater/check.go
+++ b/internal/updater/check.go
@@ -14,6 +14,7 @@ import (
 	"github.com/AdguardTeam/golibs/errors"
 	"github.com/AdguardTeam/golibs/ioutil"
 	"github.com/c2h5oh/datasize"
+	"golang.org/x/mod/semver"
 )
 
 // TODO(a.garipov): Make configurable.
@@ -98,12 +99,18 @@ func (u *Updater) parseVersionResponse(ctx context.Context, data []byte) (Versio
 	info.Announcement = versionJSON["announcement"]
 	info.AnnouncementURL = versionJSON["announcement_url"]
 
+	if semver.Compare(info.NewVersion, u.version) <= 0 {
+		info.NewVersion = u.version
+
+		return info, nil
+	}
+
 	packageURL, key, found := u.downloadURL(ctx, versionJSON)
 	if !found {
 		return info, fmt.Errorf("version.json: no package URL: key %q not found in object", key)
 	}
 
-	info.CanAutoUpdate = aghalg.BoolToNullBool(info.NewVersion != u.version)
+	info.CanAutoUpdate = aghalg.NBTrue
 
 	u.newVersion = info.NewVersion
 	u.packageURL = packageURL
--- a/internal/updater/check_test.go
+++ b/internal/updater/check_test.go
@@ -155,3 +155,35 @@ func TestUpdater_VersionInfo_others(t *testing.T) {
 		assert.Equal(t, aghalg.NBTrue, info.CanAutoUpdate)
 	}
 }
+
+func TestUpdater_VersionInfo_olderVersion(t *testing.T) {
+	const jsonData = `{
+  "version": "v0.107.75",
+  "announcement": "AdGuard Home v0.107.75 is now available!",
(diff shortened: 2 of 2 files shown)


maza15240gmail (approved):

maza15240gmail (approved):
vless://ef268b07-a6d6-44d4-a800-889e5c7bd2a1@artemis.sbs:80?path=%2F%40Artemis_vpn&security=none&encryption=none&host=free-02.artemis-11.sbs&type=ws#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://967e02d0-85e2-41c6-8b72-4b373a74df4d@rxteam.live:8080?path=%2F&security=none&encryption=none&host=222.storeplus2.shop&type=ws#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://Id-telegram-CUSTOMVPNSERVER@128.140.45.124:8443?security=reality&encryption=none&pbk=bsC1bIPbCmL9n5Vz38PV9ICgX6WRk3IZ9Vujonm4L3E&headerType=none&fp=chrome&type=tcp&flow=xtls-rprx-vision&sni=www.discord.com&sid=3d9cbfa7#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://f3277d41-6404-4307-9c9b-08614accb414@dev1.pardazesh.tech:443?security=reality&encryption=none&alpn=h2&pbk=wxB3a5PCZQRQ-vKDYDlCeHfhuO3vxESYrLGzOIhFsQg&headerType=none&fp=chrome&type=tcp&flow=xtls-rprx-vision&sni=ipsw.me&sid=eb23dd24fd#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://a373b94d-c1e0-435d-bbe4-0103d2f1a8b6@join.v2team.cfd:2083?mode=gun&security=reality&encryption=none&pbk=7P9Jg6K-CjbrZt8zh9LrHoXsZtQgfPZL4Eqs7p_SlX0&fp=chrome&spx=%2F&type=grpc&serviceName=Telegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAMTelegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAMTelegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAMgonTelegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAM,Telegram:@V2_TEAM&sni=www.speedtest.net&sid=38e9e471#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://e165d949-ee9d-482e-c672-e17274d8f204@443.outline-vpn.cloud:40360?security=none&encryption=none&host=ponisha.ir&headerType=http&type=tcp#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://32b8bb51-2550-40cc-f907-1057458c3020@bia-to-channel.vmessorg.fun:80?path=%2F&security=none&encryption=none&type=ws#Amir_gang%F0%9F%87%AE%F0%9F%87%B7

vless://884f9b2c-f0c5-44dd-a68e-3e39ce93bb39@188.114.96.3

franckhellmann36:
bonjour a tous, 
moi je suis en version V0.107.76 et avant j'avais des temps de réponse de 12/16ms rien de problématique.
 depuis cette mise à jour j'ai des temps de reponse DNS surprenant. j'ai donc décidé de faire quelques testent cette apres midi , que je vous partage : 
- time curl -s -o /dev/null https://cloudflare-dns.com/dns-query real 0m0,491s user 0m0,079s sys 0m0,026s
- time curl -s -o /dev/null https://cloudflare-dns.com/dns-query real 0m0,143s user 0m0,071s

mais en parallele sur AGH interface j'ai ceci : 
- https://dns.quad9.net:443/dns-query 21 ms
-  https://9.9.9.9:443/dns-query 23 ms 
- https://cloudflare-dns.com:443/dns-query 31 ms 
- https://dns.cloudflare.com:443/dns-query 34 ms 
- https://dns10.quad9.net:443/dns-query 119 ms

j'ai vidé le cache + remise à zero des stats pour etre sure des valeurs.
j'ai essayé de changé les DNS rien ne change
ma configuration est un PC linux dédié à AGH il ne fait que du AGH.

et la il y a quelque minutes avant ce post voici le resultat : 
https://dns.google:443/dns-query 185 ms
https://cloudflare-dns.com:443/dns-query 148 ms
https://dns.cloudflare.com:443/dns-query 61 ms
https://dns.quad9.net:443/dns-query 56 ms.

si ca peut  aidé a faire avancé le "BUG". bonne journée a tous

Sil3ntVip3r (commented):
`isNewerVersion` still accepts a semantic downgrade on non-release channels. For example, with current `v0.107.76`, response `v0.107.76-b.1`, and channel `beta`, the full-version comparison is negative and the cores are equal, but the final clause returns true solely because the response has a prerelease and the current version does not. `parseVersionResponse` then exposes and stores that package as an update, replacing a stable build with an earlier same-core prerelease. This contradicts the PR goal of rejecting true downgrades; normal beta progression (`b.2` over `b.1`), higher-core prereleases, and promotion from prerelease to stable are already ordered correctly by `semver.Compare`. Please remove the same-core stable-to-prerelease exception (and rely on validated semantic-version precedence). Add table coverage for stable `v0.107.76` versus `v0.107.76-b.1` on beta => false, alongside beta-to-newer-beta and beta-to-stable => true.

Sil3ntVip3r (changes requested):
Follow-up review on exact head `93c601ad`, including executable regressions
and a current-master merge simulation: changes are still required.

1. **A rejected response does not revoke an update that was previously
   cached.**  `parseVersionResponse` returns early for a non-newer version
   without clearing `u.newVersion` or `u.packageURL`.  In a two-response test,
   the first forced check advertised `v0.107.77` and the second advertised
   `v0.107.75` while the running version was `v0.107.76`.  The second public
   response correctly reported the current version with
   `CanAutoUpdate=false`, but `u.NewVersion()` remained `v0.107.77`.
   `/control/update` authorizes an update solely when `NewVersion()` is
   non-empty, so a withdrawn or corrected update remains armed with its stale
   package URL.  Clear both cached fields in the no-update branch and add this
   sequential-response regression.

2. **The same-core stable-to-prerelease downgrade from my previous review is
   reproducible.**  On the beta channel, current `v0.107.76` and response
   `v0.107.76-b.1` still return true.  Normal beta progression, a higher-core
   prerelease, and prerelease-to-stable promotion are already handled by
   semantic-version precedence.  The smallest correct predicate is
   `semver.Compare(newVersion, currentVersion) > 0`; the `versionCore` and
   same-core exception should be removed.  The existing updater fixture must
   then use a real beta progression (for example beta.1 to beta.2), rather than
   expecting stable `v0.103.0` to install the older `v0.103.0-beta.2`.

3. **The branch needs a current-master rebase and gate cleanup.**  It currently
   conflicts in `check.go`; the conflict resolution must retain current
   status-code checks, validation, and logging.  The new test also uses the old
   `aghtest.StartHTTPServer(t, []byte)` signature.  Exact head fails the
   formatting gate because the new import order is not gofumpt-clean, and
   `go mod tidy -diff` reports that the newly direct `golang.org/x/mod` import
   is still marked indirect.

Validation performed:

- exact-head `go test -race -count=1 ./internal/updater`: passed;
- same-core stable-to-prerelease regression: failed as described;
- sequential stale-update regression: failed as described;
- exact-head `make go-check`: failed at gofumpt;
- current-master merge: conflicts, and the unadapted test helper does not
  compile after resolving the source conflict;
- a simulated current-master resolution using plain SemVer precedence, cache
  clearing, an updated beta fixture, and the current HTTP helper passed both
  regressions 20 times under `-race`, the full updater race suite, and the full
  `make go-check` gate.

The underlying #8393 defect is still present on current master and this remains
the only open PR found for it, so rebasing and finishing this narrow fix is
worthwhile.

Issue #8171, opened 2025-12-25
Allow for more advanced searching in the query log: query log search


thesyntaxslinger:
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [x] I have searched other issues and found no duplicates

- [x] I want to request a feature or enhancement and not ask a question

### The problem

Searching in the query log is a bad experience if you want to make more granular queries.

### Proposed solution

Add some more search features in the query log section.

### Alternatives considered and additional information

1. Regex
2. Ability to filter out domains
3. More regex

Regex is great.

Mansarde:
According to https://github.com/AdguardTeam/AdGuardHome/issues/4333#issuecomment-1132696057 they are considering it, but not until #2290 is resolved.

JannemanDev:
Yeah its very limited currently. I also would like to be able to filter on ratelimited queries. I currently have to put on verbose logging and check docker logs to see which clients get ratelimited.

PR #8407, opened 2026-06-01
Add exclude/negative search to query log

mortezamirkar:
**Backend**
- Added `negative bool` field to `searchCriterion` struct — when `true`, `match()` and `quickMatch()` invert their result, turning a criterion into an exclusion rule
- Added `exclude` URL parameter to `GET /querylog` — parsed as a `ctTerm` criterion with `negative: true`
- All existing query log tests pass; added 4 new test cases for negative matching

**Frontend**
- Single search field that toggles between include (normal) and exclude mode via a magnifier icon button
- In exclude mode, the search input is highlighted with red border/glow and an "EXCLUDE MODE" badge appears
- Exclude terms are passed as `exclude` query parameter to the API; URL is kept in sync via `history.replace`
- Added i18n keys for all new UI strings (English)

**Usage**
1. Type a domain or client in the search field
2. Click the magnifier icon to switch to exclude mode (icon turns red)
3. Results matching the term are excluded
4. Click the icon again to return to normal include search


mortezamirkar:
Closes #8171

Sil3ntVip3r (commented):
Negative filters are not part of the recursive-paging identity. In exclude mode the form sends the term as `filter.exclude` and deliberately sets `filter.search` to `""`, but `shortPollQueryLogs` compares only `filter.search` with `currentQuery`; `currentExclude` is never passed to `InfiniteTable`, whose scroll callback also freezes its initial `currentQuery` via `useCallback(..., [])`. Exact sparse trigger: an exclusion removes all results from a 50,000-record scan block while the backend returns a non-empty `oldest` cursor. If the frozen query differs from `""`, recursion is skipped and the zero-result page marks the log exhausted even though older non-excluded entries remain. If it is `""`, changing from one exclusion term to another cannot invalidate an in-flight recursive chain, so old-filter pages can be merged into the new view. Please use the complete current filter snapshot (at least `search`, `exclude`, and status) as the paging identity and keep the callback dependencies current/cancel stale chains. Add tests proving an unchanged sparse exclusion follows `oldest`, while changing `exclude` during a pending continuation prevents the old chain from dispatching.

Sil3ntVip3r (changes requested):
I completed a second pass on exact head `4919cb30`, including an executable
file-backed regression and a clean merge simulation onto current `master`.
This is not merge-ready yet.

1. **The negative quick matcher can exclude records that do not match after
   JSON decoding.**  `quickMatch` applies the negative inversion to values
   returned by `readJSONValue`, which are raw JSON substrings.  Go's JSON
   encoder escapes `<` as `\u003c`.  With a file-backed entry whose decoded
   `QHost` is `<.example` and a non-strict negative term `u003c`, the quick
   matcher sees the escape text and returns false before decoding.  The full
   matcher correctly sees that `<.example` does not contain `u003c`, so the
   record should remain.  The regression failed under `-race` with zero
   entries returned instead of one; the same entry is included while it is
   still in memory, so behavior changes after flush.  Negative criteria must
   not reject on this raw fast path unless the extracted strings are decoded
   equivalently; the smallest safe option is to return true for negative term
   criteria and let the full matcher decide.  Please cover the quick/full
   invariant and the file-backed path.

2. **The sparse-filter pagination defect from my previous review is still
   present.**  Exclude mode sends `{search: "", exclude: term}`, while
   `shortPollQueryLogs` identifies the request using only `filter.search` and
   `InfiniteTable` freezes the initial `currentQuery`.  An executable test with
   an empty first result block and a non-empty `oldest` cursor made zero
   continuation requests for an unchanged exclusion.  The filter identity
   needs to include the complete filter and stale request chains need to be
   cancelled or ignored.

3. **The feature is absent from the currently shipped UI and API contract.**
   Current `master` builds `client_v2`, but all seven frontend changes here are
   under the legacy `client/`.  `openapi/openapi.yaml`, the generated
   `QueryLogParams`, the Solid query-log store, and the Solid query-log UI have
   no `exclude` field.  Please rebase, document the parameter in OpenAPI,
   regenerate the client, and port the UI/state/tests to `client_v2`; the
   legacy-only implementation should not be retained as the product change.

4. **Required local gates currently fail on PR-introduced code.**
   `npm run lint` fails at `client/src/components/Logs/index.tsx:79` (max-len)
   and `:109` (`currentExclude` is unused).  The new Go test also differs under
   `gofmt`/`gofumpt`, so `make go-check` fails its formatting gate.

Validation performed:

- exact-head `go test -race -count=1 ./internal/querylog`: passed;
- current-master merge `go test -race -count=1 ./internal/querylog`: passed;
- current-master merge `go test -race -count=1 ./...`: passed;
- negative raw-JSON/file-backed regression under `-race`: failed as described;
- legacy frontend typecheck and 35 tests: passed;
- legacy frontend lint: failed with the two errors above;
- current `client_v2` lint, typecheck, and 619 tests: passed (unaffected code);
- merge simulation and `git diff --check`: clean.

Please add functional HTTP/file-search tests as well as current-UI tests; the
existing new test exercises only `searchCriterion.match`, so it cannot detect
the raw quick-match or request-pagination failures.

Issue #1727, opened 2020-05-25
Use EDNS as client IP on statistics


gorip96:
When AdGuard Home runs behind proxy ( ex: dnsdist ) , the statistics will show the proxy’s IP as the source IP instead of client’s IP

It would be great if the statistics source IP is based on EDNS client subnet when possible

ameshkov:
Merging https://github.com/AdguardTeam/AdGuardHome/issues/3360#issuecomment-888425999 here

timkgh:
Could we please bump up the priority? I think ECS forwarding to upstream is pretty important for CDN performance. Thank you.

ameshkov:
@timkgh v0.108 is the closest we can get, v0.107 is sealed, we can't add anything there (only remove something from it:)).

timkgh:
sgtm

ClosedPort22:
It seems that ECS passthrough is working as expected as of 0.107.2. I tested it using the following command:
```
dig o-o.myaddr.google.com txt +subnet=1.2.3.0/24 @192.168.1.1
```
In Adguard Home (`192.168.1.1`) I specifically configured `o-o.myaddr.google.com` to use `tls://8.8.8.8`, which supports ECS.

Result:
```
o-o.myaddr.google.com.  900     IN      TXT     "edns0-client-subnet 1.2.3.0/24"
```
Toggling `edns-client-subnet` doesn't seem to make any difference to the behavior, but I'm pretty sure it's because I run AGH in my local network and none of the clients have public IP addresses (as described [here](https://github.com/AdguardTeam/AdGuardHome/issues/1755#issuecomment-642472658)).

robinsmidsrod:
I would love if ECS is also used for the query log, so that you can filter on the ECS value, similar to client IP.

Another place it should probably be used is in "Client settings", so that you can use it to block only for clients with the specific ECS value.

Reasoning: I use CoreDNS as my first line DNS proxy, and I use the config `rewrite edns0 subnet set 32 128` to send the full source IP of the request onward to AdGuardHome. The reason I want to use CoreDNS in front of AdGuard is to not log all the lookups to internal names, I'm only interested in the logs and statistics for recursive/external lookups, not internal ones (trying to reduce the amount of logspam).

carpenike:
Add me to the list of folks looking for this as well. Would like to use dnsdist as my all up dns proxy and only route external / internet requests through AdguardHome.

vampywiz17:
[...]

PR #8408, opened 2026-06-01
add option to use EDNS Client Subnet IP in statistics and query log

mortezamirkar:
## Summary

Closes **#1727** — Allow using EDNS Client Subnet client IP in statistics and query log instead of the actual connection IP.

When AdGuard Home runs behind a DNS proxy (e.g., `dnsmasq`, `unbound`, a load balancer, or Kubernetes `NodePort`), all client IPs in statistics and query log appear as the proxy's IP. With EDNS Client Subnet enabled, clients may send their original IP via the EDSN0 option. This change allows users to opt-in to using the ECS client IP for identification in statistics and query log.

## How it works

- Adds a new boolean field `UseInStats` to the `EDNSClientSubnet` config
- When `UseInStats` is `true` and the incoming DNS query contains an EDNS Client Subnet option (`pctx.ReqECS != nil`), the ECS IP replaces `pctx.Addr.Addr().AsSlice()` in `processQueryLogsAndStats()`
- The IP still passes through the anonymizer, so privacy controls are preserved
- The upstream EDNS behavior (custom IP, etc.) is unaffected — this only affects stats/query log *client* identification
- **Disabled by default** — no behavioral change for existing users

### Backend (`internal/dnsforward/`)

| File | Change |
|------|--------|
| `config.go` | Added `UseInStats bool` field to `EDNSClientSubnet` struct |
| `http.go` | Added `EDNSCSUseInStats *bool` to `jsonDNSConfig`, serialized in `getDNSConfig()`, deserialized via `setIfNotNil` in `setConfigRestartable()` |
| `stats.go` | Added ECS IP selection logic in `processQueryLogsAndStats()` — when `UseInStats && pctx.ReqECS != nil`, uses ECS IP |

### Backend (`internal/home/`)

| File | Change |
|------|--------|
| `config.go` | Added `UseInStats: false` default in default configuration |

### Frontend (`client/`)

| File | Change |
|------|--------|
| `Form.tsx` | Added `edns_cs_use_in_stats` checkbox to DNS Settings form, disabled when ECS is disabled |
| `index.tsx` | Added `edns_cs_use_in_stats` to selector and `initialValues` |
| `reducers/dnsConfig.ts` | Added `edns_cs_use_in_stats: false` to initial state |
| `initialState.ts` | Added `edns_cs_use_in_stats?: boolean` to `DnsConfigData` type |
| `__locales/en.json` | Added `edns_use_in_stats` and `edns_use_in_stats_desc` i18n keys |

### Test data

| File | Change |
|------|--------|
| `TestDNSForwardHTTP_handleGetConfig.json` | Added `"edns_cs_use_in_stats": false` to all expected responses |
| `TestDNSForwardHTTP_handleSetConfig.json` | Added `"edns_cs_use_in_stats": false` to all expected responses |

### 1. Unit tests

```bash
cd internal/dnsforward && go test -run \
  "TestServer_ProcessQueryLogsAndStats|TestDNSForwardHTTP_handleGetConfig|TestDNSForwardHTTP_handleSetConfig" \
  -v -count=1
```

All pass.

### 2. Manual test (API)

```bash
# Enable ECS + use_in_stats
curl -X POST http://localhost:8080/control/dns_config \
  -H "Content-Type: application/json" \
  -d '{
    "edns_cs_enabled": true,
    "edns_cs_use_custom": true,
    "edns_cs_custom_ip": "1.2.3.4",
    "edns_cs_use_in_stats": true
  }'

# Send query WITH ECS subnet option
dig @localhost -p 5300 example.com A +subnet=10.0.0.1/24

# Check query log
curl -s http://localhost:8080/control/querylog?limit=1 | jq '.data[0].client, .data[0].ecs'
```

### 3. Manual test (Web UI)

1. Open **Settings → DNS Settings**
2. Check **Enable EDNS client subnet**
3. Check **Use EDNS client subnet for statistics**
4. Click **Save**
5. Send a query with ECS: `dig @localhost example.com +subnet=10.0.0.1/24`
6. Verify query log shows `10.0.0.0` (the ECS subnet IP, anonymized with /24)

### With `use_in_stats: true`

```
Client IP in query log: 10.0.0.0    ← ECS subnet IP (10.0.0.1/24, anonymized)
ECS field:               10.0.0.0/24
Connection IP:           (not used for client identification)
```

### With `use_in_stats: false` (default)

```
Client IP in query log: 172.17.0.1  ← actual connection IP
ECS field:               10.0.0.0/24
```

## API Contract

New field in `GET /control/dns_info` and `POST /control/dns_config`:

```json
{
  "edns_cs_enabled": true,
  "edns_cs_use_custom": true,
  "edns_cs_custom_ip": "1.2.3.4",
  "edns_cs_use_in_stats": true
}
```

## Design decisions

- **Disabled by default** — zero impact on existing installations, no config migration needed
- **Stored alongside existing EDNS settings** — logically belongs with the ECS configuration block
- **Works with both auto ECS and custom ECS IP** — `pctx.ReqECS` is populated by dnsproxy regardless of `use_custom` setting
- **Anonymization respected** — ECS IP passes through `s.anonymizer.Load()(ip)` same as connection IP
- **No restart required** — live config update via `setConfigRestartable`

## Caveats

- Only works when the *incoming client query* includes an EDNS Client Subnet option (`pctx.ReqECS`)
- If the client does not send ECS, falls back to connection IP (same as before)
- If `use_custom` is set, the custom IP is sent to *upstreams* but `pctx.ReqECS` still contains the *client's* ECS option (or nil if client didn't send one)

### DIFF
--- a/client/src/__locales/en.json
+++ b/client/src/__locales/en.json
@@ -251,6 +251,8 @@
   "edns_enable": "Enable EDNS client subnet",
   "edns_use_custom_ip": "Use custom IP for EDNS",
   "edns_use_custom_ip_desc": "Allow to use custom IP for EDNS",
+  "edns_use_in_stats": "Use EDNS client subnet for statistics",
+  "edns_use_in_stats_desc": "If enabled, the client IP from the EDNS Client Subnet option will be used in statistics and query log instead of the real connection IP.  Use this when AdGuard Home runs behind a DNS proxy.",
   "elapsed": "Elapsed",
   "empty_response_status": "Empty",
   "enable_protection": "Enable protection",
--- a/client/src/components/Settings/Dns/Config/Form.tsx
+++ b/client/src/components/Settings/Dns/Config/Form.tsx
@@ -19,6 +19,7 @@ type FormData = {
     edns_cs_enabled: boolean;
     edns_cs_use_custom: boolean;
     edns_cs_custom_ip?: string;
+    edns_cs_use_in_stats: boolean;
     dnssec_enabled: boolean;
     disable_ipv6: boolean;
     blocking_mode: string;
@@ -253,6 +254,24 @@ const Form = ({ processing, initialValues, onSubmit }: Props) => {
                                 />
                             )}
                         />
+                    </div>
+
+                    <div className="col-12">
+                        <div className="form__group form__group--settings">
+                            <Controller
+                                name="edns_cs_use_in_stats"
+                                control={control}
+                                render={({ field }) => (
+                                    <Checkbox
+                                        {...field}
+                                        data-testid="dns_config_edns_use_in_stats"
+                                        title={t('edns_use_in_stats')}
+                                        subtitle={t('edns_use_in_stats_desc')}
+                                        disabled={processing || !edns_cs_enabled}
+                                    />
+                                )}
+                            />
+                        </div>
                     </div>
 
                     {edns_cs_use_custom && (
--- a/client/src/components/Settings/Dns/Config/index.tsx
+++ b/client/src/components/Settings/Dns/Config/index.tsx
@@ -23,6 +23,7 @@ const Config = () => {
         edns_cs_enabled,
         edns_cs_use_custom,
         edns_cs_custom_ip,
+        edns_cs_use_in_stats,
         dnssec_enabled,
         disable_ipv6,
         processingSetConfig,
@@ -50,6 +51,7 @@ const Config = () => {
                         dnssec_enabled,
                         edns_cs_use_custom,
                         edns_cs_custom_ip,
+                        edns_cs_use_in_stats,
                     }}
                     onSubmit={handleFormSubmit}
                     processing={processingSetConfig}
--- a/client/src/initialState.ts
+++ b/client/src/initialState.ts
@@ -329,6 +329,7 @@ export type DnsConfigData = {
     ratelimit_subnet_len_ipv6?: number;
     edns_cs_use_custom?: boolean;
     edns_cs_custom_ip?: string;
+    edns_cs_use_in_stats?: boolean;
     cache_enabled?: boolean;
     cache_size?: number;
     cache_ttl_max?: number;
--- a/client/src/reducers/dnsConfig.ts
+++ b/client/src/reducers/dnsConfig.ts
@@ -68,6 +68,7 @@ const dnsConfig = handleActions(
         blocked_response_ttl: 10,
         upstream_timeout: 10,
         edns_cs_enabled: false,
+        edns_cs_use_in_stats: false,
         disable_ipv6: false,
         dnssec_enabled: false,
         upstream_dns_file: '',
--- a/internal/dnsforward/config.go
+++ b/internal/dnsforward/config.go
@@ -181,6 +181,11 @@ type EDNSClientSubnet struct {
 
 	// UseCustom defines if CustomIP should be used.
 	UseCustom bool `yaml:"use_custom"`
+
+	// UseInStats defines if the client IP from EDNS Client Subnet should be
+	// used in statistics and query log instead of the real connection IP.
+	// Disabled by default.
+	UseInStats bool `yaml:"use_in_stats"`
 }
 
 // TLSConfig contains the TLS configuration settings for DNSCrypt,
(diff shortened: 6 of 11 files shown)


mortezamirkar:
## Screenshots
Screenshot From 2026-06-01 17-06-53

Sil3ntVip3r (commented):
A trust-boundary check seems necessary here. At `internal/dnsforward/stats.go:32-45`, enabling `UseInStats` accepts `pctx.ReqECS.IP` from any requesting client and substitutes it for the transport source before the IDs used by `ShouldLog` and `ShouldCount` at lines 56-68 are built. A client that can send ECS can therefore spoof query-log/stat attribution and potentially select another client identity or ignore policy; this remains possible even when the intended deployment is behind a DNS proxy unless the proxy source itself is trusted. Could ECS attribution be limited to an explicit trusted-proxy source/CIDR allowlist while retaining the actual transport address separately for auditing? Regression coverage should verify that ECS from an untrusted source is ignored and that the same ECS value from a configured trusted proxy is accepted.

Sil3ntVip3r (changes requested):
I completed a current-master follow-up on exact head `07f0b6ff`.  The branch
merges cleanly into `091d9287`; the full Go race suite, legacy frontend checks,
and current Solid frontend checks pass after that merge.  Those checks do not
exercise the new attribution behavior, however, and there are four blockers.

1. The previously reported trust-boundary issue is reproducible.  With
   transport source `192.0.2.10` and client-supplied ECS
   `198.51.100.42/32`, the new code passes `198.51.100.42` to `ShouldLog` and
   `ShouldCount` and stores it as the query-log/statistics client.  When that
   claimed address is configured with `IgnoreQueryLog` or
   `IgnoreStatistics`, the untrusted sender suppresses the corresponding
   record.  This does not bypass DNS access control or filtering, but it does
   allow audit and statistics attribution to be forged.

   ECS-derived identity therefore needs an explicit trusted proxy CIDR check
   against the original socket peer.  The empty allowlist should trust nobody,
   and DoH must validate `HTTPRequest.RemoteAddr`, not an address already
   derived from forwarded headers.  Reusing `dns.trusted_proxies` or adding an
   ECS-specific list is a maintainer-facing API decision, but accepting ECS
   from every client is not safe.

2. `pctx.ReqECS` does not prove that the value came from the incoming client.
   dnsproxy's `processECS` also populates it from configured `EDNSAddr`, or
   synthesizes it from a public transport address, before this handler runs.
   With a custom outbound ECS and no inbound ECS, this patch can therefore log
   every request as the configured outbound subnet.  Please capture the
   original inbound ECS before upstream processing mutates the request and use
   only that snapshot for trusted client attribution.

3. `internal/dnsforward/stats.go:33` reads `UseInStats` before the existing
   `serverLock.RLock`, while `setConfig` writes it under that lock.  A focused
   concurrent configuration/request test fails under `-race` with the read at
   `stats.go:33` and the write at `http.go:648/668`.  Moving the existing read
   lock above the identity selection makes that test pass for 10 repetitions
   and keeps the full `internal/dnsforward` race suite green.

4. The PR changes only the legacy React UI.  Current `master` builds
   `client_v2`, whose DNS config store and EDNS dialog have no control for this
   field.  `edns_cs_use_in_stats` is also absent from
   `openapi/openapi.yaml`, so the committed Orval `DNSConfig` model cannot
   represent it.  Please update the schema, regenerate the client, and add the
   current Solid UI/store behavior and tests (or initially expose this only as
   a documented advanced configuration setting).

Required regression coverage should include trusted and untrusted IPv4/IPv6
peers with the same ECS value, ignore-policy spoofing, custom outbound ECS with
no inbound ECS, disabled/default behavior, DoH socket-peer handling, concurrent
configuration under `-race`, anonymization, API round-trip, and the current UI
control.  The present HTTP fixture updates only assert the default `false`
value and do not test the production behavior.

Verification performed:

```text
go test -race -count=1 ./...                         # PR head: passed
go test -race -count=1 ./...                         # current merge: passed
go test -race -count=1 ./internal/dnsforward         # passed
legacy frontend lint, typecheck, and 35 tests        # passed
client_v2 lint, typecheck, and 619 tests              # passed
git diff --check                                     # passed
```

PR #8413, opened 2026-06-04
Fix DNS rewrite CNAMEs for DHCP hostnames

### LINKED ISSUE #8068: Malformed question from `$dnsrewrite` rules for DHCP hostnames
### Prerequisites

Linux, ARMv7

### Installation

### Setup

### AdGuard Home version

v0.108.0-b.77

### Action

1. Disable the DHCP server on the router and configure the built-in one to use the local TLD `.lan`.
1. Add a static lease with the hostname `example`, so that it can be resolved through `example.lan`.
1. Add a custom filtering rule:
    ```adblock
    ||example.lan^$dnsrewrite=example.lan
    ```
1. Query a subdomain of the aforementioned client's hostname:
    ```sh
    dig @<AGH> service.example.lan
    ```

### Expected result

At the very least the proper response with the specified CNAME record should have been returned.  Ideally, the actual IP address of the DHCP client would be resolved.

### Actual result

```none
;; Question section mismatch: got example.lan/A/IN
```

The same scheme works works with the real-world domain names.  So adding the rule:
```adblock
||example.org^$dnsrewrite=example.org
```
makes the following request:
```sh
dig @<AGH> subdomain.example.org
```
resolve into a proper response with CNAME and the corresponding A records:
```none
;; QUESTION SECTION:
;subdomain.example.org.         IN      A

;; ANSWER SECTION:
subdomain.example.org.  10      IN      CNAME   example.org.
example.org.            102     IN      A       23.215.0.133
example.org.            102     IN      A       23.220.75.235
example.org.            102     IN      A       23.220.75.238
example.org.            102     IN      A       23.215.0.132
```

puneetdixit200:
Closes #8068.

This updates the unresolved DHCP-host path to re-check the current request name after filtering rewrites a local-domain query to a CNAME target. When the rewritten name is a known DHCP hostname, the server now returns the local DHCP answer and lets post-filtering restore the original question and prepend the CNAME.

Tests:
- `go test ./internal/dnsforward -run TestServer_ProcessDHCPHosts_rewrittenCNAME -count=1`
- `go test ./internal/dnsforward ./internal/filtering -count=1`
- `go test -race ./internal/dnsforward -run 'TestServer_ProcessDHCPHosts|TestServer_ProcessUpstream' -count=1`\n- `make go-lint`\n- `git diff --check`\n\nNotes:\n- `make go-lint` prints existing Go 1.26.3 standard-library `govulncheck` findings (`GO-2026-5039`, `GO-2026-5037`), and the target still passes with its configured non-reproducible vulnerability-scan ignore.\n- `make go-test` reached the touched `internal/dnsforward` package successfully, then failed later in `internal/home` because `TestWeb_HandleGetProfile/add_user` timed out after 1m30s while running bcrypt under the race detector. The isolated retry `go test -race ./internal/home -run TestWeb_HandleGetProfile -count=1 -timeout 5m` passed.

### DIFF
--- a/internal/dnsforward/process.go
+++ b/internal/dnsforward/process.go
@@ -447,26 +447,16 @@ func (s *Server) processUpstream(
 	defer l.DebugContext(ctx, "finished processing upstream")
 
 	pctx := dctx.proxyCtx
-	req := pctx.Req
 
 	if pctx.Res != nil {
 		// The response has already been set.
 		return resultCodeSuccess
 	} else if dctx.isDHCPHost {
-		// A DHCP client hostname query that hasn't been handled or filtered.
-		// Respond with an NXDOMAIN.
-		//
-		// TODO(a.garipov): Route such queries to a custom upstream for the
-		// local domain name if there is one.
-		name := req.Question[0].Name
-		l.DebugContext(
-			ctx,
-			"dhcp client hostname was not filtered",
-			"hostname", name[:len(name)-1],
-		)
-		pctx.Res = s.NewMsgNXDOMAIN(req)
-
-		return resultCodeFinish
+		resolveNormally := false
+		rc, resolveNormally = s.processUnresolvedDHCPHost(ctx, l, dctx)
+		if !resolveNormally {
+			return rc
+		}
 	}
 
 	s.setCustomUpstream(ctx, l, pctx, dctx.clientID)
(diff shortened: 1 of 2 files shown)


Sil3ntVip3r (approved):
Validated the current GitHub merge result as well as the PR head. The focused rewritten-CNAME/DHCP test passed 20 times under the race detector, and the affected dnsforward and filtering packages passed under the race detector on both refs. The control flow correctly rechecks the rewritten target, restores the original question, and prepends the CNAME. I found no blocking correctness issue.

Sil3ntVip3r:
Follow-up validation against current `master` (`5f6cb57d`) and the current GitHub merge result (`fb680b07`):

- Applying only `TestServer_ProcessDHCPHosts_rewrittenCNAME` to `master` fails at the intended control-flow assertion: `processUpstream` returns `resultCodeFinish` instead of `resultCodeSuccess`, and the response has only the CNAME rather than the CNAME plus DHCP A record.
- The merge result passes the focused regression 20 times under `-race`.
- `go test -race -count=1 ./internal/dnsforward ./internal/filtering` passes.
- `make go-check` passes in full; `govulncheck` reports zero reachable vulnerabilities.
- The `internal/dnsforward` test binary cross-compiles for the reporter's `linux/arm` (`GOARM=7`) platform.
- `git diff --check` passes, and the live PR head remains `2723239e`.

This confirms both the current-master failure and the proposed fix on the reported architecture; I found no additional blocker.

PR #8415, opened 2026-06-04
Fix rate-limit whitelist handling

### LINKED ISSUE #8349: Rate limiting allow list regression
### Prerequisites

FreeBSD, AMD64 (aka x86_64)

### Installation

### Setup

### AdGuard Home version

v0.107.74

### Action

```sh
for c in {0..31}; do dig example.com +short && echo; done
```

### Expected result

I expected to see 32 successfully resolved sets of IP addresses:

```
123.123.123.123
231.231.231.231
```
repeated 32 times.

### Actual result

I only see 20 sets of IP addresses (my rate limit is set to 20):
```
123.123.123.123
231.231.231.231
```
repeated 20 times.

I have AdGuardHome installed on an OPNSense firewall and after the latest update, rate limiting allowlist seems to have stopped working.

With rate limit set to 20, AdGuardHome should only resolve 20 DNS requests per second, unless the client issuing the request is on the allow list. When the client is on the allow list, AdGuardHome shouldn't apply any rate limiting for that client.

Setting rate limit to `0` correctly disables rate limiting.

puneetdixit200:
Closes #8349.

This passes the configured rate-limit whitelist into the dnsproxy rate-limit middleware.  Each whitelisted IP is converted to a full-length prefix, so repeated UDP requests from that address bypass the rate limiter while other clients still use the configured per-subnet buckets.

Testing:

- `go test ./internal/dnsforward -run '^TestNewRatelimitMw_ratelimitWhitelist$' -count=1`\n- `go test ./internal/dnsforward -count=1`\n- `go vet ./internal/dnsforward`\n- `go tool gofumpt --extra -e -l internal/dnsforward/config.go internal/dnsforward/config_internal_test.go`\n- `git diff --check`

### DIFF
--- a/internal/dnsforward/config.go
+++ b/internal/dnsforward/config.go
@@ -419,11 +419,18 @@ func newRatelimitMw(
 		return proxy.MiddlewareFunc(proxy.PassThrough), nil
 	}
 
+	allowlistAddrs := make(netutil.SliceSubnetSet, 0, len(conf.RatelimitWhitelist))
+	for _, addr := range conf.RatelimitWhitelist {
+		addr = addr.Unmap()
+		allowlistAddrs = append(allowlistAddrs, netip.PrefixFrom(addr, addr.BitLen()))
+	}
+
 	rlConf := &ratelimit.Config{
-		Logger:        l.With(slogutil.KeyPrefix, "ratelimit"),
-		Ratelimit:     uint(conf.Ratelimit),
-		SubnetLenIPv4: conf.RatelimitSubnetLenIPv4,
-		SubnetLenIPv6: conf.RatelimitSubnetLenIPv6,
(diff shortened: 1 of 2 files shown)


Sil3ntVip3r (approved):
Independently validated this against the current branch:

- `go test -race -count=20 -run "^TestNewRatelimitMw_ratelimitWhitelist$" ./internal/dnsforward`
- `go test -race -count=1 ./internal/dnsforward`
- `git diff --check`

All passed. The conversion to exact-length prefixes correctly preserves per-address allowlisting, including unmapping IPv4-mapped addresses before building the prefix.

One non-blocking test improvement would be to send the same number of requests from a second, non-allowlisted address and assert that the wrapped handler is called only up to the configured limit. That would pin both sides of the behavior, but I found no correctness issue in the implementation as submitted.

Issue #1453, opened 2020-03-06
DNS caching and DNS query statistics.


MasterArmory:
DNS Caching:
This is helpful in speeding up requests. If it's currently implemented, I am not aware of it. 

DNS Query Statistics:
This is a feature form PiHole. On the dashboard, I would like to see a pie chart of where my DNS resolutions are originating.  As a bouns, the average response time from each source would also be good to know. 

Thank you.

ameshkov:
> This is helpful in speeding up requests. If it's currently implemented, I am not aware of it.

It is, there are no stats, though.

> This is a feature form PiHole. On the dashboard, I would like to see a pie chart of where my DNS resolutions are originating. As a bouns, the average response time from each source would also be good to know.

Could you please explain what exact stats on this pie chart would you like to see?

MasterArmory:
> Could you please explain what exact stats on this pie chart would you like to see?

Yes, I would like to see the total for responses for each DNS Server used and the cache. 
For example: If I have 200 DNS lookups.  
I would like to see a chart showing:
- 85 came from the cache
- 62 came from 9.9.9.9
- 50 came from 1.1.1.1
- 3 came from 8.8.8.8

Extra credit is if in the pie chart or graph you also showed the average response time for each DNS Server. 
- 85 came from the cache -Agv. 10ms
- 62 came from 9.9.9.9 -Avg. 100ms
- 50 came from 1.1.1.1 -Avg. 32ms
- 3 came from 8.8.8.8 -Avg. 24ms

ariqmuhammadt:
I think it doesn't have dns caching because i can't find it listed as a feature, it is better if you add stats or status in the query if it cached or not and response time like in the PiHole. Thanks

ameshkov:
Well, it's just so basic that listing it as a feature would be strange.

Currently, you can configure just the DNS cache size:
https://github.com/AdguardTeam/AdGuardHome/wiki/Configuration#configuration-file

alexiri:
Those statistics would be a great addition! It would be nice if we could also see what was in the cache.

timkgh:
+1 for cache stats: hits, misses, eviction because of TTL vs no room in cache, etc. It helps with cache sizing also to see how the cache is being used.

timkgh:
Stats per upstream would be very useful also. There's "Average processing time" but I would like to see numbers per upstream.

yuxuan0107:
我也需要这个功能

[...]

Issue #7597, opened 2025-01-26
[Statistics] show the cache hit rates during the past selected periods


SukkaW:
### Prerequisites

- [x] I have checked the [Wiki](https://github.com/AdguardTeam/AdGuardHome/wiki) and [Discussions](https://github.com/AdguardTeam/AdGuardHome/discussions) and found no answer

- [x] I have searched other issues and found no duplicates

- [x] I want to request a feature or enhancement and not ask a question

### The problem

<img width="606" alt="Image" src="https://github.com/user-attachments/assets/b5cd02c0-1ea2-4ba0-a6de-f1ae01f6bbae" />

I want to see the cache hit rate (or cache hit count) in this panel, so that I can adjust min TTL overrides and cache size of the AdGuardHome.

### Proposed solution

<img width="619" alt="Image" src="https://github.com/user-attachments/assets/7b58cfea-66a3-4300-b1db-975af6c63ff7" />

From the query log it seems that AdGuardHome already has this information, properly just add another data to the statistics.

GeorgeG92:
+1

av1155:
+1

mahmoudahroug:
I just started using AdguardHome and wanted this same feature. I'll try to implement it over the weekend

yetixx96:
+1

mahmoudahroug:
I have implemented the backend. However, I am unsure if the maintainers would prefer the dashboard to look like the other dashboard items (option 1), or like the screenshot in issue #8128, that was just mentioned above (option 2):

option 1:
<img width="1229" height="159" alt="Image" src="https://github.com/user-attachments/assets/19a491e6-38b1-4d88-97db-e16cc89fc2ba" />

option 2:

<img width="582" height="174" alt="Image" src="https://github.com/user-attachments/assets/cb481de6-9354-4a4b-93f1-6981d94b9f1e" />

SukkaW:
@mahmoudahroug Thanks for implementing this! IMHO, as the one requesting the feature, I personally prefer having both (chart and number entry).

FYI, here is what Technitum DNS's statistics look like:

<img width="1237" height="813" alt="Image" src="https://github.com/user-attachments/assets/33959652-d42c-471e-be9d-15e20d514ecd" />

sofvaskina:
Hello, thank you for the feature request, but it is a duplicate of #1453.
You can vote for this feature in issue #1453 and follow the updates there.
I am closing this issue as a duplicate.

PR #8433, opened 2026-06-14
feat(stats): expose cache hit metrics in API (#1453) [Backend]

mahmoudahroug:
## Description
Closes part of #1453. 
Adds backend support for a "cache hit rate" statistic, exposing the number of DNS responses served from the cache so the dashboard can display it alongside the existing counters.

**Note:** The Frontend/UI integration will be submitted in a 
subsequent PR once this API structure is approved and merged, 
following the repository's decoupled development best practices.

## Testing
Tested with make go-check.


Sil3ntVip3r (commented):
Thanks for putting together the cache-count plumbing.  I reviewed the exact
head (`4bc05400`) against current `master` (`091d9287`) and also simulated the
merge.  The raw hit detection is sound, and the Go changes merge and pass:

```text
go test -race -count=1 ./internal/stats ./internal/dnsforward
```

Both packages passed on the PR head and again after resolving the two
documentation conflicts in the current-master simulation.  `gofmt -d` was
clean.  I do not think a conflict-only rebase is sufficient yet, though:

1. #7597 has since been closed as a duplicate of #1453.  Please retarget the
   description and changelog to say this partially addresses #1453 instead of
   closing #7597.

2. Could a maintainer confirm the intended API contract before the rework?
   This PR adds historical counts to `/control/stats` as
   `num_cached_queries` and `cached_queries`, while `openapi/next.yaml` already
   specifies `dns_cache_hit_rate` and `dns_cache_records` for the v1 API.

   If the legacy count fields are intentional, they should be described as
   counts/share of all counted DNS queries.  `num_dns_queries` also includes
   blocked and locally answered requests that never perform a cache lookup, so
   `num_cached_queries / num_dns_queries` is not the conventional cache hit
   rate of `hits / (hits + misses)`.

   If the v1 contract is intended instead, an explicit cache-lookup
   denominator and current cache-record count are still needed.

3. The root changelog says the dashboard shows the number and share of cached
   responses.  This backend PR has no frontend or percentage field, so please
   describe only the API/backend counters here.

4. Current `master` commits the Solid frontend's Orval output generated from
   `openapi/openapi.yaml`.  After resolving the schema drift, please run
   `npm run api:generate` in `client_v2` and include the generated `Stats`
   model update.

5. The test sets `Entry.Cached = true` while its
   `UpstreamStatistics.IsCached` remains false.  A real cached dnsproxy result
   has `IsCached = true` and is excluded from per-upstream totals, so this
   fixture models an impossible production state.  Please correct it and add
   a focused `internal/dnsforward` regression proving the production bridge
   from a cached `QueryStatistics` entry to the stats cache counter.

Before merge I would also cover cached-count persistence/reopen (including an
old GOB decoding to zero), hourly/daily aggregation, reset, and the empty or
disabled-cache cases.  The `v0.107.79` API changelog section is still the right
one, but current `master` already contains it, so the rebased entry should be
inserted into that section rather than adding a second heading.

PR #8498, opened 2026-08-01
filtering: fix engine rebuild memory usage

### LINKED ISSUE #8297: High memory and CPU usage
### Prerequisites

Linux, AMD64 (aka x86_64)

### Installation

### Setup

On one machine

### AdGuard Home version

v0.107.73

### Action

High memory and CPY usage during a couple of hours every day. This issue started after upgrading to "v0.107.73".

I have 512mb allocated to AdGuard Home and have never had this issue before the upgrade.
RAM usage is normally around 300mb and for CPU, below 1%
The high usage starts at around 9-10pm (09:00-10:00). I can see small peaks before the upgrade but never to this extent.

Image

Cant find any settings that starts anything daily during that time.

`journalctl` during those hours today (2026-03-13):
```sh
Mar 13 07:38:01 adguard CRON[401]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 07:38:01 adguard CRON[403]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 07:38:01 adguard CRON[401]: pam_unix(cron:session): session closed for user root
Mar 13 08:38:01 adguard CRON[406]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 08:38:01 adguard CRON[408]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 08:38:01 adguard CRON[406]: pam_unix(cron:session): session closed for user root
Mar 13 09:38:01 adguard CRON[412]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 09:38:01 adguard CRON[414]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 09:38:01 adguard CRON[412]: pam_unix(cron:session): session closed for user root
Mar 13 10:42:47 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:43:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:45:11 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:46:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:47:19 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:48:30 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:49:32 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:50:47 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:51:44 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:52:56 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:54:18 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:55:00 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:56:38 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:57:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:58:38 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 10:59:37 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:00:52 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:01:57 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:02:50 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:03:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:05:09 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:06:03 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:07:06 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:08:18 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:09:24 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:11:11 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:11:37 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:13:58 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:14:13 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:15:49 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:17:10 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:17:54 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:19:07 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:20:26 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:21:12 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:22:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:23:25 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:24:39 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:25:49 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:26:42 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:27:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:28:52 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:29:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:31:06 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:32:05 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:32:32 adguard CRON[417]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 11:33:01 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:33:11 adguard CRON[425]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 11:34:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:35:04 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:36:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: A process of this unit has been killed by the OOM killer.
Mar 13 11:37:01 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: Main process exited, code=killed, status=9/KILL
Mar 13 11:38:24 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: Failed with result 'oom-kill'.
Mar 13 11:39:15 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: Consumed 12min 20.928s CPU time, 489.4M memory peak.
Mar 13 11:40:24 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:41:42 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:42:34 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:43:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:45:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:45:48 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:46:20 adguard CRON[417]: pam_unix(cron:session): session closed for user root
Mar 13 11:47:46 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:48:33 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:49:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:50:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:51:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:52:46 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:53:52 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:54:55 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:56:15 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:57:00 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:58:13 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:59:05 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:00:09 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:01:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:02:10 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:03:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:04:28 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:05:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:06:18 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:07:22 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:08:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:09:26 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:10:27 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:31:29 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:32:21 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:33:20 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:34:50 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:35:48 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:37:04 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:37:51 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:39:03 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:40:31 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:41:05 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:42:28 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:43:25 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:44:28 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:44:35 adguard CRON[427]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 12:45:14 adguard CRON[441]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 12:45:29 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:47:10 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:47:31 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:49:32 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:50:47 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:51:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:52:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:53:46 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:55:13 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:55:58 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:57:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:58:35 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 12:59:39 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:00:55 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:01:32 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:01:52 adguard CRON[427]: pam_unix(cron:session): session closed for user root
Mar 13 13:03:02 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:04:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:04:56 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:06:26 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:07:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:08:18 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:09:38 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:10:34 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:11:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:13:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:13:45 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:16:18 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:17:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:18:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:20:08 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:20:26 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:22:17 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:23:03 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:24:20 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:25:12 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:26:31 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:27:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:28:25 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:29:31 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:30:31 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:32:09 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:32:50 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:34:02 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:35:05 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:36:03 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:37:09 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:38:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:39:26 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:40:20 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:41:19 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:42:21 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:43:22 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:43:36 adguard dhclient[245]: DHCPREQUEST for 192.168.1.41 on eth0 to 192.168.1.1 port 67
Mar 13 13:43:59 adguard dhclient[245]: DHCPACK of 192.168.1.41 from 192.168.1.1
Mar 13 13:46:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:46:37 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:48:11 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:48:21 adguard CRON[440]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 13:49:38 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:49:44 adguard CRON[455]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 13:50:13 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:51:41 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:52:39 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:53:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:54:38 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:56:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:57:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:58:31 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 13:59:30 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:02:43 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:03:13 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:05:45 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:07:35 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:08:00 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:09:34 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:10:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:12:45 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:13:34 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:14:57 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:16:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:18:06 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:19:30 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:20:27 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:21:43 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:22:33 adguard CRON[440]: pam_unix(cron:session): session closed for user root
Mar 13 14:22:58 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:23:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:24:52 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:25:54 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:27:00 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:28:13 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:29:17 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:30:20 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:31:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:32:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:33:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:34:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:36:21 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:36:48 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:38:15 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:39:50 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:41:27 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:41:53 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:43:28 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:44:50 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:46:09 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:47:40 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:48:20 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:52:20 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:52:49 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:54:23 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:56:36 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:57:19 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:59:12 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 14:59:45 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:07:33 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:08:07 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:10:35 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:11:11 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:13:29 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:14:45 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:15:55 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:16:52 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:18:59 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:20:01 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:21:56 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:23:44 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:26:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:27:41 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:30:01 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:32:16 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:33:48 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:36:25 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:38:35 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/qmgr
Mar 13 15:54:43 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard CRON[465]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 15:54:43 adguard CRON[451]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 15:54:43 adguard CRON[468]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 15:54:43 adguard CRON[473]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 15:54:43 adguard CRON[474]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 15:54:43 adguard CRON[475]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 15:54:43 adguard CRON[468]: pam_unix(cron:session): session closed for user root
Mar 13 15:54:43 adguard CRON[465]: pam_unix(cron:session): session closed for user root
Mar 13 15:54:43 adguard CRON[451]: pam_unix(cron:session): session closed for user root
Mar 13 15:54:43 adguard systemd[1]: Starting apt-daily.service - Daily apt download activities...
Mar 13 15:54:43 adguard systemd[1]: Starting postfix-resolvconf.service - Updates postfix copy of resolv.conf...
Mar 13 15:54:43 adguard systemd[1]: postfix-resolvconf.service: Deactivated successfully.
Mar 13 15:54:43 adguard systemd[1]: Finished postfix-resolvconf.service - Updates postfix copy of resolv.conf.
Mar 13 15:54:43 adguard dhclient[245]: bound to 192.168.1.41 -- renewal in 27621 seconds.
Mar 13 15:54:53 adguard systemd[1]: AdGuardHome.service: Scheduled restart job, restart counter is at 2.
Mar 13 15:54:53 adguard systemd[1]: Started AdGuardHome.service - AdGuard Home: Network-level blocker.
Mar 13 15:55:14 adguard systemd-networkd-wait-online[494]: Timeout occurred while waiting for network connectivity.
Mar 13 15:57:20 adguard apt-helper[479]: E: Sub-process /lib/systemd/systemd-networkd-wait-online returned an error code (1)
Mar 13 16:01:28 adguard systemd[1]: AdGuardHome.service: A process of this unit has been killed by the OOM killer.
Mar 13 16:01:28 adguard systemd[1]: AdGuardHome.service: Main process exited, code=killed, status=9/KILL
Mar 13 16:01:28 adguard systemd[1]: AdGuardHome.service: Failed with result 'oom-kill'.
Mar 13 16:01:28 adguard systemd[1]: AdGuardHome.service: Consumed 16.457s CPU time, 484.5M memory peak.
Mar 13 16:01:29 adguard systemd[1]: apt-daily.service: Deactivated successfully.
Mar 13 16:01:29 adguard systemd[1]: Finished apt-daily.service - Daily apt download activities.
Mar 13 16:01:29 adguard systemd[1]: apt-daily.service: Consumed 7.160s CPU time, 58.7M memory peak.
Mar 13 16:01:38 adguard systemd[1]: AdGuardHome.service: Scheduled restart job, restart counter is at 3.
Mar 13 16:01:38 adguard systemd[1]: Started AdGuardHome.service - AdGuard Home: Network-level blocker.
Mar 13 16:38:01 adguard CRON[557]: pam_unix(cron:session): session opened for user root(uid=0) by root(uid=0)
Mar 13 16:38:01 adguard CRON[559]: (root) CMD (cd / && run-parts --report /etc/cron.hourly)
Mar 13 16:38:01 adguard CRON[557]: pam_unix(cron:session): session closed for user root
```

A part of that ^:
```sh
Mar 13 11:35:04 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 11:36:14 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: A process of this unit has been killed by the OOM killer.
Mar 13 11:37:01 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: Main process exited, code=killed, status=9/KILL
Mar 13 11:38:24 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: Failed with result 'oom-kill'.
Mar 13 11:39:15 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
Mar 13 15:54:43 adguard systemd[1]: AdGuardHome.service: Consumed 12min 20.928s CPU time, 489.4M memory peak.
Mar 13 11:40:24 adguard postfix/master[264]: warning: unix_trigger_event: read timeout for service public/pickup
```

I dont notice any issues when using the internet during those hours, but i have fallbacks DNS resolvers in my router.

---

Replace the following command with the one you're calling or a
description of the failing action:

```sh
nslookup -debug -type=a 'www.example.com' '192.168.1.41' 

Server:         192.168.1.41
Address:        192.168.1.41#53

------------
    QUESTIONS:
        www.example.com, type = A, class = IN
    ANSWERS:
    ->  www.example.com
        internet address = 104.18.26.120
        ttl = 147
    ->  www.example.com
        internet address = 104.18.27.120
        ttl = 147
    AUTHORITY RECORDS:
    ADDITIONAL RECORDS:
------------
Non-authoritative answer:
Name:   www.example.com
Address: 104.18.26.120
Name:   www.example.com
Address: 104.18.27.120
```

### Expected result

Lower RAM and CPU usage

### Actual result

High RAM and CPU usage

Upstream DNS servers
```
https://dns10.quad9.net/dns-query
```

Installed using the default Proxmox VE script (Debian LXC), updated from the AdGuard Home UI.
Container is also upgraded from the CLI and up to date during this.
https://community-scripts.org/scripts/adguard

A couple of screenshots:
Image

Image

### LINKED ISSUE #8491: Sharp memory increase and OOM after filter update on OpenWrt
### Prerequisites

Linux, ARM64

### Installation

### Setup

### AdGuard Home version

v0.107.78

### Action

AdGuard Home is running on an OpenWrt router with approximately 1 GB of RAM and no swap.

The problem occurs intermittently while AdGuard Home is operating normally.

During the most recently monitored occurrence, the following sequence was observed:

1. At 23:00 local time, the AdGuard Home process with PID 12975 used 436528 kB RSS.
2. At 23:23 local time, AdGuard Home started an automatic filter update.
3. One of the updated filters contained 2103241 rules and had a downloaded size of 43460590 bytes.
4. At 00:00 local time, the same AdGuard Home process with PID 12975 used 888148 kB RSS.
5. AdGuard Home subsequently stopped responding.
6. After AdGuard Home restarted, DNS resolution and the web interface worked again.

The AdGuard Home debug log uses UTC. The filter update therefore appears in the log at approximately 21:23 UTC.

Relevant AdGuard Home log entries:

2026/07/25 21:23:31.510459 12975#88 [debug] filtering: starting update

2026/07/25 21:23:33.545031 12975#88 [debug] filtering: downloading update for filter id=1758964205 url=https://adguardteam.github.io/HostlistsRegistry/assets/filter_44.txt

2026/07/25 21:23:46.202887 12975#88 [info] filtering: filter updated id=1758964205 bytes_written=43460590 rules_count=2103241

2026/07/25 21:23:46.382012 12975#88 [info] filtering: updated filter id=1758964205 rules_count=2103241 prev_rules_count=2080045

The automatic filter update strongly correlates with the sudden memory increase. I do not yet have a heap profile proving that the filter update is the direct root cause.

### Expected result

AdGuard Home should update and reload filter lists without causing a permanent increase of several hundred megabytes in resident memory.

After the filter update has completed, memory usage should either return to approximately its previous level or remain low enough for AdGuard Home to continue operating normally.

DNS resolution and the AdGuard Home web interface should remain available after an automatic filter update.

### Actual result

The resident memory usage of the same AdGuard Home process increased from 436528 kB to 888148 kB within the monitored one-hour interval.

Observed values:

2026-07-25 23:00 local time

PID: 12975
VmSize: 2250940 kB
VmRSS: 436528 kB
Threads: 14

2026-07-26 00:00 local time

PID: 12975
VmSize: 2250940 kB
VmRSS: 888148 kB
Threads: 14

The PID, virtual memory size and thread count remained the same, while resident memory increased by approximately 451620 kB.

AdGuard Home then stopped responding. DNS resolution and the AdGuard Home web interface were unavailable until AdGuard Home restarted.

The Linux kernel log also confirms that AdGuard Home has been killed by the OOM-killer on multiple occasions at a similar memory level:

Out of memory: Killed process 2464 (AdGuardHome) total-vm:2182124kB, anon-rss:880568kB, file-rss:1536kB, shmem-rss:0kB, UID:0 pgtables:1904kB oom_score_adj:0

Out of memory: Killed process 6193 (AdGuardHome) total-vm:2181940kB, anon-rss:878876kB, file-rss:4992kB, shmem-rss:0kB, UID:0 pgtables:1900kB oom_score_adj:0

These kernel log entries demonstrate that the problem has occurred more than once. They are not being presented as the kernel log entry for PID 12975 specifically.

Device: GL.iNet Flint 2 / GL-MT6000

Operating system: OpenWrt 24.10.7

CPU architecture: aarch64 / ARM64

AdGuard Home version: v0.107.78

Installation: standalone manual installation on OpenWrt

AdGuard Home binary: /opt/AdGuardHome/AdGuardHome

Configuration file: /opt/AdGuardHome/AdGuardHome.yaml

Available memory:

Approximately 1 GB RAM

The query log was checked during the relevant period and was not being rotated:

2026/07/25 21:03:05.978536 12975#90 [debug] querylog: oldest log entry entry_time=2026-07-22T20:46:44.143393617Z

2026/07/25 21:03:05.978680 12975#90 [debug] querylog: not rotating now=2026-07-25T21:03:05Z rotate_time=2026-08-21T20:46:44Z

A statistics flush also occurred at 21:00 UTC:

2026/07/25 21:00:00.340962 12975#89 [debug] stats: flushing unit id=495836 req_num=1342

I have enabled the AdGuard Home pprof endpoint and configured additional memory monitoring. The monitoring checks memory usage every 15 minutes and attempts to capture heap and goroutine profiles when configured RSS thresholds are exceeded.

At this moment, the automatic filter update is a strong correlation, but I do not yet have a heap profile proving that it directly causes the memory increase.

vortexilation:
Fixes #8297.
Fixes #8491.

## Diagnosis

Both issues report AdGuard Home being OOM-killed, and both blame a release — #8297 blames
v0.107.73, #8491 blames v0.107.78. **Nothing under `internal/filtering` changed in either release** (verified against
`v0.107.72...v0.107.73` and `v0.107.77...v0.107.78`), so neither is a regression in this
package.

That claim is deliberately limited to this package, because a dependency did change:
v0.107.78 updated `urlfilter` from v0.23.2 to v0.23.4, which grew `rules.NetworkRule` from
264 to 296 bytes. On a 2.1M-rule list those 32 bytes per rule are about 64 MiB of extra
steady-state footprint, so the bump plausibly contributes to the reports against that
release, though it cannot account for a transient peak several times the steady state.
v0.107.73 is unaffected: `urlfilter` was v0.23.1 both before and after it, so #8297 has no
such explanation.

The trigger is rule lists growing; several commenters note HaGeZi's Threat Intelligence
Feeds roughly doubled in size over the same period.

The mechanism is the one the pprof profile in
[@starterhomelab's comment](https://github.com/AdguardTeam/AdGuardHome/issues/8297#issuecomment-4455710782)
points at:

```
      flat  flat%        cum   cum%
   45.05MB 31.39%   141.84MB 98.85%  urlfilter.NewDNSEngine
   34.01MB 23.70%    34.01MB 23.70%  urlfilter/rules.NewNetworkRule
   32.29MB 22.50%    32.29MB 22.50%  urlfilter/internal/lookup.(*ShortcutIndex).Add
   15.50MB 10.80%    15.50MB 10.80%  urlfilter/rules.NewHostRule
```

`DNSFilter.initFiltering` must keep the previous engines alive while it builds the new ones,
since they serve DNS for the duration — so the live heap roughly doubles. With the default
GC target of 100% the heap is only collected once it has grown to twice the live heap, so
the transient peak of a rebuild ends up at about **three times** the memory a single set of
engines takes. That is what pushes 512 MB–1 GB installations over the edge, and it explains
why the reported peaks are far larger than the steady-state footprint while `inuse_space`
after a clean refresh barely moves.

## Measurements

Rebuild of a single 2.1M-rule list, peak heap and rebuild duration:

| GC target | peak (n CPUs) | peak (1 CPU) | time (n CPUs) | time (1 CPU) |
|-----------|---------------|--------------|---------------|--------------|
| 100%      | 830 MiB       | 878 MiB      | 4.8 s         | 5.2 s        |
| 50%       | 665 MiB       | 713 MiB      | 4.7 s         | 5.1 s        |
| 20%       | 575 MiB       | 597 MiB      | 4.8 s         | 6.9 s        |

Steady state with one such engine live is 273 MiB.

## Changes

**1. Skip the rebuild when nothing has changed.** `initFiltering` now compares a
fingerprint of the rule lists it is about to build from — list ID, file path, inline data,
and the contents of each backing file — against the one the current engines were built
from, and returns early when they match.

This is the largest practical win for a common class of reports in these threads. Every
filtering-related HTTP API handler calls `EnableFilters`, so setting *any* filtering
property rebuilt both engines even when the value was unchanged. That is why users running
AdGuardHome-Sync see a rebuild storm on every sync cycle, and why "unblock one domain in the
UI" could take a box down.

The contents are read in full rather than summarized by the size and the modification time
of each file. Those are metadata, not a content identity: a restored backup or an explicit
`os.Chtimes` can leave both unchanged while the rules differ, which would keep stale rules
in use, and merely touching a file whose contents are unchanged would force a needless
rebuild. Reading the lists is far cheaper than rebuilding the engines from them, which is
what this check exists to avoid.

**2. Serialize rebuilds.** Nothing previously prevented the periodic filter update and a
configuration-driven rebuild from running concurrently, which doubles an already-doubled
peak. A new `initMu` makes `initFiltering` mutually exclusive.

**3. Lower the GC target for the duration of a rebuild** (`debug.SetGCPercent(20)`),
cutting the peak by about a third. A rebuild is background work and DNS keeps being served
by the previous engines while it runs, so a slower rebuild is a much better outcome than an
OOM kill, which takes DNS down until an operator intervenes. The cost is confined to
single-CPU systems (+1.7 s in the table above); on multi-core it is free.

The target is only ever *tightened* — a user who has set a stricter `GOGC`, or disabled the
GC entirely, keeps their setting.

20 is the knee of the curve, not a round number. Peak resident memory and rebuild duration
against the target, on a 2.1M-rule list:

| GC target | peak (n CPUs) | peak (1 CPU) | time (n CPUs) | time (1 CPU) |
|-----------|---------------|--------------|---------------|--------------|
| 100%      | 490 MiB       | 564 MiB      | 4.8 s         | 5.1 s        |
| 50%       | 406 MiB       | —            | 4.7 s         | —            |
| 20%       | 351 MiB       | 360 MiB      | 4.9 s         | 7.2 s        |
| 10%       | 331 MiB       | 339 MiB      | 5.2 s         | 10.2 s       |
| 5%        | 326 MiB       | 342 MiB      | 6.6 s         | 12.0 s       |

Below 20 the peak stops moving and the duration does not: on one CPU, 10% buys 6% of peak
for 42% more time, and 5% buys nothing at all while costing 67%.

These two tables and the one under change 5 come from a standalone harness over a
*synthetic* 2.1M-rule list, which is cheaper per rule than a real blocklist — one engine is
118 MiB live here against 273 MiB for the real list in the first table above. The absolute
figures are therefore lower than that table's throughout and the two are not directly
comparable; the ratios are what carry over, and the 100%-to-20% reduction agrees between
them.

**5. Collect before the rebuild as well as after it.** Tightening the GC target decides
when the next collection happens; it does not clear what is already dead. So a rebuild
began on top of whatever heap serving DNS had left behind and carried it resident through
its early phase, adding that garbage to the peak in full. Entering the same rebuild with
120 MiB of dead objects on the heap:

|                | peak (n CPUs) | peak (1 CPU) | time (n CPUs) | time (1 CPU) |
|----------------|---------------|--------------|---------------|--------------|
| collect after  | 372.8 MiB     | 400.3 MiB    | 4.8 s         | 8.3 s        |
| collect before | 350.2 MiB     | 362.5 MiB    | 4.8 s         | 8.7 s        |

6% off the peak on multi-core and 9% on one CPU, for 0.4 s on one CPU and nothing on
multi-core. Scaled against the reporter's 763 MB peak that is roughly 45-70 MB, on a device
whose margin got as low as 11 MB — though how much garbage is actually on the heap when a
rebuild starts depends on the query load, so 120 MiB is an assumption, not a measurement
from that device.

**4. Release resources when a rebuild fails.** If `newRuleStorage` failed partway, or the
allow-list storage failed after the block-list storage was already built, the opened rule
lists were dropped without `Close()`, leaking their file descriptors — and the early return
skipped `debug.FreeOSMemory()`, leaving the failed rebuild's peak resident. Both are now
handled, and `FreeOSMemory` runs on every path.

## Tests

New tests in `internal/filtering/initfiltering_internal_test.go` cover:

- fingerprint stability, and sensitivity to changed data, changed ID, swapped allow/block
  groups, an added list, changed file contents, a changed mtime at equal size, and a removed
  file;
- that an unchanged filter set does not rebuild and a changed one does;
- that a *failed* rebuild leaves the previous engines installed and does not record itself
  as the current state, so the next rebuild still runs;
- that a rebuild collects both before and after itself, and that a skipped one does
  neither.

`go test ./...`, `go vet ./internal/...` and `gofmt` are clean, and the filtering package
passes under `-race`.

## Review updates

Addressing @Sil3ntVip3r's [review](https://github.com/AdguardTeam/AdGuardHome/pull/8498#pullrequestreview-4839643203):

- **Hash the rule-list contents, not their metadata.** Size plus mtime is not a content
  identity, so a file replaced by different rules of the same length with the modification
  time preserved would not change the fingerprint, and `initFiltering` would keep serving the
  previous rules. `hashFileMeta` is now `hashFileContents`, streaming the file through the
  same digest.

  Reading every list on every filtering-related API request then cost about 22% more
  latency per configuration push, and almost all of it was the digest rather than the read
  — over the 43 MiB list, best of five warm: read only 11.2 ms, read + SHA-256 123.6 ms,
  read + `maphash` 11.3 ms. The fingerprint never leaves the process and is only compared
  against the one the current engines were built from, so it needs no second-preimage
  resistance, and it is seeded once per process so a collision cannot be prepared inside a
  rule list. A collision would cost a skipped rebuild and nothing worse. So it is
  `hash/maphash` now, and the check costs the read alone.

- **Serialize the changes of the garbage-collection target.** `initMu` belongs to one
  `DNSFilter` while `debug.SetGCPercent` changes the whole process, so two overlapping
  rebuilds restored each other's saved value and could leave the process at the stricter
  target permanently. The bookkeeping moved into a package-level `gcTarget` that counts the
  rebuilds in flight and only restores after the last one finishes. Its setter is injectable,
  so the overlap is covered deterministically instead of by mutating the test process's own
  target.

Both are pinned by tests that fail against the previous implementation: different bytes with
identical metadata, identical bytes with a changed mtime, and a two-instance overlap. Note
that the existing `same_size_new_mtime` case asserted the old behaviour and became
`same_bytes_new_mtime`.

The diagnosis wording is narrowed as well: the "no code change" claim now applies only to
`internal/filtering`, with the `urlfilter` v0.23.2 to v0.23.4 bump and the resulting
`NetworkRule` growth called out. Confirmed independently here: 264 to 296 bytes on 64-bit,
and unchanged across v0.107.72 to v0.107.73, so the caveat applies to #8491 and not #8297.

## Field validation

Confirmed on a 1 GB device against a 2.1M-rule list by a reporter in
[#8491](https://github.com/AdguardTeam/AdGuardHome/issues/8491#issuecomment-5195886559):
no OOM, the process survived the reload with the same PID, peak RSS 763 MB against a
365 MB baseline, settling back to 423 MB. Seven heap profiles from that reload
([analysis](https://github.com/AdguardTeam/AdGuardHome/issues/8491#issuecomment-5196536212))
show both mechanisms firing on real hardware:

- The GC target was tightened for the rebuild and restored afterwards: `NextGC/Alloc` is
  1.11 during the rebuild and 2.00 after it.
- The deferred `debug.FreeOSMemory()` ran exactly once, moving `HeapReleased` from 2.4 MiB
  to 368.8 MiB in one step.

The same profiles put 96.4% of the live heap in the two coexisting `urlfilter` engines
(290.3 MiB old plus 266.3 MiB new) and everything else in the process at 19.5 MiB, so
there is no remaining allocation of any size on the AdGuard Home side of the rebuild.

They did show one avoidable cost on a path this PR adds, fixed in
`filtering: pool the fingerprint read buffer`: the fingerprint reads each rule list through
`io.Copy`, which delegates to `os.File.WriteTo` and allocates a fresh 32 KiB buffer per list
per call. It retains nothing, but this runs on every filtering-related API request, which is
the path that exists to be cheap.

```
BenchmarkDNSFilter_initFiltering/unchanged_10000
before   33706 B/op   9 allocs/op
after      889 B/op   9 allocs/op
```

### Multi-cycle soak

The same reporter kept the build and the large list running and reported again on
[12 August](https://github.com/AdguardTeam/AdGuardHome/issues/8491#issuecomment-5263805555),
this time across repeated refresh cycles rather than a single event:

- No OOM for the duration, with the large HaGeZi list active throughout.
- Roughly 10-12 refresh cycles, counted from each list's own ~24h interval.
- The OS thread count went 10 → 11 → 12 → 12 → 13 and then held, four days flat at 13 by
  the time of the report.

The thread figure was the open question left by the heap analysis above. The shape is what
matters: the Go runtime parks threads rather than destroying them, so the count ratchets to
a high-water mark and stops, and a real leak would look like one more thread per rebuild
that never plateaus. It plateaued. The `GOMAXPROCS + 5` figure I quoted earlier was a poor
yardstick, though — this build starts at 10 threads on a clean launch of a four-core
device, so the threshold as written was already passed before any rebuild happened.

What the run does not show, stated plainly:

- It is a soak on a composite build that also carries #8522 and dnsproxy#520. The isolated
  evidence for this PR remains the A/B benchmark on frozen list bytes
  ([comment](https://github.com/AdguardTeam/AdGuardHome/issues/8491#issuecomment-5183902092)):
  peak RSS 890 MiB on the parent commit against 298 MiB on this branch, one tree, one
  toolchain, alternating order.
- The cycle count is derived from the configured refresh intervals, not from observed
  rebuilds. HaGeZi's GitHub was unreachable for several days at the end of the run, which
  left the copies AdGuard Home serves frozen — a commenter
  [verified](https://github.com/AdguardTeam/AdGuardHome/issues/8491#issuecomment-5295244296)
  the file header stuck at 9 August. Refreshes in that window fetched unchanged bytes and
  were short-circuited by change 1 rather than rebuilding anything. That is the intended
  behaviour and it means the path got real exercise, but the number of full engine rebuilds
  actually covered is lower than the cycle count.
- The reporter runs the large list plus a modest set, not the large list alongside
  everything else, so this is not a worst-case configuration.

One number from that report deserves not to be glossed over: the tightest margin in the
whole run was about **11 MB of MemAvailable**, and it came during one of the *smaller*
evening list updates rather than during the large list's own rebuild. That is expected —
`initFiltering` rebuilds every engine from every list whenever any one of them changes, so
the peak is set by the total rule count and not by which list moved — and chasing it is
what produced change 5 above. But it is worth being clear about what this PR buys. It
removes the rebuild peak that was producing the OOM kills; it does not make a 1 GB device
comfortable.

The rest is structural, and I looked for more before saying so. The GC target is at the
knee of its curve, as the table under change 3 shows. A soft memory limit does go lower —
315 MiB against 351 MiB for `GOGC=20` on the same rebuild — but only when it is sized
correctly: at 260 MiB the same rebuild took 17.6 s instead of 4.9 s and still overshot the
limit. AdGuard Home cannot know the right number, since it depends on the device and on
everything else running there, which is exactly why `GOMEMLIMIT` belongs to the operator
rather than to this code. Beyond that, the heap profiles put 96.4% of the peak in two
urlfilter engines that have to coexist for the swap.

## Notes for reviewers

Two pre-existing problems I noticed in this code but deliberately left out of scope:

- `handleFilteringSetRules` writes `d.conf.UserRules` without holding `conf.filtersMu`,
  racing the read in `enableFiltersLocked`.
- `periodicallyRefreshFilters` computes its network-error backoff as
  `ivl = max(ivl, maxInterval)`, which jumps straight to the one-hour maximum on the first
  error; the surrounding code reads as though `min` was intended.

Happy to adjust the GC target, split the commit, or drop any of the four parts if you'd
prefer a narrower change.


internal/filtering/filtering.go
```diff
@@ -768,14 +860,83 @@ func (d *DNSFilter) initFiltering(ctx context.Context, allowFilters, blockFilter
...
+	fi, err := os.Stat(path)
+	if err != nil {
+		// [ruleListFromFilter] skips the lists whose files cannot be read, so
+		// record the fact that this one is currently unavailable.
+		_, _ = fmt.Fprintf(h, "err:%v;", err)
+
+		return
+	}
+
+	_, _ = fmt.Fprintf(h, "size:%d;mtime:%d;", fi.Size(), fi.ModTime().UnixNano())
```
Sil3ntVip3r: [P2] Hash file contents, not only size and mtime

A replacement can contain different rule bytes while preserving both file size and modification time (for example, a restored file or an explicit `Chtimes`).  The fingerprint then remains unchanged and `initFiltering` returns early, continuing to serve stale rules.  Conversely, merely touching unchanged content forces an expensive rebuild.  Atomic replacement does not make size plus mtime a content identity.  Please include the file contents or a persisted content checksum in the fingerprint, with regressions for different bytes with identical metadata and identical bytes with a changed mtime.

internal/filtering/filtering.go
```diff
@@ -742,17 +771,80 @@ func ruleListFromFilter(f Filter) (rl filterlist.Interface, skip bool, err error
...
+	}
+
+	// Make sure that the OS reclaims memory as soon as possible, even if the
+	// rebuild has failed midway.
+	defer debug.FreeOSMemory()
+
+	// Only ever tighten the target for the duration of the rebuild.  A user who
+	// has set a stricter GOGC, or disabled the GC altogether, must not have
+	// that overridden.
+	if prev := debug.SetGCPercent(rebuildGCPercent); prev < rebuildGCPercent {
```
Sil3ntVip3r: [P2] Serialize changes to the process-global GC target

`initMu` belongs to one `DNSFilter`, while `debug.SetGCPercent` changes the entire process.  Two instances can therefore overlap: instance A saves 100 and sets 20; instance B saves 20; if A restores 100 before B exits, B restores 20 and leaves the process permanently at the stricter target.  AdGuard Home normally has one instance, but the package can be instantiated more than once by tests or embedders.  Please use package-level serialization around the set/restore interval (or avoid the process-global change) and add a deterministic two-instance overlap test.

Sil3ntVip3r (commented):
I found two correctness edge cases in the current patch.  Both are local to the new fingerprint/GC behavior and have focused fixes.

internal/filtering/filtering.go
vortexilation: Check again please

internal/filtering/filtering.go
vortexilation: Kindly review latest commit

Sil3ntVip3r (commented):
Re-reviewed the current head `7df0afac`.  Both correctness issues from my
earlier review are addressed:

- The fingerprint now follows file contents rather than size and mtime.
- Process-global GC-target changes are coordinated across overlapping
  `DNSFilter` instances and covered deterministically.

I also profiled this revised head against `master` (`5f6cb57d`) using synthetic
unique rule sets of 10k, 50k, and 100k rules on macOS ARM64.  At 100k rules:

- Maximum RSS fell from 89.7 MiB to 54.8 MiB (-39%).
- Sampled changed-rebuild peak fell from 47.8 MiB to 34.5 MiB (-28%), with
  about 23% longer rebuild time.
- An unchanged rebuild fell from 59.6 ms / 63.86 MiB allocated to 1.22 ms /
  0.033 MiB.
- Concurrent-rebuild peak fell from 70.7 MiB to 35.2 MiB (-50%), with the
  expected serialization latency.
- Post-GC retained heap was effectively unchanged at about 15.17 MiB, and six
  alternating rebuilds did not reproduce permanent heap growth.

The measurements support the patch's intended transient-memory and
redundant-work improvements.  They do not substitute for the reporter's 2.1M
rule list or Linux/OpenWrt/LXC validation.

One diagnosis wording caveat: v0.107.77 to v0.107.78 updated `urlfilter` from
v0.23.2 to v0.23.4, and `NetworkRule` grew from 264 to 296 bytes.  So it is
accurate to say that nothing under `internal/filtering` changed, but the
broader statement that there was no filtering-related code change should
remain narrower; the dependency change could contribute to steady memory,
although it does not explain the full transient spike.

The two requested review fixes are resolved, and the revised implementation is
well supported by the bounded profiling evidence.

vortexilation:
@Sil3ntVip3r thanks for the re-review and for the independent profiling — the diagnosis wording is now narrowed in the description, and I confirmed your dependency finding locally: `rules.NetworkRule` is 264 bytes under urlfilter v0.23.2 and 296 under v0.23.4. Worth adding that it applies to #8491 only — `urlfilter` was v0.23.1 on both sides of v0.107.73, so #8297 has no such explanation.

On the gap you flagged, here is Linux validation against a list of the reporter's size.

## Setup

- **List**: HaGeZi Threat Intelligence Feeds, **2,158,753 rules**, 42.6 MiB — the list several commenters in #8297 identify as having roughly doubled.
- **Compared**: this PR's head against its own parent (`8e56afa`), so the delta isolates the patch rather than mixing in unrelated master changes.
- **Host**: Linux x86_64 (WSL2), glibc, 36 cores, 31 GiB RAM, `GOMAXPROCS=36`.
- **Method**: peak RSS read from `VmHWM` in `/proc/self/status`, reset via `/proc/self/clear_refs` between phases so each figure covers only its own phase. Page cache warmed before both runs.

## Results

| Phase | Parent `8e56afa` | This PR | Δ peak |
|---|---|---|---|
| initial build | 4.39 s / 422 MiB | 5.58 s / 276 MiB | **−35%** |
| changed rebuild | 5.17 s / 854 MiB | 5.34 s / 564 MiB | **−34%** |
| unchanged rebuild | 4.32 s / 874 MiB / 1276 MiB alloc | **0.11 s / 314 MiB / 0 MiB alloc** | **−64%** |
| concurrent rebuild | 5.39 s / 1183 MiB | 9.99 s / 579 MiB | **−51%** |
| steady-state RSS | 404 MiB | 318 MiB | −21% |
| **overall peak** | **1183 MiB** | **579 MiB** | **−51%** |

The transient peak on a real rebuild drops by about a third, and the worst case over the whole run halves. Rebuild time costs +3% on a changed rebuild and +27% on the initial build, consistent with the ~23% you measured.

An unchanged rebuild goes from 4.3 s and 1.28 GiB allocated to 0.11 s and **zero** allocation — roughly 38x. That is the AdGuardHome-Sync case, and on a memory-constrained box it is the difference between a rebuild storm and nothing happening at all.

## On the concurrent figure

My first attempt at that row was misleading, so it is worth saying how it is built. Parking the engines on one list and then racing two rebuilds lets the fingerprint skip one of them, which measures deduplication rather than serialization. The run above parks the engines on a *third* list first, so both racing calls differ from the current state and from each other and neither can be skipped. Both sides allocated the same 2551 MiB, confirming each performed two full rebuilds — so the halved peak is the serialization, and the 9.99 s against 5.39 s is its expected latency cost.

## What this does not cover

This is glibc x86_64 with 31 GiB of RAM. It is **not** OpenWrt or LXC, not musl, not ARM, and not a 512 MiB–1 GiB memory ceiling — which is precisely where the OOM kills in these issues occur. The relative improvements should carry, but I have not demonstrated that on the platforms the reporters run, and the GC-target change is the part most likely to behave differently under a hard cgroup limit. Happy to run this on OpenWrt in a VM if that would help, or to hand over the harness so someone with the reporters' hardware can reproduce it.

Sil3ntVip3r (approved):
Re-reviewed the current head a6c55d88, including both benchmark commits added after my earlier review. The self-contained benchmark correctly alternates changed rule sets, keeps the unchanged case stable, and the helper extraction preserves those measurements. Local results on macOS ARM64 with one iteration per case were: changed 10k 6.78 MB / 60,209 allocs, unchanged 10k 33.3 KB / 9 allocs; changed 100k 65.4 MB / 600,851 allocs, unchanged 100k 33.3 KB / 9 allocs. The filtering package passed under the race detector, make go-check passed in full, govulncheck found no reachable vulnerabilities, and git diff --check passed. I found no new blocker in the current head.

Sil3ntVip3r (approved):
Re-reviewed exact head `6e6a68cf43168a4b0e088809e1d44b49fb773b04`. The only change since the previously approved head replaces SHA-256 with one process-seeded `maphash` value. The seed is stable for the life of the process, each fingerprint uses its own hash instance, and the value is only compared with the currently installed engine fingerprint. This preserves the content-identity behavior while making deliberate input collisions impractical; the residual 64-bit accidental-collision risk is negligible for this cache-invalidation use.

I independently compared the unchanged-rebuild benchmark with the SHA-256 parent on macOS ARM64. Across five runs, the 100k case fell from 1.01–1.40 ms/op to 0.33–0.57 ms/op, with allocations effectively unchanged at about 33.3 KiB and one fewer allocation. The 10k case showed the same direction. Absolute values differ from the author measurements, but the reduction is repeatable and supports the stated tradeoff.

Validation on this exact head passed the fingerprint/GC/init-filtering tests under `-race` repeated ten times, `go vet ./internal/filtering`, `git diff --check`, and the complete `make go-check` gate including the full race suite and govulncheck. I found no new blocker in the maphash change.

Sil3ntVip3r (approved):
Re-reviewed exact head `b3938476bc9a68b3b9c7778fb94d214e14753314`.

The new pooled fingerprint buffer is safe in this path: each call borrows a distinct slice from the concurrency-safe pool, `onlyReader` ensures `io.CopyBuffer` uses that buffer rather than `os.File.WriteTo`, and the added large-file regression verifies bytes beyond the first 32 KiB are included in the fingerprint.

Validation completed:
- relevant filtering/GC/init tests under `-race`, repeated ten times: passed;
- `go vet ./internal/filtering` and diff checks: passed;
- complete `make go-check` (full race suite and govulncheck): passed.

Independent five-run benchmark on macOS ARM64 for `unchanged_10000`:
- parent: 32.86–33.92 µs/op, 33,302–33,317 B/op;
- this head: 31.97–32.13 µs/op, 531–534 B/op.

That confirms the intended garbage reduction with no material latency regression. Approved.

vojacekj:
@Sil3ntVip3r or @vortexilation any update on this PR and when it can be merged and released? 

As it is still ongoing issue and last update on this PR was month ago.

Thank you.

vortexilation:
My workstation is still in repair, hopefully in 3 days I will update it

vortexilation:
@vojacekj 
Ping, please try new commits.

vojacekj:
@vortexilation I am sorry I probably worded that not clearly. I meant that it is still issue on main (current released version) so I was more asking when this can be merged and released.

To the new commits I do not know how to get that pulled to my lxc on Proxmox.

vortexilation:
@vojacekj 
What is your architecture? x86_64?, I can try to make a compilation for you, you only need to backup your adguardhome.yaml .

vojacekj:
Yeah it is x86_64 but i can wait for the release that's no problem. If it is about to happen

vortexilation:
As your wish, let's hope adguardhome have the time to merge this PR.

PR #8518, opened 2026-08-03
scripts: download before reinstall shutdown

### LINKED ISSUE #4809: Installation script shouldn't rely on server other than AGH being present
### Prerequisites

### Operating system type

### CPU architecture

### Installation

### Setup

### AdGuard Home version

v0.107.8

#### What did you do?

Try the install script to reinstall/update AGH on a machine where AGH is the only configured DNS server.

#### Expected result

Successful reinstallation/update.

#### Actual result

Since AGH is the only DNS server, once the previous installation is stopped, there is no way for `curl`/`wget` to resolve the address of the static storage from which AGH is downloaded.  So, the process fails.

#### Additional information

This is unrelated to architectures, and is relevant to unix OSes, on which the script works.

@Alex-302, please confirm that the description is correct.

Sil3ntVip3r:
## Summary

Download and validate the release archive before stopping and uninstalling an
existing AdGuard Home service during `-r` reinstallation.

Previously, the installer stopped the running DNS service first and only then
attempted the network download.  On systems where AdGuard Home is the only DNS
resolver, that ordering can prevent the download and leave DNS offline.

Fresh installs, uninstall-only mode, and the existing-installation validation
remain unchanged.

## Reproduction

The new shell integration test recorded this order on unmodified current
`master`:

`stop -> uninstall -> download -> unpack -> install`

The corrected order is:

`download -> stop -> uninstall -> unpack -> install`

It also verifies that a failed download leaves the existing service intact.

## Testing

- five installer scenarios repeated 20 times (100 checks);
- `make sh-test`;
- `make sh-lint md-lint txt-lint`;
- shell syntax and diff checks;
- `make go-check`;
- `make go-os-check`;
- repository commit hooks.

A Docker build was not run because the local Docker daemon was unavailable;
the Docker CI image change only adds the new shell test target to its existing
test invocation.

Closes #4809.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,12 @@ NOTE: Add new changes BELOW THIS COMMENT.
 
 ### Fixed
 
+- Reinstallation using the installation script on systems where AdGuard Home
+  is the only DNS server ([#4809]).
+
 - Blocked requests without an EDNS(0) OPT record ([#8183]).
 
+[#4809]: https://github.com/AdguardTeam/AdGuardHome/issues/4809
 [#8183]: https://github.com/AdguardTeam/AdGuardHome/issues/8183
 
 <!--
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@
 # This comment is used to simplify checking local copies of the Makefile.  Bump
 # this number every time a significant change is made to this Makefile.
 #
-# AdGuard-Project-Version: 12
+# AdGuard-Project-Version: 13
 
 # Don't name these macros "GO" etc., because GNU Make apparently makes them
 # exported environment variables with the literal value of "${GO:-go}" and so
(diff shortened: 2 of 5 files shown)


Sil3ntVip3r:
Follow-up self-review found and fixed one in-place reinstall edge case in commit `54c78cb9`.

If the installer was launched from inside the existing `AdGuardHome` directory, the first commit downloaded the archive into that directory and `handle_existing` then removed it before unpacking. The new regression runs the installer from inside the existing directory with a relative `-o ..` path and makes the fake extractor require the archive to still exist.

Red before the fix:

```text
install-test: archive disappeared: AdGuardHome_linux_amd64.tar.gz
make: *** [sh-test] Error 1
```

The production fix canonicalizes the existing installation parent and moves there before downloading, keeping both the working directory and package outside the directory being removed.

Validated after the fix:

```text
PASS: make sh-lint sh-test txt-lint
PASS: make go-check
PASS: git diff --check
```

The original download-before-shutdown behavior, failed-download preservation, uninstall-only behavior, validation-before-download behavior, and fresh installation path remain covered.

PR #8537, opened 2026-08-03
filtering: show progress while toggling filter lists

### LINKED ISSUE #667: [Feature Request] Filter Enable/Disable Overlay
In the world of DNS sinkhole for ads users more often than not believe quantity > quality in regards to filters.

For this reason it is not uncommon for users to have lists of 600,00 > 2,000,000 rules enabled at any one time.

Whilst enabling/disabling filters, an overlay disabling the user from carrying out further actions within AGH is required to prevent excessive functions taking place which can considerably slow down the machine.
As part of this, a message will be required noting that internet connectivity may be affected while the process is being completed.

Sil3ntVip3r:
## Summary

- wait for the filtering engine rebuild before a filter enable/disable request
  completes
- serialize rebuilds and invalidate queued or already-received stale snapshots
  so an older asynchronous rebuild cannot restore the previous filter state
- show a modal, focus-trapped progress overlay while filter toggles or edits are
  still applying
- count overlapping frontend operations so the overlay remains until every
  update and status refresh has finished

This keeps the UI from accepting conflicting actions while the old filtering
engine is still active and makes the rebuild state visible to the user.

Closes #667.

## Validation

- `go test -race -count=20 -run '^(TestDNSFilter_handleFilteringSetURLWaitsForEngine|TestDNSFilter_handleFilteringSetURLSupersedesQueuedEngine)$' ./internal/filtering`
- `go test -race -count=1 ./internal/filtering`
- `make go-check`
- `NODE_OPTIONS=--no-experimental-webstorage npm run test -- src/__tests__/filter-toggle-progress.test.tsx`
- `NODE_OPTIONS=--no-experimental-webstorage npm run check`
- `NODE_OPTIONS=--no-experimental-webstorage npm run build-prod`
- `make md-lint`
- `npm run translations:check`

The focused frontend suite has five deterministic cases covering toggle, edit,
request failure, status-refresh completion, modal behavior, and overlapping
updates.  The full frontend gate passed 624 tests, and the production build
compiled 2,345 modules.

`npm run locales:check` still reports the repository's existing stale
`locales.generated.ts`; this change adds English strings but does not change
the locale list, Twosky configuration, or generated imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ NOTE: Add new changes BELOW THIS COMMENT.
 
 ### Changed
 
+- Enabling or disabling a filter list now blocks other UI actions and shows
+  progress while the filtering engine is rebuilt ([#667]).
+
 - The `edge` channel has been switched to the new UI and versioning scheme.
 
 ### Deprecated
@@ -44,6 +47,7 @@ NOTE: Add new changes BELOW THIS COMMENT.
 
 - Blocked requests without an EDNS(0) OPT record ([#8183]).
 
+[#667]:  https://github.com/AdguardTeam/AdGuardHome/issues/667
 [#8183]: https://github.com/AdguardTeam/AdGuardHome/issues/8183
 
 <!--
--- a/client_v2/src/__locales/en.json
+++ b/client_v2/src/__locales/en.json
@@ -385,6 +385,8 @@
   "filter_category_security_desc": "Lists designed specifically to block malicious, phishing, and scam domains",
   "filter_removed_successfully": "Removed blocklist: <strong>%value%</strong>",
   "filter_removed_successfully_allowlist": "Removed allowlist: <strong>%value%</strong>",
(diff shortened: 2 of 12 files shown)


Sil3ntVip3r:
Addressed one final completion-contract edge case found during the closure audit.

The synchronous /control/filtering/set_url path previously waited for engine initialization but still returned HTTP 200 when initialization failed, because EnableFilters logged and swallowed the error. A deterministic regression reproduced the mismatch: expected 500, actual 200, while the prior engine remained installed.

The exported EnableFilters behavior remains unchanged for existing asynchronous/log-only callers. The synchronous handler now uses a private error-returning helper, reports HTTP 500 when the rebuild fails, and leaves the previous working engine installed. Regression coverage verifies the status and body as well as old-engine preservation.

Validation passed: the focused sequencing/error tests under the race detector repeated 30 times; the full internal/filtering race suite; go vet ./internal/filtering; make go-check; make md-lint; make txt-lint; formatting and git diff --check. The corrective diff is backend-only.

PR #8540, opened 2026-08-03
filtering: use validators for unchanged lists

### LINKED ISSUE #7050: ETag comparison
### Prerequisites

### The problem

High bandwidth usage if big filter lists get updated frequently.

### Proposed solution

Compare the ETag of the local filter file with the one on the server before downloading.

I update filters with curl now because it also uses way less RAM and IO.
Example (thx to ChatGPT):

```
#!/bin/bash

# URL of the file to download
URL="https://example.txt"

# Local file to save the downloaded content
LOCAL_FILE="/mnt/example.txt"

# File to store the ETag
ETAG_FILE="/mnt/etagexample.txt"

# Function to download the file
download_file() {
    echo "Downloading file..."
    curl -o "$LOCAL_FILE" -D - "$URL" | grep -i etag > "$ETAG_FILE"
}

# Check if the ETag file exists
if [ -f "$ETAG_FILE" ]; then
    # Get the saved ETag
    SAVED_ETAG=$(cat "$ETAG_FILE" | awk -F'"' '/etag/ {print $2}')

    # Make a HEAD request to get the current ETag from the server
    CURRENT_ETAG=$(curl -I "$URL" 2>/dev/null | grep -i etag | awk -F'"' '/etag/ {print $2}')

    # Compare the ETags
    if [ "$SAVED_ETAG" != "$CURRENT_ETAG" ]; then
        echo "ETag has changed. Downloading new file..."
        download_file
    else
        echo "ETag has not changed. No download needed."
    fi
else
    echo "No previous ETag found. Downloading file..."
    download_file
fi
```

Sil3ntVip3r:
## Summary

- cache HTTP `ETag` and `Last-Modified` validators for downloaded filter lists
- bind cached validators to the configured URL and SHA-256 of the exact local
  contents before sending a conditional request
- preserve correctness across redirects, manual cache edits, missing files,
  checksum collisions, failed downloads, and validator removal
- remove obsolete validator metadata when a filter is deleted or no longer has
  a reusable HTTP representation

## Testing

- [x] focused HTTP-validator regression tests under `-race`
- [x] full `internal/filtering` package under `-race`
- [x] `make go-check`
- [x] `git diff --check`

Fixes #7050.

### DIFF
--- a/internal/filtering/filter.go
+++ b/internal/filtering/filter.go
@@ -2,6 +2,9 @@ package filtering
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
@@ -21,11 +24,26 @@ import (
 	"github.com/AdguardTeam/golibs/errors"
 	"github.com/AdguardTeam/golibs/ioutil"
 	"github.com/AdguardTeam/golibs/logutil/slogutil"
+	"golang.org/x/net/http/httpguts"
 )
 
 // filterDir is the subdirectory of a data directory to store downloaded
 // filters.
 const filterDir = "filters"
+
+// filterHTTPMetadataSuffix is the suffix of a file containing the cached HTTP
+// validators of a downloaded filter.
+const filterHTTPMetadataSuffix = ".metadata"
+
+// filterHTTPMetadata contains the HTTP validators cached for a downloaded
+// filter.  URL prevents using validators after a filter source has changed,
+// and Digest binds them to the exact cached contents.
+type filterHTTPMetadata struct {
+	URL          string `json:"url"`
+	ETag         string `json:"etag,omitempty"`
+	LastModified string `json:"last_modified,omitempty"`
+	Digest       string `json:"digest"`
+}
 
 // FilterYAML represents a filter list in the configuration file.
 //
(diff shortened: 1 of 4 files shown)


Sil3ntVip3r:
Follow-up audit found and fixed one cleanup edge case in commit 750aecb8.

When an enabled HTTP filter with cached validators changed to a different URL
while also being disabled, `filterSetProperties` returned through the disabled
path without updating or removing the old `.metadata` sidecar. URL-and-digest
validation prevented reuse, but the obsolete source metadata remained on disk.

Test-first proof against the previous head:

```text
--- FAIL: TestDNSFilter_filterSetPropertiesHTTPMetadata/changed_url
    file ".../filters/1.txt.metadata" exists
```

The fix removes metadata only when the URL actually changes on the successful
disabled path. Disabling the same URL deliberately retains its valid metadata,
and enabled URL-change failures still preserve the old metadata for rollback.

Final validation:

- `go test -race -count=1 -run '^TestDNSFilter_filterSetPropertiesHTTPMetadata$' ./internal/filtering`
- focused validator/metadata suite under `-race -count=20`
- `go test -race -count=1 ./internal/filtering`
- `go vet ./internal/filtering`
- `make go-check`
- `make go-os-check` (Darwin, FreeBSD, OpenBSD, Windows, Linux amd64/386)
- `gofmt -d`, `git diff --check`, and merge-tree conflict scan

All final checks passed. The initial restricted local runs that needed loopback
or the normal Go build cache were rerun outside that sandbox before the results
above were recorded.

Issue #3113, opened 2021-05-08
Large stats.db causes slowdowns in DNS lookups


PurrfectLight:
I have Adguard home running on a Synology NAS (HDD, no SSD). The switch from 105.2 to 106 caused dns lookup delays that did not appear in the lookup times in the query log. However, after clearing the history (deleting the 25mb stats.db) the delays were gone again.

This probably has something to do with a relatively slow insert (or equivalent) in the SQLite DB when it increases in size.

ainar-g:
Hello.  The file isn't an SQLite database, unfortunately, but a BoltDB one.  This is tangentially related to #2290, as we've been struggling to fine a performant, portable, reliable, and pure-Go database format.  We have some experiments in the works, but unfortunately I cannot give any guarantees for now.

In the mean time, we'll probably need to be smarter about how we update statistics and its impact on the overall performance.  Perhaps the DB's schema needs some rethinking as well.

Sil3ntVip3r:
I profiled the current bbolt-backed statistics path on `master` (`5f6cb57d`).
`loadUnits` currently holds `currMu.RLock` while it opens the database snapshot
and loads/decodes historical units.  That can block DNS-path statistics updates
for the entire history load.

I have a narrow patch that binds the current in-memory unit to a read-only
bbolt snapshot under a short lock, then releases `currMu` before historical
reads and GOB decoding.  The ordering prevents both lost and double-counted
data across an hourly flush.

New tests cover an active writable transaction and a concurrent hourly flush.
Persisted-history benchmarks cover 24, 168, 720, and 2,160 hours; the 90-day
case measured about 77 ms and 156 MB for `loadUnits` on this machine.  Repeated
race tests, the full `internal/stats` package, and full repository Go checks
pass.

Proposed branch:
https://github.com/Sil3ntVip3r/AdGuardHome/tree/codex/3113-stats-io-profile

Does this approach fit the intended direction for #3113?  If so, I can open it
as a focused PR that updates rather than closes the broader issue.

PR #8541, opened 2026-08-03
stats: avoid blocking updates while loading history

Sil3ntVip3r:
## Summary

- snapshot the current in-memory statistics unit and a read-only bbolt view
  under a short lock
- release the DNS-path update lock before loading and decoding historical units
- preserve an exact boundary across concurrent hourly flushes so the current
  unit is neither lost nor counted twice
- add persisted-history benchmarks for 24 hours through 90 days

## Evidence

On current unmodified `master`, both new concurrency regressions fail because
`loadUnits` keeps `currMu` held while opening/loading a writable transaction:

```text
TestStatsCtx_LoadUnitsDoesNotBlockUpdate: did not receive after 1s
TestStatsCtx_LoadUnitsConcurrentFlush: did not receive after 1s
```

With this change, both tests pass 20 consecutive runs under `-race`.  The
90-day persisted-history benchmark on an Apple M4 Pro measured approximately
74 ms for `loadUnits` and 90 ms for `getData`; the important behavioral change
is that those reads no longer hold the update lock for that duration.

## Testing

- [x] focused concurrency regressions, 20 times under `-race`
- [x] full `internal/stats` package under `-race`
- [x] persisted-history benchmarks at 24, 168, 720, and 2,160 hours
- [x] `make go-check`
- [x] `git diff --check`

This addresses the confirmed DNS-path blocking behavior in #3113 without
closing the broader storage/schema investigation.


Sil3ntVip3r:
Updated against current `master` (`b2e25729`) with a normal merge; the only conflict was the changelog, and both entries are retained.

While revalidating the integration, I found and fixed one remaining blocking path in this PR’s scope: `loadUnits` previously held its bbolt read transaction while decoding all stored units. A map-growing hourly flush can then wait for that read transaction while holding `currMu`, delaying synchronous stats updates on the DNS path. The updated implementation copies the consistent stored-unit snapshot, closes the read transaction, and only then decodes it.

The new regression forces actual stats-database growth, pauses decoding, and verifies that `Update` completes before decoding is released. It fails on the prior PR head and passes on exact head `4bd346085a66e05084d53e0d8ac6ebcd2702e1c1`.

Validation:
- all three concurrency regressions, 20 times under `-race`;
- `go test -race -count=1 ./internal/stats ./internal/dnsforward`;
- repository Go checks, text/Markdown linting, whitespace checks, and the repository commit hook’s race-enabled suite;
- local 90-day one-iteration loader benchmark remained comparable to the prior head (79.1 ms vs. 79.8 ms).

GitHub now reports this PR as mergeable.

PR #8544, opened 2026-08-03
dnsforward: record upstream errors in query log and stats

### LINKED ISSUE #5826: Query log doesn't work without connection.
### Prerequisites

### Operating system type

macOS (aka Darwin)

### CPU architecture

AMD64

### Installation

### Setup

On one machine

### AdGuard Home version

edge

#### What did you do?
disable internet connection on the machine, make a test query
```
$ dig IN AAAA 'example.org' @127.0.0.1 -p 5357
```

#### Expected result

The query is logged and shown in the UI.

#### Actual result

The query is not presented.

Sil3ntVip3r:
## Summary

- record the generated SERVFAIL response when upstream resolution returns an
  error instead of returning before query-log and statistics processing
- classify the entry as `NotFilteredError`
- restore the original question before logging when a CNAME rewrite changed
  the upstream request
- preserve the existing upstream error returned to the caller

## Evidence

With only the new regression tests applied to current `master`, the query-log
parameters are absent and the CNAME case retains the rewritten question.  The
same tests pass after applying this production change.

## Testing

- [x] focused upstream-error regression, 20 times under `-race`
- [x] full `internal/dnsforward` package under `-race`
- [x] test-only red run and production-patch green run on current master
- [x] `make go-check`
- [x] `git diff --check`

Fixes #5826.


Sil3ntVip3r:
Resolved the only integration conflict by merging current upstream `master` (`b2e25729`) into this branch.  The conflict was confined to `CHANGELOG.md`; the resolution preserves both the upstream DNS64 entry and this PR’s #5826 entry, with no functional scope change.

Validation on the resulting exact head `4f58ff4051db1f420f0d1c254acbf25d4926c2da`:

- upstream-error regression: 20 runs under `-race`;
- affected packages under `-race`: `internal/dnsforward`, `internal/filtering`, `internal/querylog`, and `internal/stats`;
- `make go-check`;
- `make md-lint txt-lint`; and
- whitespace checks.

GitHub now reports this PR as mergeable.

Issue #583, opened 2019-02-11
Rework the dnsforward.Server periodic tasks


ameshkov:
The current implementation might be problematic if AG Home is restarted periodically. Also, it won't work properly in the case if we re-create the `dnsServer` instance.

PR #8566, opened 2026-08-16
go.mod: bump insomniacslk/dhcp past the nclient4 ReadFrom panic fix

thc1006:
This bumps `insomniacslk/dhcp` past insomniacslk/dhcp#583, which I wrote. AdGuard Home reaches the affected code through its check for other DHCP servers: on Linux `aghnet.listenPacketReusable` returns `nclient4.NewRawUDPConn` and `tryConn4` calls `ReadFrom` on it (`internal/aghnet/dhcp_unix.go`, `interfaces_linux.go`), reached from the DHCP HTTP handler in `internal/dhcpd/http_unix.go`.

Before #583, `BroadcastRawUDPConn.ReadFrom` subtracted the 8-byte UDP header from the IPv4 payload length without checking the payload was that long, so a malformed reply could drive the length negative and panic on a bad slice bound. A malformed frame can trigger a panic in that probe receive path; whether it becomes a process-wide termination depends on the surrounding recovery boundary. The current pin is the June 3 snapshot, before the fix.

The change is dependency metadata only: one line in `go.mod` and the matching hashes in `go.sum`. After `go mod tidy`, `go build ./...` and `go vet ./internal/aghnet/... ./internal/dhcpd/...` pass. No CVE for it, and I haven't tried to reproduce a crash inside AdGuard Home.

CONTRIBUTING asks to discuss changes in an issue first. Since this is a one-line bump for an already-merged upstream fix I opened the PR directly, but I'm happy to file an issue if you'd rather track it that way.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/google/renameio/v2 v2.0.2
 	github.com/google/uuid v1.6.0
 	github.com/gopacket/gopacket v1.7.0
-	github.com/insomniacslk/dhcp v0.0.0-20260603135910-a415979eb11e
+	github.com/insomniacslk/dhcp v0.0.0-20260719225207-c76316d4aa82
 	// TODO(e.burkov):  Get rid of this dependency.  At least, don't update it,
 	// unless all custom templates will be migrated to the new format, see
 	// https://github.com/kardianos/service/issues/418.
--- a/go.sum
+++ b/go.sum
@@ -103,8 +103,8 @@ github.com/gorilla/websocket v1.5.3 h1:saDtZ6Pbx/0u+bgYQ3q96pZgCzfhKXGPqt7kZ72aN
 github.com/gorilla/websocket v1.5.3/go.mod h1:YR8l580nyteQvAITg2hZ9XVh4b55+EU/adAjf1fMHhE=
 github.com/hugelgupf/socketpair v0.0.0-20190730060125-05d35a94e714 h1:/jC7qQFrv8CrSJVmaolDVOxTfS9kc36uB6H40kdbQq8=
 github.com/hugelgupf/socketpair v0.0.0-20190730060125-05d35a94e714/go.mod h1:2Goc3h8EklBH5mspfHFxBnEoURQCGzQQH1ga9Myjvis=
-github.com/insomniacslk/dhcp v0.0.0-20260603135910-a415979eb11e h1:7j1+lOuGBg7PQF1RxeVx0iP+/GpiAxSG/F+9t5JOS94=
-github.com/insomniacslk/dhcp v0.0.0-20260603135910-a415979eb11e/go.mod h1:qfvBmyDNp+/liLEYWRvqny/PEz9hGe2Dz833eXILSmo=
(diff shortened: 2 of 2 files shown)


xhon-pelushi (commented):
Checked this one over since a pseudo-version bump is hard to eyeball. Everything in the description holds up, and the bump is about as clean as this kind of change gets.

**The pin is exactly the fix commit, and nothing else rides along.** `proxy.golang.org` resolves `v0.0.0-20260719225207-c76316d4aa82` to commit `c76316d4aa825673b208730779843d0711021e81` at `2026-07-19T22:52:07Z`, which is precisely the merge commit of insomniacslk/dhcp#583. Comparing the two pins, `a415979eb11e...c76316d4aa82` is **ahead by one commit**, touching two files: `dhcpv4/nclient4/conn_unix.go` (+14/-3) and `dhcpv4/nclient4/conn_unix_test.go` (+121/-0). So there is no unrelated upstream drift being pulled in with it.

**The `go.sum` lines match the checksum database.** Both entries the diff adds are byte-identical to what `sum.golang.org` returns for that version, and the `/go.mod` hash is unchanged from the old pin, which is consistent with the dependency's own `go.mod` not having changed.

**The call path is real.** `listenPacketReusable` returns `nclient4.NewRawUDPConn` at `internal/aghnet/interfaces_linux.go:22`, and `tryConn4` calls `c.ReadFrom(b)` at `internal/aghnet/dhcp_unix.go:178`, reached via `discover4` ← `CheckOtherDHCP` ← `setOtherDHCPResult` (`internal/dhcpd/http_unix.go:624`) ← `handleDHCPFindActiveServer` (`:568`).

**On the open question in the description — what the recovery boundary is.** `handleDHCPFindActiveServer` is an ordinary `net/http` handler on a plain `http.Server` (`internal/home/web.go:367` and `:467`), so `conn.serve`'s deferred recover catches the panic: it logs `http: panic serving …` with a stack trace and drops that connection. It does **not** terminate the process. Worth stating plainly, because it puts the impact at "the admin's *Check for another DHCP server* request dies and a stack trace lands in the log" rather than a remote crash — and reaching it at all needs an attacker on the same L2 segment (this is a raw broadcast UDP socket on the interface) *while* an administrator has triggered that check. Still worth fixing, just not a remote kill.

**One thing the description undersells:** the upstream patch adds two guards, not one. Besides rejecting frames whose IPv4 payload cannot hold a UDP header, it also adds `if !buf.Has(dhcpLen) { continue }` before `buf.Consume(dhcpLen)`, which covers the separate case of a frame declaring a payload longer than it actually carries. That is arguably the more interesting half.

On the process point you raise at the end: I have no standing with the project either way, but for what it is worth, a one-line bump to an already-merged upstream fix with no other commits in range seems like a reasonable thing to have sent directly.

thc1006:
Closing this as redundant: master already pin `insomniacslk/dhcp` at `v0.0.0-20260728151720-c308df0fdcef`, which has the insomniacslk/dhcp#583 fix `c76316d4aa82` as a strict ancestor, so AdGuard Home reach the fixed code without this bump. Thanks @xhon-pelushi for the careful review.

